PR #1, opened 2016-08-25
A few bug fixes, readme, VS build, content update

mlorbetske:
Remove build status from old build system from readme.md
Fix VS build for Edge
Address bug in CppStyleEvaluatorDefinition where symbols would be
resolved inside literals
Update GlobalRunSpec to handle some of the new classes of "if"
Update SimpleConfigModel to accept some of the new classes of "if"
Fix reading bools and ints from config (fixed in JExtensions)
Update the built in templates to Preview 2 content


mlorbetske:
+ @seancpeters @livarcocc @blackdwarf

PR #2, opened 2016-08-26
Added proper handling of block comments within conditional blocks.

seancpeters:
TODO: decide if we want to go further with dealing with unbalanced
comments.

### DIFF
--- a/test/Microsoft.TemplateEngine.Core.UnitTests/ConditionalTests.cs
+++ b/test/Microsoft.TemplateEngine.Core.UnitTests/ConditionalTests.cs
@@ -57,16 +57,25 @@ namespace Microsoft.TemplateEngine.Core.UnitTests
         {
             get
             {
+                // This is the operationId (flag) for the balanced nesting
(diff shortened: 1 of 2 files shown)


test/Microsoft.TemplateEngine.Core.UnitTests/ConditionalTests.cs
```diff
@@ -57,16 +57,25 @@ private IOperationProvider[] XmlStyleCommentConditionalsOperations
         {
             get
             {
+                // This is the operationId (flag) for the balanced nesting
+                string commentFixingOperationId = "Fix pseudo comments";
+
+                // This is not an operationId (flag), it does not toggle the operation.
+                // But conditional doesn't care, it takes the flags its given and sets them as appropriate.
+                // It lets BalanceNesting know it's been reset
+                string commentFixingResetId = "Reset pseudo comment fixer";
```
mlorbetske: We should reuse (copy) these helpers from the tests in the config reader for making the default XML/HTML/Razor style directives

mlorbetske:
:shipit:

PR #3, opened 2016-08-26
Cleanup

mlorbetske:
Fix some code quality things
Break incorrect dependency on Core by creating Utils
Update setup to get that working
Update test project to get VS build more reliable
Expose and use No<T> instead of one-off empty list singletons


mlorbetske:
@seancpeters

src/Microsoft.TemplateEngine.Utils/No.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System.Collections.Generic;
+
+namespace Microsoft.TemplateEngine.Utils
+{
+    public static class No<T>
```
seancpeters: Possibly consider a different name for this, maybe: "Empty"?

seancpeters:
Looks good to me.

Issue #5, opened 2016-08-26
Make a more generic construction mechanism for operations and macros


mlorbetske:
Currently the config system has to understand everything, delegating the settings read to the operation factory itself seems like a better approach

PR #15, opened 2016-09-01
Ground work for #5

mlorbetske:


src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/Macros/ConstantMacro.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace Microsoft.TemplateEngine.Orchestrator.RunnableProjects.Macros
+{
+    internal class ConstantMacro : IMacro
+    {
+        public string Type => "constant";
+
+        public void Evaluate(string variableName, IVariableCollection vars, JObject def, IParameterSet parameters, ParameterSetter setter)
+        {
+            string format = def.ToString("action");
+            bool utc = def.ToBool("utc");
```
seancpeters: This is the same as NowMacro. I think NowMacro is correct and this one isn't.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/GlobalRunSpec.cs
```diff
@@ -3,21 +3,38 @@
...
+                new IncludeConfig(),
+                new MacrosConfig(),
+                new RegionConfig(),
+                new ReplacementConfig()
+            };
+
+            operationConfigReaders.Sort((x, y) => x.Order.CompareTo(y.Order));
+            OperationConfigReaders = operationConfigReaders;
+        }
+
```
seancpeters: Do we want to allow configuration of the list of operationConfigReaders, as opposed to having they explicitly stated in the static constructor?

PR #16, opened 2016-09-01
Move creation of strongly typed param values

seancpeters:
Resolving the data types of parameters used to be in GlobalRunSpec, but could happen many times per parameter. Now the logic is in RunnableProjectGenerator.cs and it's leveraged via TemplateCreator.cs


src/Microsoft.TemplateEngine.Edge/Template/TemplateCreator.cs
```diff
@@ -173,7 +173,9 @@ public static void ResolveUserParameters(IParameterSet templateParams, IReadOnly
                     {
                         if (paramFromTemplate.DataType == "bool")
                         {
-                            templateParams.ResolvedValues[paramFromTemplate] = "true";
+                            // could probably directly assign bool true here, but best to have evrything go through the same process
```
mlorbetske: typo: "evrything"

PR #20, opened 2016-09-02
Setup processing of post operation configuration

seancpeters:
TODO: process the action operations.


livarcocc:
@mmitche Hey Matt, I added a netci.groovy file to this repo, besides that, do we need to do anything to get CI going for it?

src/Microsoft.TemplateEngine.Abstractions/IPostAction.cs
```diff
@@ -0,0 +1,15 @@
...
+{
+    public interface IPostAction
+    {
+        int Order { get; }
+
+        string Name { get; }
+
+        IReadOnlyList<IPostActionOperation> Operations { get; }
+
+        IReadOnlyList<IPostActionOperation> AlternateOperations { get; }
```
livarcocc: Is this alternate in the send of run one and then another or should this be Alternative operations?

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/ConfigModel.cs
```diff
@@ -72,47 +76,47 @@ public IRunnableProjectConfig ReprocessWithParameters(IParameterSet parameters,
...
             List<string> classifications = new List<string>();
-            temp.Classifications = classifications;
+            config.Classifications = classifications;
             foreach (JToken item in source.Get<JArray>(nameof(Classifications)))
             {
                 classifications.Add(item.ToString());
             }
 
-            temp.Config = new Dictionary<string, JObject>();
+            config.Config = new Dictionary<string, JObject>();
```
livarcocc: That is super meta now.

src/Microsoft.TemplateEngine.Abstractions/IPostAction.cs
seancpeters: Alternative is probably a better term than Alternate. The idea is that the Operations will run if they can be run. If they can, then the alternate operations will be ignored.
If the regular operations can't be run, then see if the alternate operations can be run (and run them if they can be).

mmitche:
Did you follow the rest of the instructions here?: https://github.com/dotnet/dotnet-ci/blob/master/docs/CI-SETUP.md

src/Microsoft.TemplateEngine.Abstractions/IPostAction.cs
mlorbetske: I think these are intended to just be user facing instructions, we should rename/retype as such. Perhaps following up with Damian for clarification on what should happen if we didn't understand the command family (`name` I'm guessing is the equivalent to the family of commands).

I'm also a little unclear on `CommandText`, does it give the flexibility for callers to understand how to do parameter substitutions or whatnot if required? I suppose they could run through the processor to replace tokens and whatnot, but giving some sort of a structured payload (maybe just key value pairs) instead and leaving it up to the thing that interprets the command family in the caller to appropriately structure the command line or determine the component in the caller (maybe it invokes something in VS/XS/etc. to take an action instead of shelling out).

My imagination on how this would be populated (without changing away from CommandText):
Order => 0
Name => NPM
Operations => [ { CommandText => "npm install" } ]
ManualInstructions (or something like that) => [ "Run 'npm install', if you don't have npm, you can get it from (place that one gets npm from)" ]

seancpeters: I agree with all of what you’ve mentioned here. I set up “CommandText” basically as a placeholder until we we’re more clear regarding what exactly we want. It may not be necessary if we agree that what I’m currently calling “Name” is mapped in code into exactly what we want run. What I mean is (using your example), if name = “npm” gives us everything we need to run the npm action, then CommandText is superfluous.

Let’s find a couple minutes to go over this and finzalize what’s needed.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/ConfigModel.cs
mlorbetske: I'll tweak the naming in the next PR

PR #21, opened 2016-09-09
Fix local dev environment setup, fix test run, switch from netcoreapp1.0 to netstandard1.3, support single quoting support in the expression evaluator, add quote matching

mlorbetske:


mlorbetske:
Forgot to deal with the `config.Config` thing - I'll add a commit in the morning to address it if anyone feels strongly, otherwise it can be sorted out later

seancpeters:
Looks good. I'm wondering if we want to deal with nested quotes - double quotes nested inside single quotes or vice-versa. It was outside the scope of my original plan, but if nested quotes are encountered, they cause the debug assert near the end of the method. We'd have to do something more like the brace depth than the current handling of quotes.

PR #22, opened 2016-09-09
Added handling of nested quotes in the Cpp evaluator.

seancpeters:
When quotes are unbalanced, the behavior is left-greedy, meaning if there
are unbalanced quotes within a quoted string, the inner unbalanced quotes
will be assumed to end when their outer quoting ends.

### DIFF
--- a/src/Microsoft.TemplateEngine.Core/Expressions/Cpp/CppStyleEvaluatorDefinition.cs
+++ b/src/Microsoft.TemplateEngine.Core/Expressions/Cpp/CppStyleEvaluatorDefinition.cs
@@ -170,7 +170,6 @@ namespace Microsoft.TemplateEngine.Core.Expressions.Cpp
 
                         TokenFamily foundTokenFamily = (TokenFamily)token;
 
-                        // new
                         if (foundTokenFamily == TokenFamily.QuotedLiteral || foundTokenFamily == TokenFamily.SingleQuotedLiteral)
                         {
                             if (quoteNesting.Count == 0)
@@ -266,7 +265,6 @@ namespace Microsoft.TemplateEngine.Core.Expressions.Cpp
                 bufferLength = processor.CurrentBufferLength;
             }
 
-            //Debug.Assert(currentTokenFamily != TokenFamily.QuotedLiteral && currentTokenFamily != TokenFamily.SingleQuotedLiteral,
             Debug.Assert(quoteNesting.Count == 0,
                 $"Malformed predicate due to unmatched quotes. InitialBuffer = {initialBuffer} currentTokenFamily = {currentTokenFamily} | TokenFamily.QuotedLiteral = {TokenFamily.QuotedLiteral} | TokenFamily.SingleQuotedLiteral = {TokenFamily.SingleQuotedLiteral}");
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.TemplateEngine.Core/Expressions/Cpp/CppStyleEvaluatorDefinition.cs
```diff
@@ -113,6 +116,8 @@ public static bool CppStyleEvaluator(IProcessorState processor, ref int bufferLe
             }
 
             bool first = true;
+            Stack<TokenFamily> quoteNesting = new Stack<TokenFamily>();
```
mlorbetske: Quotes in quotes shouldn't have to be tracked (presumably for balancing?) - ex `"they're"` is perfectly OK. We should just maintain a "is in quoting mode and here's my original open quote type" to determine when the block ends (a quote matching the original)

src/Microsoft.TemplateEngine.Core/Expressions/Cpp/CppStyleEvaluatorDefinition.cs
```diff
@@ -30,6 +30,9 @@ public static bool EvaluateFromString(string text, IVariableCollection variables
 
         public static bool CppStyleEvaluator(IProcessorState processor, ref int bufferLength, ref int currentBufferPosition)
         {
+            // for debugging:
```
mlorbetske: I'll remove this after the merge

src/Microsoft.TemplateEngine.Core/Expressions/Cpp/CppStyleEvaluatorDefinition.cs
```diff
@@ -167,30 +172,30 @@ public static bool CppStyleEvaluator(IProcessorState processor, ref int bufferLe
...
-                                // this is the lead quote
-                                currentTokenFamily = foundTokenFamily;
-                                currentTokenBytes.AddRange(trie.Tokens[token]);  // Add the quote. It's necessary for the evaluator to infer the type - which itself needs work.
+                            {   // this is a different quote type. Treat it like a non-match, just add the token to the currentTokenBytes
+                                currentTokenBytes.AddRange(trie.Tokens[token]);
                             }
                         }
-                        else if (currentTokenFamily == TokenFamily.QuotedLiteral || currentTokenFamily == TokenFamily.SingleQuotedLiteral)
-                        {   // we're inside a quoted literal, the token found by the trie should not be processed, just included with the literal
+                        else if (inQuoteType != TokenFamily.Whitespace)
```
mlorbetske: Feels like we should have a separate enum for quote types and do the translation back & forth rather than having to use the convention that non-"in quote" situations are whitespace

seancpeters: I was thinking that we should really have a TokenFamily.None value for the enum. I used WhiteSpace in lieu of that.

mlorbetske: Agreed, I have something similar to this in the new evaluator - we should see if it's possible to converge these two

Issue #25, opened 2016-09-11
Support inline XML conditions


mlorbetske:
Several items here:

* Port an MSBUILD style condition evaluator https://github.com/mono/monodevelop/blob/master/main/src/core/MonoDevelop.Core/MonoDevelop.Projects.MSBuild/MSBuildEvaluationContext.cs and/or this https://github.com/Microsoft/msbuild/blob/master/src/XMakeBuildEngine/Evaluation/ConditionEvaluator.cs for syntax support.
* Brace balancing should not be authoritative for expression ends (the value of the conditioning attribute can be pre-read and processed, searching for the end-quote, collecting the expression along the way)
* The root variable collection should be passed with a format string of `$({0})`
* Files matching the pattern `**/*.*proj` should use this new conditional evaluator instead of the existing `Conditional` provider
* `**/*.*proj` files should have a default attribute for evaluation called `Condition`
  * This may produce several variations that consume different common surrounding whitespace configurations
  * The basic form of this would be something like `Condition="`
* If the condition evaluates to false...
  * Reposition the output stream back to the start tag (by scanning backwards for `<`), or earlier depending on the leading whitespace behavior
  * Set the "open element" count to 1
  * Scan forward in the input buffer
    * If the end element start token `</` is encountered, decrement the open element count and seek to the element end token `>`
    * If the start element start token `<` is encountered, increment the open element count
    * If the self closing token `/>` is encountered, decrement the open element count
    * If the open element count is `0`, process the trailing whitespace behavior
* In order to avoid having to use `flags` settings to disable and enable `conditional` behavior for things like the default property groups, references to unresolved symbols should not be treated as false-y (as they are in the C++ style condition evaluator), instead, the expression (and originally matched `conditional` token) should be emitted to the output stream if the expression contains any unresolved symbols

Ex.
``` XML
<root>
  <element if="'$(variable)' == 'value'"/>
  <element if="'$(variable)' != 'value'">
    <child>
        <grandchild />
    </child>
  </element>
</root>
```

If `variable` == `"value"`, the output would be
``` XML
<root>
  <element />
</root>
```

If `variable` != `"value"`, the output would be
``` XML
<root>
  <element>
    <child>
        <grandchild />
    </child>
  </element>
</root>
```

mlorbetske:
@sayedihashimi @mhutch agree/thoughts?

mlorbetske:
The parts that #27 does not address:
[...]

PR #27, opened 2016-09-12
Support for inline conditions in markup

mlorbetske:
Addresses most of #25. Only bit left is the actual MSBuild evaluator part.


src/Microsoft.TemplateEngine.Core/Operations/InlineMarkupConditional.cs
```diff
@@ -0,0 +1,233 @@
...
+            }
+
+            public string Id { get; }
+
+            public IReadOnlyList<byte[]> Tokens { get; }
+
+            public int HandleMatch(IProcessorState processor, int bufferLength, ref int currentBufferPosition, int token, Stream target)
+            {
+                bool flag;
+                if (processor.Config.Flags.TryGetValue("conditionals", out flag) && !flag)
```
seancpeters: "conditionals" is not the correct flag here. We Need a new one for this operation.
Actually, the checking of these flags was partly moved to ProcessorState.Run(), but apparently the checking was not completely removed from the operations.

seancpeters:
Very interesting. Looks good to me, but I'll leave it to you to merge - I think you'd said you wanted to add more to it.

src/Microsoft.TemplateEngine.Core/Operations/InlineMarkupConditional.cs
mlorbetske: This is by design, as it's the same family of operations as regular conditionals - but I agree it's a bit odd. The reason for internalizing the global flag checks is for the flags operator to determine whether it's got the turn flags on flag (when flags processing is currently off)

src/Microsoft.TemplateEngine.Core/Expressions/MSBuild/MSBuildStyleEvaluatorDefinition.cs
```diff
@@ -0,0 +1,120 @@
...
+            {Tokens.GreaterThanOrEqualTo, Operators.GreaterThanOrEqualTo},
+            {Tokens.LessThan, Operators.LessThan},
+            {Tokens.LessThanOrEqualTo, Operators.LessThanOrEqualTo},
+            {Tokens.EqualTo, Operators.EqualTo},
+            {Tokens.NotEqualTo, Operators.NotEqualTo},
+        };
+
+        public static bool MSBuildStyleEvaluator(IProcessorState processor, ref int bufferLength, ref int currentBufferPosition, out bool faulted)
+        {
+            ITokenTrie tokens = GetSymbols(processor);
```
mlorbetske: How to handle the various operators will need to come from here in somehow so that things like handling of XML entities/escape sequences can be handled as appropriate

src/Microsoft.TemplateEngine.Core/CommonOperators.cs
```diff
@@ -0,0 +1,203 @@
...
+        }
+
+        private static IEvaluable CreateGreaterThanOrEqualTo(IEvaluable active)
+        {
+            return CreateBinaryChild(active, Operators.GreaterThanOrEqualTo, x => false, (l, r) => ((IComparable)l).CompareTo(r) >= 0);
+        }
+
+        private static IEvaluable CreateLessThan(IEvaluable active)
+        {
+            return CreateBinaryChild(active, Operators.LessThan, x => false, (l, r) => ((IComparable)l).CompareTo(r) > 0);
```
seancpeters: < 0

src/Microsoft.TemplateEngine.Core/CommonOperators.cs
```diff
@@ -0,0 +1,203 @@
...
+        }
+
+        private static IEvaluable CreateLessThanOrEqualTo(IEvaluable active)
+        {
+            return CreateBinaryChild(active, Operators.LessThanOrEqualTo, x => false, (l, r) => ((IComparable)l).CompareTo(r) >= 0);
```
seancpeters: <= 0

src/Microsoft.TemplateEngine.Core/Expressions/MSBuild/MSBuildStyleEvaluatorDefinition.cs
```diff
@@ -0,0 +1,120 @@
...
+                // quotes
+                trie.AddToken(processor.Encoding.GetBytes("'"));
+                TokenCache[processor.Encoding] = tokens = trie;
+            }
+
+            return tokens;
+        }
+
+        private enum Tokens
+        {
```
seancpeters: This should have a "None" defined.

src/Microsoft.TemplateEngine.Core/CommonOperators.cs
```diff
@@ -0,0 +1,203 @@
+﻿using System;
+using System.Collections.Generic;
+using Microsoft.TemplateEngine.Core.Expressions;
+
+namespace Microsoft.TemplateEngine.Core
+{
+    public enum Operators
+    {
```
seancpeters: This should have a "None" defined (or some other name with similar semantic meaning).

src/Microsoft.TemplateEngine.Core/CommonOperators.cs
mlorbetske: oops

src/Microsoft.TemplateEngine.Core/CommonOperators.cs
mlorbetske: same oops

src/Microsoft.TemplateEngine.Core/Expressions/MSBuild/MSBuildStyleEvaluatorDefinition.cs
mlorbetske: The case for "None" was around the quote matching - in this use case (in the evaluator), the active quoting character is a nullable token that indicates either the character that must be matched to terminate the run or null to indicate that a quoted run is not in progress

mlorbetske:
The last open item before this should be merged is to emit the entire expression in the event that a symbol is not resolved

src/Microsoft.TemplateEngine.Core/Expressions/MSBuild/MSBuildStyleEvaluatorDefinition.cs
```diff
@@ -0,0 +1,397 @@
...
+                //EOLs
+                trie.AddToken(processor.Encoding.GetBytes("\r\n"));
+                trie.AddToken(processor.Encoding.GetBytes("\n"));
+                trie.AddToken(processor.Encoding.GetBytes("\r"));
+
+                // quotes
+                trie.AddToken(processor.Encoding.GetBytes("'"));
+
+                // variable start
+                trie.AddToken(processor.Encoding.GetBytes("$("));
```
mlorbetske: Not in love with this, looking for ideas (please comment). This works as follows:

1) All variables in this document are known to be of the form $(FOO)
2) All variables are part of the trie and therefore will be consumed as a whole unit
3) In the event that a free $( is found, it's not part of a known symbol (binds to the BadSyntax token type)
4) The BadSyntax token is checked for in the scope builder and rejects the expression if encountered
5) If the builder rejects the expression, all of the bytes for the expression are written back to the target (via a callback which presumably also emits anything else that has been excluded from the target stream pending the result of the evaluation)

Issue #37, opened 2016-09-21
Make the config reading process more uniform


mlorbetske:
Work for this is in progress in #28.

SimpleConfig and GlobalRunSpec are hard to follow and deal directly with the JSON for determining what to do. Fixing this involves:
* Creating a clear progression in which parameters/variables are populated
* The configuration of operations and macros are standardized (no special calls to configure a specific operation or macro - they should be delegated to the relevant read logic supplied by the operation or macro provider)
* Enhancements such as #26 and #24 should be more intuitive to add
* Adding custom configuration for both the standard and the special (file glob specific) operations, including: Custom operations, custom variable formatting, custom flags.

Issue #26, opened 2016-09-12
Support "condition" on top level sources


mlorbetske:
Sources can have conditioned modifiers, but sources themselves should support conditions as well

seancpeters:
Conditions in modifiers was added in:
https://github.com/dotnet/templating/pull/28/commits/94fdaa4c93cc494c4064c6bd328eefe57076a17b
#28

seancpeters:
I misunderstood the requirement, reopening. Will be fixed in a PR soon.

seancpeters:
This commit addresses the issue:
The nginx.conf and robots.txt setup is complete with https://github.com/dotnet/templating/pull/39/commits/d33aa93c0963a099f71613fb2def35e9eb58d90e

Issue #24, opened 2016-09-11
Add something along the lines of "outputs" or "artifacts" to the config file


mlorbetske:
There will need to be some kind of a notion of what the result of a creation is so that consuming systems can add the artifact to the thing it was generated into (if required)

Issue #4, opened 2016-08-26
Update SimpleConfig to have better defaults for known extensions


mlorbetske:
See https://github.com/ligershark/microsoft-templateengine/issues/56 for what the additional defaults should be

seancpeters:
dotnet/templating@1c0d078 provides setup for nginx.conf and batch file style comments.

seancpeters:
The nginx.conf and robots.txt setup is complete with https://github.com/dotnet/templating/pull/39/commits/d33aa93c0963a099f71613fb2def35e9eb58d90e

PR #39, opened 2016-09-22
More custom configuration

seancpeters:
- Custom operation setup
- Custom variable formats
- Custom flag prefix
- New fileglobs

### DIFF
--- a/src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/GlobalRunSpec.cs
+++ b/src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/GlobalRunSpec.cs
@@ -47,6 +47,8 @@ namespace Microsoft.TemplateEngine.Orchestrator.RunnableProjects
 
         public IReadOnlyDictionary<string, string> Rename { get; }
 
+        public string PlaceholderFilename { get; }
+
         public bool TryGetTargetRelPath(string sourceRelPath, out string targetRelPath)
         {
             return Rename.TryGetValue(sourceRelPath, out targetRelPath);
@@ -59,7 +61,8 @@ namespace Microsoft.TemplateEngine.Orchestrator.RunnableProjects
             IParameterSet parameters, 
             IVariableCollection variables,
             IGlobalRunConfig globalConfig,
-            IReadOnlyList<KeyValuePair<string, IGlobalRunConfig>> fileGlobConfigs)
+            IReadOnlyList<KeyValuePair<string, IGlobalRunConfig>> fileGlobConfigs,
+            string placeholderFilename)
         {
             EnsureOperationConfigs(componentManager);
 
@@ -67,6 +70,7 @@ namespace Microsoft.TemplateEngine.Orchestrator.RunnableProjects
             CopyOnly = SetupPathInfoFromSource(source.CopyOnly);
             Exclude = SetupPathInfoFromSource(source.Exclude);
             Rename = source.Rename ?? new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+            PlaceholderFilename = placeholderFilename;
 
             // regular operations
             RootVariableCollection = variables;
--- a/src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/GlobalRunSpec.cs
+++ b/src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/GlobalRunSpec.cs
@@ -56,7 +56,8 @@ namespace Microsoft.TemplateEngine.Orchestrator.RunnableProjects
             return Rename.TryGetValue(sourceRelPath, out targetRelPath);
         }
 
-        public GlobalRunSpec(FileSource source,
+        public GlobalRunSpec(ITemplateEngineHost host,
+            FileSource source,
             IDirectory templateRoot,
             IComponentManager componentManager,
             IParameterSet parameters, 
@@ -73,7 +74,7 @@ namespace Microsoft.TemplateEngine.Orchestrator.RunnableProjects
 
             // regular operations
             RootVariableCollection = variables;
-            Operations = ResolveOperations(operations, componentManager, templateRoot, variables, parameters);
+            Operations = ResolveOperations(host, operations, componentManager, templateRoot, variables, parameters);
 
             // special operations
             Dictionary<IPathMatcher, IRunSpec> specials = new Dictionary<IPathMatcher, IRunSpec>();
@@ -87,7 +88,7 @@ namespace Microsoft.TemplateEngine.Orchestrator.RunnableProjects
 
                     if (specialEntry.Value != null)
                     {
-                        specialOps = ResolveOperations(specialEntry.Value, componentManager, templateRoot, variables, parameters);
+                        specialOps = ResolveOperations(host, specialEntry.Value, componentManager, templateRoot, variables, parameters);
                         specialVariables = VariableCollection.SetupVariables(parameters, specialEntry.Value.VariableSetup);
                     }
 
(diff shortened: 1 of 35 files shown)


src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/GlobalRunSpec.cs
```diff
@@ -111,6 +149,51 @@ private static IReadOnlyList<IOperationProvider> SetupOperations(IComponentManag
...
+        private static IReadOnlyList<IOperationProvider> SetupCustomOperations(IReadOnlyList<ICustomOperationModel> customModel, IComponentManager componentManager, IDirectory templateRoot, IVariableCollection variables, IParameterSet parameters)
+        {
+            List<IOperationProvider> customOperations = new List<IOperationProvider>();
+
+            foreach (ICustomOperationModel opModelUntyped in customModel)
+            {
+                CustomOperationModel opModel = opModelUntyped as CustomOperationModel;
+                if (opModel == null)
+                {
+                    continue;   // TODO: decide if this is ok, or if we need some other handling.
```
mlorbetske: if the host interface could be accessible here - this would be a good case to issue a warning back to the broker for

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/SimpleConfigModel.cs
```diff
@@ -189,52 +189,130 @@ IReadOnlyList<FileSource> IRunnableProjectConfig.Sources
...
+                    defaultSpecials.Add(new SpecialOperationConfigParams("**/*.cs", "//", ConditionalType.CNoComments));
+                    defaultSpecials.Add(new SpecialOperationConfigParams("**/*.cpp", "//", ConditionalType.CNoComments));
+                    defaultSpecials.Add(new SpecialOperationConfigParams("**/*.h", "//", ConditionalType.CNoComments));
+                    defaultSpecials.Add(new SpecialOperationConfigParams("**/*.hpp", "//", ConditionalType.CNoComments));
+                    defaultSpecials.Add(new SpecialOperationConfigParams("**/*.*proj", "<!--/", ConditionalType.Xml));
+                    defaultSpecials.Add(new SpecialOperationConfigParams("**/*.*htm", "<!--", ConditionalType.Xml));
+                    defaultSpecials.Add(new SpecialOperationConfigParams("**/*.*html", "<!--", ConditionalType.Xml));
+                    defaultSpecials.Add(new SpecialOperationConfigParams("**/*.jsp", "<!--", ConditionalType.Xml));
+                    defaultSpecials.Add(new SpecialOperationConfigParams("**/*.asp", "<!--", ConditionalType.Xml));
+                    defaultSpecials.Add(new SpecialOperationConfigParams("**/*.aspx", "<!--", ConditionalType.Xml));
```
mlorbetske: Would this be a good time to expand on the defaults with the ones mentioned here https://github.com/ligershark/microsoft-templateengine/issues/56#issuecomment-240029851? Missing the batch/config/robots.txt configs at least

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/SimpleConfigModel.cs
```diff
@@ -189,52 +189,130 @@ IReadOnlyList<FileSource> IRunnableProjectConfig.Sources
...
+
+                    List<KeyValuePair<string, IGlobalRunConfig>> specialOperationConfig = new List<KeyValuePair<string, IGlobalRunConfig>>();
+
+                    // put the custom configs first in the list
+                    HashSet<string> processedGlobs = new HashSet<string>();
+
+                    foreach (ICustomFileGlobModel customGlobModel in SpecialCustomSetup)
+                    {
+                        SpecialOperationConfigParams defaultParams = defaultSpecials.Where(x => x.Glob == customGlobModel.Glob).FirstOrDefault();
```
mlorbetske: Could OrderedDictionary help us out here (not sure if it's available in our target runtime)? We'd get order preservation and not need as much Linq/iteration

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/SimpleConfigModel.cs
```diff
@@ -189,52 +189,130 @@ IReadOnlyList<FileSource> IRunnableProjectConfig.Sources
...
-                            ["**/*.aspx"] = ProduceOperationSetup("<!--", false, ConditionalType.Xml),
-                        };
-                    _specialOperationConfig = operationSpecials;
+                            // the custom config is for a default-defined glob. Producing the setup uses the custom info.
+                            runConfig = ProduceOperationSetup(defaultParams.SwitchPrefix, false, defaultParams.ConditionalStyle, customGlobModel);
+                        }
+                        else
+                        {
+                            // the custom config is not for a known glob type. Only use the custom info
+                            runConfig = ProduceOperationSetup(string.Empty, false, ConditionalType.None, customGlobModel);
```
mlorbetske: The switch prefix will have to come from the config so that we don't break the language of whatever the custom pattern matches

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/SimpleConfigModel.cs
```diff
@@ -517,6 +606,32 @@ public static SimpleConfigModel FromJObject(JObject source)
...
+            {
+                config.CustomOperations = CustomFileGlobModel.FromJObject((JObject)globalCustomConfigData, string.Empty);
+            }
+            else
+            {
+                config.CustomOperations = null;
+            }
+
+            // Custom operations for specials
+            IReadOnlyDictionary<string, JToken> allSpecialOpsConfig = source.ToJTokenDictionary(StringComparer.OrdinalIgnoreCase, "SpecialCustomOperations");
```
mlorbetske: I wonder if we should call specials something more descriptive of their nature - in the config at least. Haven't thought of a good, concise way of expressing "configuration for the actions that will be taken on files whose paths match a pattern", so I've got no suggestion

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/GlobalRunSpec.cs
seancpeters: Makes sense.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/SimpleConfigModel.cs
seancpeters: The prefix gets resolved in ProduceOperationSetup(). I've reworked that method a little, including its input params so it's less misleading as to what happens with the switch prefix. Note also, I'm slowly changing it to "flag prefix" as I find instances of switch prefix.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/SimpleConfigModel.cs
seancpeters: I've also been trying to come up with a better name for these things. I've started to use "FileGlob" as part of the names of some related symbols. So for specials, the best rename I've come up with is FileGlobSpecificOperations or just FileGlobOperations. I'm not 100% satisfied with that name, so haven't made changes yet.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/SimpleConfigModel.cs
seancpeters: In the latest push to this PR, I implemented a new built-in conditional format, signified by ConditionalType.HashSignLineComment. It's not wired up to any fileglobs yet, but there are a number of unit tests which exercise it.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/SimpleConfigModel.cs
seancpeters: It's not available, and I'm not sure if it would actually help performance since it doesn't have a proper TryGetValue - it's not clear to me whether the contains implementation is O(1) or O(n). Also, there is no generic version of OrderedDictionary, so there would be a lot of explicit casting involved in retrieving the values :(

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/SimpleConfigModel.cs
seancpeters: Now it's wired up

mlorbetske:
This is starting to get pretty big & spans several issues, let's get it merged as soon as possible and go for smaller, more targeted PRs after this

seancpeters:
This PR takes care of many issues, including #37, #26, #24, and #4

seancpeters:
I had meant to merge, not close. Will merge when the checks complete.

PR #41, opened 2016-09-23
Globbing

### LINKED ISSUE #40: File Glob matching is too greedy
The matches are not anchored and so can match partial paths. Possibly other issues with overlap between what are meant to be disjoint match sets.

mlorbetske:
Fixes #40

### DIFF
--- a/src/Microsoft.TemplateEngine.Utils/Glob.cs
+++ b/src/Microsoft.TemplateEngine.Utils/Glob.cs
@@ -1,0 +1,275 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.TemplateEngine.Utils
+{
+    public class Glob
+    {
+        private readonly IReadOnlyList<IMatcher> _matchers;
+
+        private Glob(IReadOnlyList<IMatcher> matchers)
+        {
+            _matchers = matchers;
+        }
+
+        private interface IMatcher
+        {
+            int MinConsume { get; }
+
+            bool ProducesCheckpoint { get; }
+
+            bool CanConsume(string test, int startAt, out int endPosition);
+        }
+
+        public static Glob Parse(string pattern)
+        {
+            List<IMatcher> matchers = new List<IMatcher>();
+
+            for (int i = 0; i < pattern.Length; ++i)
+            {
+                switch (pattern[i])
+                {
+                    case '[':
+                        List<char> values = new List<char>();
+                        for (; i < pattern.Length; ++i)
+                        {
+                            if (pattern[i] == ']')
+                            {
+                                break;
+                            }
+
+                            values.Add(pattern[i]);
+                        }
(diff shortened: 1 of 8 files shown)


src/Microsoft.TemplateEngine.Utils/Glob.cs
```diff
@@ -0,0 +1,275 @@
...
+                            }
+
+                            values.Add(pattern[i]);
+                        }
+                        matchers.Add(new LiteralMatcher(values));
+                        break;
+                    case '*':
+                        if (pattern.Length > i + 1 && pattern[i + 1] == '*')
+                        {
+                            if (i < pattern.Length - 2 && pattern[i + 2] == '/')
```
seancpeters: Maybe use:
   if (pattern.Length > i + 2)
to be consistent with the previous if.

src/Microsoft.TemplateEngine.Utils/Glob.cs
```diff
@@ -0,0 +1,275 @@
...
+
+        public bool IsMatch(string test)
+        {
+            Stack<Checkpoint> checkpoints = new Stack<Checkpoint>();
+
+            int currentMatcher = 0;
+            int i = 0;
+
+            //See if we can just do a name match
+            if (!_matchers.Any(x => x is PathMatcher || x is ExactPathMatcher || (x as LiteralMatcher)?.Char?.FirstOrDefault() == '/'))
```
seancpeters: This could be evaluated once in the constructor and the result stored in a private bool, to be checked here, instead of rechecking each time.

src/Microsoft.TemplateEngine.Utils/Glob.cs
```diff
@@ -0,0 +1,275 @@
...
+            public Checkpoint(IMatcher matcher, int nextMatcher, int stringPosition)
+            {
+                Matcher = matcher;
+                NextMatcher = nextMatcher;
+                StringPosition = stringPosition;
+            }
+
+            public IMatcher Matcher { get; }
+
+            public int NextMatcher { get; }
```
seancpeters: nit: NextMacthIndex might make it more intuitive that this isn't a matcher. Due to the similarity of the names "Matcher" & "NextMatcher", I assumed they stored the same type of thing until I looked at the implementation.

src/Microsoft.TemplateEngine.Utils/Glob.cs
mlorbetske: Done

src/Microsoft.TemplateEngine.Utils/Glob.cs
mlorbetske: Done

src/Microsoft.TemplateEngine.Utils/Glob.cs
mlorbetske: Done

Issue #51, opened 2016-09-29
Update the C++ evaluator to not require paren matching for end detection


mlorbetske:
The general purpose expression evaluator should be used for this as well - bitwise operators & value comparisons will need to be introduced

mlorbetske:
The current issue here (blocking wholesale adoption of Cpp2StyleEvaluatorDefinition.cs) is that it currently fails to appropriately detect condition ends in markup languages in cases where the markup closing tags are also valid operators/operands in the evaluation language.

Ex.
`<!--#if (TEST) -->` will produces a tree like `(((Identity(TRUE) minus ()) minus ()) greaterThan ())`

Ex2.
`@* #if (TEST) *@` will produces a tree like `(Identity(TRUE) multiply '@')`

grinrag:
This bug is outdated (original date is Sep 2016), closing it now. Feel free reopen if it still valid.

PR #55, opened 2016-10-01
C++ evaluator using the general evaluator

mlorbetske:
Partially addresses #51, some of the tests for Conditional fail when it uses this - those will need to be investigated before the new evaluator can be made the default and the issue closed.


src/Microsoft.TemplateEngine.Core/Expressions/Cpp2/Cpp2StyleEvaluatorDefinition.cs
```diff
@@ -0,0 +1,349 @@
...
+                    {
+                        ld = lh;
+                    }
+                    else
+                    {
+                        return null;
+                    }
+                }
+            }
+
```
seancpeters: This method is difficult to follow due to the short variable names. 
Also, the code within the "if (!leftIsDouble)" and within (!rightIsDouble) is duplicated.

mlorbetske: This is somewhat fixed in the next round of changes, left and right are still duplicated but just call the core conversion methods

Microsoft.TemplateEngine.sln
```diff
@@ -1,31 +1,29 @@
...
 EndProject
-Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Microsoft.TemplateEngine.Abstractions", "src\Microsoft.TemplateEngine.Abstractions\Microsoft.TemplateEngine.Abstractions.xproj", "{0C92EDA7-492F-4CBA-9F36-61932CA5C1F4}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "dotnet-new3", "src\dotnet-new3\dotnet-new3.csproj", "{76D8C970-B5CA-4364-B1E3-45D771922F74}"
 EndProject
-Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Microsoft.TemplateEngine.Orchestrator.RunnableProjects", "src\Microsoft.TemplateEngine.Orchestrator.RunnableProjects\Microsoft.TemplateEngine.Orchestrator.RunnableProjects.xproj", "{578CE255-E412-4CC7-9A03-1EDCA522DA27}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.TemplateEngine.Abstractions", "src\Microsoft.TemplateEngine.Abstractions\Microsoft.TemplateEngine.Abstractions.csproj", "{0C92EDA7-492F-4CBA-9F36-61932CA5C1F4}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "dotnet-new3.VSIX", "src\dotnet-new3.VSIX\dotnet-new3.VSIX.csproj", "{848010BF-F566-4EE1-9CA8-F3B8DFFF3411}"
```
seancpeters: dotnet-new3.VSIX is removed, not replaced. Might not be an issue.

mlorbetske: This is to enable solution build from the .net cli. We can always build that project directly where applicable.

Issue #52, opened 2016-09-29
Propagate the host object everywhere needed to report configuration anomalies.


seancpeters:
There are many places in the code where invalid config is quietly ignored. It would be quite helpful to users to get diagnostic info about their config.

seancpeters:
Update:
The host is now accessible via EngineEnvironmentSettings, so this task becomes a matter of reporting config issues to the host, with no need to pass the host around everywhere.

Issue #46, opened 2016-09-28
Localization support (Runnable Project Templates)


mlorbetske:
High level summary of the current thinking:
* The host interface should always give a way of getting the current locale
* .netnew.json will have locale specific flavors (en_us.netnew.json, es_es.netnew.json, etc.)
  * `*.netnew.json` should never be emitted as content
  * All display properties should be taken from `(locale).netnew.json` if available, if a matching locale or property is missing, it should be taken from `.netnew.json`
* Each `(locale).netnew.json` should contain a `localizations` property whose value is an object
  * The property names in that object should be the full file paths in the template to the content containing localizations
  * The values of those properties should be objects
    * The properties of those objects should be the unlocalized value
    * The values of those properties should be the localized value that should be used in place of the unlocalized value
* The template display info cache should be differentiated on language (a cache for en_us, fr_fr, etc. instead of the one cache we currently have)
  * The appropriate localized display properties should be used when computing the locale specific cache
* At template instantiation, the normal matching rules should apply with the following changes
  * If no `(locale).netnew.json` applies to the current instantation, nothing changes
  * If one does, for each file to be processed:
    * The unlocalized -> localized value pairs that apply to the file (from `(locale).netnew.json`) should become a set of `Replacement` operations
    * If the file would be processed anyway the replacements should be added to the set of things that would be done to the file (clone and update the operation trie, add the operations to the set, etc.)
    * If the file wouldn't be processed (matches `copyOnly`), the file now needs to be processed, but only using the replacement operations without any other operations that would apply to the file based on any other pattern matches

seancpeters:
Handled by #57

PR #56, opened 2016-10-03
Localization processing in progress.

seancpeters:
Doesn't read the locale.netnew.json yet, it's temporarily reading from the
.netnew.json.
Reads correcly, but wrong place.
TODO: determine/provide the right file location


seancpeters:
Partial work on issue #46 and issue #52

seancpeters:
Closing for real now, after having merged.

PR #64, opened 2016-11-21
Temp fix to a unit test until we can really fix the newline problem.

seancpeters:
Newline consumption on conditionals needs work, so we don't emit extra newlines.


seancpeters:
Merging as-is. Let's revisit the tests when they're in a more easily testable state.

PR #75, opened 2016-12-13
Updating a template will no longer lose its previously installed localization

seancpeters:


template_feed/Microsoft.DotNet.MVC.Templates.1.0.0/content/WebApplication/.template.config/template.json
```diff
@@ -0,0 +1,53 @@
...
+          ]
+        }
+      ]
+    }
+  ],
+  "symbols": {
+    "auth": {
+      "type": "parameter",
+      "defaultValue": "None",
+      "description": "The type of authentication to use"
```
mlorbetske: Need to add choices for this

PR #76, opened 2016-12-16
Setup host & version specifc caches

seancpeters:
probably has a lot of old junk due to git craziness.


src/Microsoft.TemplateEngine.Abstractions/ITemplateEngineHost.cs
```diff
@@ -11,6 +11,10 @@ public interface ITemplateEngineHost
 
         string HostIdentifier { get; }
 
+        string AssemblyName { get; }
```
mlorbetske: Let's remove this one

src/Microsoft.TemplateEngine.Abstractions/ITemplateEngineHost.cs
```diff
@@ -11,6 +11,10 @@ public interface ITemplateEngineHost
...
+
+        Version AssemblyVersion { get; }
```
mlorbetske: Let's rename this one to just `Version`

src/Microsoft.TemplateEngine.Utils/EngineEnvironmentSettings.cs
```diff
@@ -13,10 +13,12 @@ namespace Microsoft.TemplateEngine.Utils
     public static class EngineEnvironmentSettings
     {
         private static readonly string DefaultLocale = "en_US";
+        private static readonly string DefaultAssemblyName = "None";
+        private static readonly Version DefaultAssemblyVersion = new Version("0.0.0.1");
 
         static EngineEnvironmentSettings()
         {
-            Host = new DefaultTemplateEngineHost(string.Empty, DefaultLocale);
+            Host = new DefaultTemplateEngineHost(string.Empty, DefaultLocale, DefaultAssemblyName, DefaultAssemblyVersion);
```
mlorbetske: This needs to be removed I think. The actual host should do a `EngineEnvironmentSettings.Host = (instance that the host provides here)`

src/Microsoft.TemplateEngine.Utils/EngineEnvironmentSettings.cs
```diff
@@ -57,7 +59,7 @@ public string BaseDir
                 {
                     if (_baseDir == null)
                     {
-                        _baseDir = Path.Combine(UserProfileDir, ".netnew");
+                        _baseDir = Path.Combine(UserProfileDir, ".netnew", Host.AssemblyName, Host.AssemblyVersion.ToString());
```
mlorbetske: This one should get captured in the renaming/changing of the name/version pair in the host settings

src/Microsoft.TemplateEngine.Utils/EngineEnvironmentSettings.cs
mlorbetske: Actually, we already do that in Program.cs, but we shouldn't have the default I don't think. This must always come from the host.

seancpeters: Agree, this should go away. It was a cheap way to get around the problem of not always having the locale when the host first needs to be invoked. I added an DefaultTemplateEngineHost.UpdateLocale method to do it more properly.

PR #92, opened 2016-12-18
Update MVC content

mlorbetske:
@phenning


src/dotnet-new3/ExtendedCommandParser.cs
```diff
@@ -56,7 +56,7 @@ public void ShowHelp()
             _helpDisplayer.ShowHelp();
         }
 
-        public void RemoveOption(CommandOption option)
+        internal void RemoveOption(CommandOption option)
```
seancpeters: I was thinking we'd eventually make this class more generalized - not specific to templates. IF we do, then we'd probably want these all to be public, as this would be in a separate library. But for now I can see the merit of internal.

seancpeters:
Looks good to me

PR #119, opened 2016-12-21
Allow for installing templates using restore. DO NOT MERGE YET

seancpeters:


seancpeters:
Everything in here was migrated to #120. Closing without merging.

PR #120, opened 2016-12-21
Installer & build cleanup

mlorbetske:
Add nuget package installer based on #119
Make failing builds fail faster
Apply fix to debug:attach from #119
Add packages folder
Rename .netnew folder to .templateengine

### DIFF
--- a/src/dotnet-new3/Installer.cs
+++ b/src/dotnet-new3/Installer.cs
@@ -1,0 +1,126 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+using Microsoft.DotNet.Cli.Utils;
+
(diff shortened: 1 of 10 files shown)


src/dotnet-new3/Installer.cs
```diff
@@ -0,0 +1,126 @@
...
+                newLocalPackages.Add(path);
+            }
+
+            Paths.User.ScratchDir.DeleteDirectory();
+            InstallLocalPackages(newLocalPackages);
+        }
+
+        private void InstallLocalPackages(IReadOnlyList<string> packageNames)
+        {
+            List<string> toInstall = new List<string>();
```
seancpeters: Nit: not used - we should delete in future PR

PR #134, opened 2016-12-22
Fix Version type and possible nullref HelpFormatter

### LINKED ISSUE #132: Version in ITemplateEngineHost must be a string not a Version
We should accomodate versioning of environments that use semantic versioning to disambiguate their releases rather than a `x.x.x.x` style version. To do that, we'll need to change this property to a string.

mlorbetske:
Fixes #132

/cc: @seancpeters


seancpeters:
:shipit:

PR #197, opened 2017-01-20
Respect the NUGET_PACKAGES env var

seancpeters:


seancpeters:
This is not the correct fix.

PR #202, opened 2017-01-22
Get rid of statics - enable parallel test runs & multiple instantiations of the engine in the same process

mlorbetske:
Right now, quite a few things assume that exactly one template engine instance will be loaded into a process space and be executing at a given time (the concern here is mostly around the first run experience). This is problematic for environments where different configuration hives would/should be used within the same process space (ex. if using this from VS, having two different template caches for two different dialogs). While it is possible (given the assertion that exactly one things could force first run at a given moment is true) to switch the host object in use in the relevant scopes (dialog 1 sets host 1 to `EngineEnvironmentSettings.Host` just before it needs it and dialog 2 sets host 2 just when it needs it), this can introduce extra complexity & difficult to track down bugs in orchestration of this state. This will also enable test parallelization for both the dotnet cli and for this repo (as a different hive can be used for each test - or an ephemeral/in-memory hive can be introduced in test contexts that deal with concurrent first-run requests by directing them to different hives).

This change removes the static qualification from `EngineEnvironmentSettings`, `SettingsLoader`, `Paths`, and `AliasRegistry`. EngineEnvironmentSettings is included on all mount points & passed around to most everything else. SettingsLoader has been added to EngineEnvironmentSettings, Paths and AliasRegistry are now constructed from an EngineEnvironmentSettings rather than accessing it as a static class.

I'm not quite convinced that `Paths` should meet this fate (as the syntax becomes awkward), perhaps we should move it back to being static & pass the environment settings as a param for the calls.

@seancpeters please review carefully, none of the changes were tricky or change architecture significantly, but this was repetitive & could have accidentally (unlikely, as all tests pass) changed some behavior

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -5,6 +5,7 @@ root = true
 
 # Newline ending every file
 [*]
+end_of_line = crlf
 indent_style = space
 indent_size = 4
 tab_width = 4
--- a/Version.props
+++ b/Version.props
@@ -3,7 +3,8 @@
     <PackageReleaseVersion>1.0.0</PackageReleaseVersion>
     <PackageBuildQuality Condition=" '$(PackageBuildQuality)' == '' ">beta1</PackageBuildQuality>
     <PackageVersion>$(PackageReleaseVersion)-$(PackageBuildQuality)</PackageVersion>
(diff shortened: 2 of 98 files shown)


mlorbetske:
Added the support for ephemeral hives to fix #203

PR #221, opened 2017-01-27
Add solution item template, fix netcoreapp versions, fix wrong versio…

mlorbetske:
…n of compression package, update new3 test namespaces


seancpeters:
:shipit:

Issue #237, opened 2017-02-02
XAML IF Conditions


mvegaca:
Hi.

How can I add IF conditions in XAML files?
Thanks

mlorbetske:
@seancpeters can you help out with the syntax for turning on conditionals for `*.XAML` files in `template.json` please? This would be XML block comment + C++ syntax. We should include XAML files (and other XML based formats) in the [engine defaults](https://github.com/dotnet/templating/blob/b96ec9c36f8c1accf6dcc608da66c768ac17ab5f/src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/SimpleConfigModel.cs#L297) as well.

mvegaca:
Hi

I'm trying with the following code but does not work.

![code](https://cloud.githubusercontent.com/assets/5556784/22555804/4a58fa58-e965-11e6-8b56-fb97f60dc8a1.PNG)

seancpeters:
Hi @mvegaca,
To add if / elseif / else conditionals for XAML files in a template, you'll need to add some configuration in the template.json file. This is what should be added:
```
  "SpecialCustomOperations": {
    "**/*.xaml": {
      "operations": [
        {
          "type": "conditional",
          "configuration": {
            "actionableIf": [ "<!--#if" ],
            "actionableElse": [ "#else", "<!--#else" ],
            "actionableElseif": [ "#elseif", "<!--#elseif" ],
            "endif": [ "#endif", "<!--#endif" ],
            "trim" : "true",
            "wholeLine": "true",
          }
        }
      ]
    }
  }
```
SpecialCustomOperations needs to be a top-level section in the file. If you also want Xml file processing, you can add "**/*.xml" configuration under SpecialCustomOperations, using a setup identical to the above XAML setup.

Note that there is no space between the comment and the conditional, for example, you'll want to use:
`<!--#if`
but not use
`<!-- #if`
In the future we will be adding built-in support for XAML, Xml, and many other file types. This configuration will match what we set up for these types of files.

If you're going to have nested if-else-elseif template processing statements in your XAML files, there is a bit of more advanced setup we can do, to help with embedded comments.

mvegaca:
Thank you very much @seancpeters it works fine!

PR #238, opened 2017-02-02
Add more default handling of file types

mlorbetske:
Related to #237

Do not merge yet, let's get some more feedback on whether our default globs match what file extensions we think are common in templates

@seancpeters @phenning @sayedihashimi @madskristensen


mlorbetske:
Closing this as we'll treat these individually with #279

PR #243, opened 2017-02-02
Update to how the environment variable that determines where home is

### LINKED ISSUE #239: UserProfileDir is incorrect when running with Mono on non-Windows platforms
The UserProfileDir is using the USERPROFILE environment variable which is not set on non-Windows platforms:

https://github.com/dotnet/templating/blob/b96ec9c36f8c1accf6dcc608da66c768ac17ab5f/src/Microsoft.TemplateEngine.Utils/EngineEnvironmentSettings.cs#L54

mlorbetske:
Fixes #239

@mrward


seancpeters:
:shipit:

PR #254, opened 2017-02-04
String review for dotnet new

mairaw:
/cc @seancpeters 

I'm wondering if we should say more about the -all option.
Also, for strings like ParameterNamePrefixError, is it correct to say parameter name(s) or should we just say option(s)?

### DIFF
--- a/src/Microsoft.TemplateEngine.Cli/LocalizableStrings.cs
+++ b/src/Microsoft.TemplateEngine.Cli/LocalizableStrings.cs
@@ -5,27 +5,27 @@ namespace Microsoft.TemplateEngine.Cli
 {
     internal class LocalizableStrings
     {
-        public const string ShowsAllTemplates = "Shows all templates";
+        public const string ShowsAllTemplates = "Shows all templates.";
 
         public const string DisplaysHelp = "Displays help for this command.";
 
-        public const string ParameterNamePrefixError = "Parameter names must start with -- or -";
+        public const string ParameterNamePrefixError = "Parameter names must start with -- or -.";
 
-        public const string ParameterReuseError = "Parameter name {0} cannot be used for multiple purposes";
+        public const string ParameterReuseError = "Parameter name {0} cannot be used for multiple purposes.";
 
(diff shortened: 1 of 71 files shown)


mairaw:
Also not sure if it would be better to say "Use the -all flag to force creation." instead of "Force creation with the -all flag."
Let me know what you guys think and I can make the changes, if needed.

seancpeters:
For the parameter help when running dotnet new, I'm leaning towards keeping it relatively short, and having more detailed explanations in other help docs.

seancpeters:
For ParameterNamePrefixError I think the help should use the word "parameter", to be consistent with a lot of the other output messages. We could use "parameter(s)" instead of "parameter name(s)".

src/Microsoft.TemplateEngine.Cli/LocalizableStrings.cs
```diff
@@ -5,27 +5,27 @@ namespace Microsoft.TemplateEngine.Cli
...
-        public const string ParameterMissingValue = "Parameter [{0}] ({1}) must be given a value";
+        public const string ParameterMissingValue = "Parameter [{0}] ({1}) must be given a value.";
 
-        public const string TemplateMalformedDueToBadParameters = "Template is malformed. The following parameter names are invalid: {0}";
+        public const string TemplateMalformedDueToBadParameters = "Template is malformed. The following parameter names are invalid: {0}.";
 
         public const string OptionVariantAlreadyDefined = "Option variant {0} for canonical {1} was already defined for canonical {2}";
 
-        public const string ListsTemplates = "List templates containing the specified name.";
+        public const string ListsTemplates = "Lists templates containing the specified argument. If no argument is specified, lists all templates.";
```
mlorbetske: I think this makes it sound like the switch takes an argument rather than referring to the argument passed prior to any switch specification - it could just be me though

mairaw: Yes, I though the same with the previous string too @mlorbetske. I'll try to come up with something different that makes this a bit more clear.

mlorbetske:
Made these changes via #275, to avoid having to rebase. Thanks for helping us get these right!

mairaw:
No problem @mlorbetske!

PR #259, opened 2017-02-04
Add back -t for compat, rename xunit test alias

mlorbetske:


mlorbetske:
Carries some removals of trailing periods in help & the removal of the content generation timing message, per the doc team as well

mlorbetske:
@blackdwarf updated

src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -91,6 +91,13 @@ public New3Command(string commandName, ITemplateEngineHost host, Action<IEngineE
 
         public string OutputPath => _app.InternalParamValue("--output");
 
+        public string TemplateType => _app.InternalParamValue("--type");
+
+        public string TemplateName
+        {
+            get { return string.IsNullOrEmpty(TemplateType) ? _templateNameArgument.Value : TemplateType; }
```
seancpeters: We should probably have an error condition if the user does something like:
`dotnet new <template name> -t <template name 2>`

I'd say it's an error even if the two template specifications are the same.

PR #279, opened 2017-02-07
Simplified & Unified conditional configuration.

seancpeters:


src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/Config/ConditionalBlockCommentConfig.cs
```diff
@@ -0,0 +1,102 @@
...
+    public static class ConditionalBlockCommentConfig
+    {
+        public static List<IOperationProvider> ConfigureFromJObject(JObject rawConfiguration)
+        {
+            string startComment = rawConfiguration.ToString("startComment");
+            string endComment = rawConfiguration.ToString("endComment");
+
+            if (string.IsNullOrWhiteSpace(startComment) || string.IsNullOrWhiteSpace(endComment))
+            {
+                throw new Exception($"Template authoring error. StartComment and EndComment must be defined");
```
mlorbetske: With the hardening effort - we should really make these custom exception types & catch them at the outer run to make sure no traces are displayed but the right message is conveyed to the author (as these will likely get picked off in testing)

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/Config/ConditionalBlockCommentConfig.cs
```diff
@@ -0,0 +1,102 @@
...
+using Microsoft.TemplateEngine.Core.Operations;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.TemplateEngine.Orchestrator.RunnableProjects.Config
+{
+    public static class ConditionalBlockCommentConfig
+    {
+        public static List<IOperationProvider> ConfigureFromJObject(JObject rawConfiguration)
+        {
+            string startComment = rawConfiguration.ToString("startComment");
```
mlorbetske: I think "comment" is wrong for these in general, as we support reading effectively "running" preprocessors - maybe this should be "startSyntax"/"endSyntax" or something along those lines (similar for single line setups)

seancpeters: Maybe "startToken" / "endToken" ?

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/Config/ConditionalBlockCommentConfig.cs
seancpeters: 👍

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/Config/ConditionalBlockCommentConfig.cs
mlorbetske: that'll work

Issue #304, opened 2017-02-10
Don't print full help when a template can't be created


billwert:
If I try to make a template that does not exist (or can't be created because it is an item template) I get a message and then full help. Given that I typed a correct command just printing the error message (and perhaps list of valid template choices) seems sufficient.

mlorbetske:
cc @sayedihashimi

sayedihashimi:
+1

Issue #303, opened 2017-02-10
--show-all should just list templates without printing the rest of help


billwert:
Printing 15 lines of text unrelated to the thing I asked for seems wasteful.

mlorbetske:
cc @sayedihashimi

sayedihashimi:
+1

PR #307, opened 2017-02-10
Updates to help characteristics

mlorbetske:
Skip showing usage help when picking an out of scope template
Skip listing templates when a template's args are wrong
List templates in context when no template matching the name was found


mlorbetske:
Related to #304 and #303

PR #324, opened 2017-02-15
Cache enhancements

seancpeters:
Tags become choices.
Lots of help changes.


template_feed/Microsoft.DotNet.Common.ItemTemplates/content/Nuget/.template.config/template.json
```diff
@@ -2,12 +2,20 @@
   "author": "Microsoft",
   "classifications": ["Config"],
   "name": "Nuget Config",
-  "tags": {
```
mlorbetske: Thinking about this more, I think we should leave `tags` in our templates for compat with the RTW release of the CLI (we'll be able to install these templates in that environment without updating the engine)

seancpeters: Makes sense.

src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -281,37 +312,56 @@ private void HandlePostActions(TemplateCreationResult creationResult)
...
+        // Checks the result of TemplatesToDisplayInfoAbout()
+        // If they all have the same group identity, return them.
+        // Otherwise retun an empty list.
+        private IEnumerable<ITemplateInfo> TemplatesToShowDetailedHelpAbout()
         {
-            IReadOnlyCollection<ITemplateInfo> results = null;
+            IReadOnlyList<ITemplateInfo> candidateTemplates = TemplatesToDisplayInfoAbout();
 
-            try
+            if (candidateTemplates.Count() == 0)
```
mlorbetske: Since this is an IReadOnlyList, we can use .Count == 0 rather than the extension method

src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -281,37 +312,56 @@ private void HandlePostActions(TemplateCreationResult creationResult)
...
-                {
-                    results = _matchedTemplates.Where(x => x.IsMatch).Select(x => x.Info).ToList();
+            string firstGroupIdentity = candidateTemplates.First().GroupIdentity;
 
-                    if (results.Count == 0)
-                    {
-                        results = _matchedTemplates.Select(x => x.Info).ToList();
-                    }
-                }
+            if (candidateTemplates.Any(x => x.GroupIdentity != firstGroupIdentity))
```
mlorbetske: This would be a handy extension method (for Utils) - something like 

``` C#
public static bool AllAreTheSame<T, TValue>(this IEnumerable<T> items, Func<T, TValue> selector)
    where TValue : IEquatable<TValue>
{
    return items.AllAreTheSame(selector, (x, y) => x?.Equals(y) ?? (y == null));
}

public static bool AllAreTheSame<T, TValue>(this IEnumerable<T> items, Func<T, TValue> selector, IEqualityComparer<TValue> comparer = null)
{
    if (comparer == null)
    {
        comparer = EqualityComparer<TValue>.Default;
    }

    return items.AllAreTheSame(selector, comparer.Equals);
}

public static bool AllAreTheSame<T, TValue>(this IEnumerable<T> items, Func<T, TValue> selector, Func<TValue, TValue, bool> comparer = null)
{
    if (comparer == null)
    {
        comparer = EqualityComparer<T>.Default.Equals;
    }

    using (IEnumerator<T> enumerator  = items.GetEnumerator())
    {
        if (!enumerator.MoveNext())
        {
            return true; //If there are no elements they're all the same - maybe?
        }

        TValue value = selector(enumerator.Current);

        while (enumerator.MoveNext())
        {
            if (!comparer.Equals(value, enumerator.Current))
            {
                return false;
            }
        }

        return true;
    }
}
```

src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -281,37 +312,56 @@ private void HandlePostActions(TemplateCreationResult creationResult)
...
-                return;
+                templateList = _matchedTemplatesWithSecondaryMatchInfo.Select(x => x.Info);
+            }
+            else if (_matchedTemplates != null && _matchedTemplates.Where(x => x.IsMatch).Count() > 0)
+            {
+                templateList = _matchedTemplates.Where(x => x.IsMatch).Select(x => x.Info);
+            }
+            else
+            {
+                templateList = PerformAllTemplatesInContextQueryAsync().Result.Where(x => x.IsMatch).Select(x => x.Info);
```
mlorbetske: We should propagate the async'ness out of this rather than `.Result`ing it

src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -323,14 +373,19 @@ private void DisplayTemplateList(bool showAll = false)
 
                 foreach (ITemplateInfo info in grouping)
                 {
-                    if (info.Tags != null && info.Tags.TryGetValue("language", out string lang))
+                    if (info.Tags != null && info.Tags.TryGetValue("language", out ICacheTag languageTag))
                     {
-                        if (string.IsNullOrEmpty(Language) && string.Equals(defaultLanguage, lang, StringComparison.OrdinalIgnoreCase))
+                        foreach (string lang in languageTag.ChoicesAndDescriptions.Keys)
```
mlorbetske: I assume that `ChoicesAndDescriptions` can never be null? Haven't made it to where it's populated yet

src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -966,30 +1095,37 @@ private void ShowInvocationExamples()
...
 
-        private void ShowTemplateHelp()
+        private void ShowTemplateHelp(ITemplateInfo templateInfo)
         {
-            ITemplateInfo tmplt = UnambiguousTemplateToUse.Info;
+            IEnumerable<string> languages = null;
+
+            if (templateInfo.Tags != null && templateInfo.Tags.TryGetValue("language", out ICacheTag languageTag))
+            {
+                languages = languageTag.ChoicesAndDescriptions.Keys.Where(x => !string.IsNullOrWhiteSpace(x));
```
mlorbetske: This gets enumerated multiple times

src/Microsoft.TemplateEngine.Edge/Settings/TemplateInfo.cs
```diff
@@ -31,18 +32,86 @@ public TemplateInfo(JObject entry)
...
-                foreach (JProperty item in tagsObject.Properties())
+
+            // parse the cached tags
+            Dictionary<string, ICacheTag> tags = new Dictionary<string, ICacheTag>();
+            JObject tagsObject = (JObject)entry[nameof(Tags)];
+            foreach (JProperty item in tagsObject.Properties())
+            {
+                Dictionary<string, string> choicesAndDescriptions = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+                JObject cdToken = (JObject)item.Value[nameof(ICacheTag.ChoicesAndDescriptions)];
+                foreach (JProperty cdPair in cdToken.Properties())
```
mlorbetske: This should probably use the helper to make sure we're reading case insensitively & have null checks in case we're reading a cache that doesn't have this new property

src/Microsoft.TemplateEngine.Edge/Settings/TemplateInfo.cs
```diff
@@ -31,18 +32,86 @@ public TemplateInfo(JObject entry)
...
+            Tags = tags;
+
+            // parse the cached params
+            JObject cacheParamsObject = (JObject)entry[nameof(CacheParameters)];
+            Dictionary<string, ICacheParameter> cacheParams = new Dictionary<string, ICacheParameter>();
+            foreach (JProperty item in cacheParamsObject.Properties())
+            {
+                ICacheParameter param = new CacheParameter
+                {
+                    DataType = item.Value[nameof(ICacheParameter.DataType)].ToString(),
```
mlorbetske: Same for these for case sensitivity

src/Microsoft.TemplateEngine.Edge/Template/MatchKind.cs
```diff
@@ -6,6 +6,8 @@ public enum MatchKind
         Unspecified,
         Exact,
         Partial,
-        Mismatch
+        Mismatch,
+        InvalidParameterName,
```
mlorbetske: Are these worse than mismatch?

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/RunnableProjectGenerator.cs
```diff
@@ -363,6 +362,24 @@ internal static object DataTypeSpecifiedConvertLiteral(IEngineEnvironmentSetting
...
 
+        private static bool TryResolveChoiceValue(string literal, ITemplateParameter param, out string match)
+        {
+            if (literal == null)
+            {
+                match = null;
+                return false;
+            }
+
+            match = param.Choices.Keys.FirstOrDefault(x => string.Equals(x, literal, StringComparison.OrdinalIgnoreCase));
```
mlorbetske: `Choices` should be hydrated with a case insensitive dictionary so that this can become `ContainsKey`

mlorbetske: feels like this should be `param.Choices.ContainsKey(literal) ? literal : null` if `Choices` can be a case insensitive dictionary

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/RunnableProjectGenerator.cs
```diff
@@ -363,6 +362,24 @@ internal static object DataTypeSpecifiedConvertLiteral(IEngineEnvironmentSetting
...
+
+            if (match == null && param.Choices.Keys.Count(x => x.StartsWith(literal, StringComparison.OrdinalIgnoreCase)) == 1)
```
mlorbetske: Another extension method candidate that can perform the iteration once:

``` C#
public static bool TryGetOnlyValue<T>(this IEnumerable<T> items, Func<T, bool> predicate, out T match)
{
    bool found = false;
    match = default(T);

    foreach (T item in items)
    {
        if (predicate(item))
        {
            if (found)
            {
                return false;
            }

            found = true;
            match = item;
        }
    }

    return found;
}
```

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/RunnableProjectTemplate.cs
```diff
@@ -23,6 +24,7 @@ public RunnableProjectTemplate(JObject raw, IGenerator generator, IFile configFi
             ShortName = config.ShortName;
             Author = config.Author;
             Tags = config.Tags;
+            CacheParameters = config.CacheParameters;
             Description = config.Description;
```
mlorbetske: Since we assert that `Tags` and `CacheParameters` are non-null by iterating them lower in the method (and these are get-only) we should probably guarantee they're non null in the ctor

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/SimpleConfigModel.cs
```diff
@@ -75,9 +75,102 @@ public SimpleConfigModel()
 
         public IReadOnlyList<ICreationPathModel> PrimaryOutputs { get; set; }
 
-        public IReadOnlyDictionary<string, string> Tags { get; set; }
+        private IReadOnlyDictionary<string, string> _tagsDeprecated { get; set; }
```
mlorbetske: If this needs to be a property, it should be cased as a method (`TagsDeprecated`), if it can be made to be a field (since it's private now) the name's OK

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/SimpleConfigModel.cs
```diff
@@ -75,9 +75,102 @@ public SimpleConfigModel()
...
+            get
+            {
+                Dictionary<string, ICacheTag> tags = new Dictionary<string, ICacheTag>();
+
+                foreach (KeyValuePair<string, ParameterSymbol> tagParameter in TagDetails)
+                {
+                    string defaultValue;
+
+                    if (string.IsNullOrEmpty(tagParameter.Value.DefaultValue)
+                        && (tagParameter.Value.Choices.Keys.Count() == 1))
```
mlorbetske: `tagParameter.Value.Choices.Count == 1`?

src/Microsoft.TemplateEngine.Edge/Settings/TemplateInfo.cs
seancpeters: This is in Edge, so we unfortunately don't have access to the helper. We'll need to come up with a general plan for this type of situation.

src/Microsoft.TemplateEngine.Edge/Template/MatchKind.cs
seancpeters: No, they're not worse, but pretty bad. This was made before the order of this enum mattered. I'll move InvalidParameterName above Mismatch.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/SimpleConfigModel.cs
seancpeters: It should be a field

src/Microsoft.TemplateEngine.Edge/Settings/TemplateInfo.cs
mlorbetske: We can add the file to this project as a link if needed. Maybe we can move those helpers to a "shared" folder and just include them in both projects.

src/Microsoft.TemplateEngine.Cli/New3Command.cs
seancpeters: Correct, they can never be null. They're populated in these places:
- TemplateInfo ctor
- TemplateCache.LocalizeCacheTags()
- SimpleConfigModel.Tags - because all TagDetails instances are from choice params.

src/Microsoft.TemplateEngine.Cli/New3Command.cs
seancpeters: D'oh!...will be fixed.

src/Microsoft.TemplateEngine.Cli/New3Command.cs
seancpeters: Adding in a new class file in Utils: EqualityExtenions.cs

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/RunnableProjectGenerator.cs
seancpeters: An extension like won't quite work here because the exact match (line 373) short circuits the starts with match. So everything must be checked. But I'll re-write this to only iterate once.

src/Microsoft.TemplateEngine.Cli/New3Command.cs
seancpeters: Discussed offline. Decided to make this synchronous since the template generators are no longer invoked during this flow.

PR #342, opened 2017-02-16
Fixed case sensitivity on reading post actions & primary outputs

seancpeters:


DavidKarlas:
When is this fix expected to be part of https://dotnet.myget.org/gallery/templating ?

PR #373, opened 2017-02-25
Have errors instruct to call --help instead of displaying help

mlorbetske:
Tried this out a few ways:

Error + help specified (no template)
```
D:\Projects\templating>dotnet new3 -h -c
Invalid input switch:
  -c
Template Instantiation Commands for .NET Core CLI

Usage: dotnet new3 [arguments] [options]

Arguments:
  template  The template to instantiate.

Options:
  -l|--list         Lists templates containing the specified name. If no name is specified, lists all templates.
  -lang|--language  Specifies the language of the template to create.
  -n|--name         The name for the output being created. If no name is specified, the name of the current directory is used.
  -o|--output       Location to place the generated output.
  -h|--help         Displays help for this command.
  --type            Shows a subset of the available templates. Valid values are "project", "item" or "other".
  --force           Forces content to be generated even if it would change existing files.
```

Error + help specified (template)
```
D:\Projects\templating>dotnet new3 mvc -h -c
Invalid input switch:
  -c


Usage:  [options]

Options:
  -l|--list         Lists templates containing the specified name. If no name is specified, lists all templates.
  -lang|--language  Specifies the language of the template to create.
  -n|--name         The name for the output being created. If no name is specified, the name of the current directory is used.
  -o|--output       Location to place the generated output.
  -h|--help         Displays help for this command.
  --type            Shows a subset of the available templates. Valid values are "project", "item" or "other".
  --force           Forces content to be generated even if it would change existing files.


ASP.NET Core Web App (C#)
Author: Microsoft
Options:
  -au|--auth           The type of authentication to use
                           None          - No authentication
                           Individual    - Individual authentication
                       Default: None

  -uld|--use-local-db  Whether or not to use LocalDB instead of SQLite
                       bool - Optional
                       Default: false

  -f|--framework
                           netcoreapp1.0    - Target netcoreapp1.0
                           netcoreapp1.1    - Target netcoreapp1.1
                       Default: netcoreapp1.1
```

Error without help specified (no template)
```
D:\Projects\templating>dotnet new3 -c
Invalid input switch:
  -c
Run dotnet new3 --help for usage information.
```

Error without help specified (template)
```
D:\Projects\templating>dotnet new3 mvc -c
Invalid input switch:
  -c
Run dotnet new3 mvc --help for usage information.
```

@sayedihashimi

### DIFF
--- a/src/Microsoft.TemplateEngine.Cli/New3Command.cs
+++ b/src/Microsoft.TemplateEngine.Cli/New3Command.cs
@@ -288,7 +288,8 @@ namespace Microsoft.TemplateEngine.Cli
                 case CreationResultStatus.OperationNotSpecified:
                     break;
                 case CreationResultStatus.InvalidParamValues:
-                    Reporter.Error.WriteLine(string.Format(LocalizableStrings.InvalidParameterValues, instantiateResult.Message, resultTemplateName).Bold().Red());
+                    string invalidParamsError = GetTemplateParameterErrorsMessage(template, out IParameterSet ps, out IReadOnlyList<string> userParamsWithInvalidValues);
+                    Reporter.Error.WriteLine(invalidParamsError.Bold().Red());
                     Reporter.Error.WriteLine(string.Format(LocalizableStrings.RunHelpForInformationAboutAcceptedParameters, $"{CommandName} {TemplateName}").Bold().Red());
                     break;
                 default:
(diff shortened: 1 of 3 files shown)


sayedihashimi:
I like this because it makes it very clear what the issue is. Also not displaying the help is good for users who knew the right switch but just mistyped it. In general we should try to avoid duplication and display the minimal text because it's easier to digest that way. Great improvement here. Thanks.

src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -529,16 +537,26 @@ private CreationResultStatus EnterMaintenanceFlow()
...
-                if(commandParseFailureMessage != null)
+                if (commandParseFailureMessage != null)
                 {
                     Reporter.Error.WriteLine(commandParseFailureMessage.Bold().Red());
                 }
 
                 ShowUsageHelp();
                 ShowTemplateHelp(UnambiguousTemplateToUse.Info);
-                return argsError ? CreationResultStatus.InvalidParamValues : CreationResultStatus.Success;
+                return CreationResultStatus.Success;
```
phenning: Couldn't we also he in a state where we had arg errors? Do we want to return success in that case? Does this translate to the error level for the process?

mlorbetske: @phenning I wasn't quite sure what to do with that state - there's an arg error, but the user asked for help, so help was successfully shown but they were also notified of an error. You'd vote for putting it back to return error if there's an error even though help was specified? (I'm down with that, just looking for someone to overcome my indecision)

phenning: I think having it be an error helps for scripting cases even though I'm not sure why someone would specify --help in that case.

src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -180,6 +178,11 @@ private IFilteredTemplateInfo UseSecondaryCriteriaToDisambiguateTemplateMatches(
 
         public static int Run(string commandName, ITemplateEngineHost host, Action<IEngineEnvironmentSettings, IInstaller> onFirstRun, string[] args)
         {
+            if (args.Any(x => string.Equals(x, "--debug:attach", StringComparison.Ordinal)))
```
seancpeters: Nice way to get around needing to parse the args to get at this flag. Makes it easier to attach earlier 👍

src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -180,6 +178,11 @@ private IFilteredTemplateInfo UseSecondaryCriteriaToDisambiguateTemplateMatches(
...
+            {
+                Console.ReadLine();
```
prafullbhosale: Might be useful to check `System.Diagnostics.Debugger.IsAttached` instead of `Console.ReadLine()`, incase this command is not run from console.

mlorbetske: Thinking about this - since this is the console project (and getting this to run from F5 in VS/VS Code is much more difficult than running it with setup/explicit launch), I think we can hold off on this one for now.

Issue #376, opened 2017-02-26
Add IsPackable=false to test project templates


tintoy:
Hi.

At present, running `dotnet pack` on a solution with test projects often generates unnecessary warnings when packing test projects. Given that there's no point in packing test projects, how about adding:

```xml
<IsPackable>false</IsPackable>
```

to the test project templates.

That way, they'll be ignored when running `dotnet pack`. If you guys are ok with this idea, I could open a PR - it's probably as simple as adding a single line to 4 files.

mlorbetske:
Sounds good to me, @codito what do you think of this idea?

motowilliams:
> Given that there's no point in packing test projects

There isn't?

tintoy:
@motowilliams why produce a NuGet package for your tests? Unless you're planning on distributing them it doesn't seem to serve any real purpose (that I can see, at least).

motowilliams:
@tintoy that's precisely the reason - we package our suites of integration, acceptance and regression tests to our deployment targets.

tintoy:
@mlorbetske puts it better than I could; I was admittedly focusing only on the common case and didn't think someone might want to do that. Perhaps if it was a switch-style template parameter that's off by default?

codito:
@mlorbetske +1. I think we should set it in the `Microsoft.NET.Test.Sdk` package (it will apply to both IDE, CLI and keep the csproj clean too).

https://github.com/Microsoft/vstest/issues/380 tracks this in vstest.

tintoy:
@codito - if it's set in `Microsoft.NET.Test.Sdk` can that still be overridden per-project, if required?

codito:
Yes, it should be overridable by user.

YunWeiaa:
Repro on D15.0/28106.9000

mlorbetske:
@YunWeiaa what’s reproducing? Visual Studio displaying an option to pack the project even with IsPackable=false or the project having a property saying IsPackable=true by default? If it’s the first one, an issue should be opened on NuGet/home, if the second one (or anything else related to the test templates) an issue should be filed on dotnet/test-templates as those templates are now maintained in that repo.

PR #377, opened 2017-02-26
Set IsPackable=false for new test projects

tintoy:
Hi :)

I'm opening this PR so there's something a little more concrete to discuss for dotnet/templating#376.

To summarise the discussion so far:

* The rationale for this is that creating NuGet packages from tests is probably not the more-common use case and so by default these project should not participate when "dotnet pack" is run at the solution level.
* Some users _do_ want to create packages for their tests, and this change doesn't prevent them from doing so if they explicitly opt into it.

Open questions:

- [ ] Should this be configurable via a template parameter (e.g. `--packable` or `--enable-packaging`)?
- [ ] Er, sorry, GitHub's saying that every line in those files has changed, but as far as I can tell that's not the case (I only changed the `IsPackable` line). I'm on a Mac, so perhaps it's a CR/LF issue? If so, could someone using Windows let me know and I'll fiddle with Git's autocrlf and try again.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -142,7 +142,7 @@ publish/
 # Publish Web Output
 *.[Pp]ublish.xml
 *.azurePubxml
-# TODO: Comment the next line if you want to checkin your web deploy settings 
+# TODO: Comment the next line if you want to checkin your web deploy settings
 # but database connection strings (with potential passwords) will be unencrypted
 *.pubxml
 *.publishproj
@@ -243,3 +243,6 @@ Microsoft.TemplateEngine.VC.db
 
 # Dotnet CLI
 .dotnet
+
+# Mac
+.DS_Store
--- a/template_feed/Microsoft.DotNet.Test.ProjectTemplates.1.x/content/MSTest-CSharp/.template.config/dotnetcli.host.json
+++ b/template_feed/Microsoft.DotNet.Test.ProjectTemplates.1.x/content/MSTest-CSharp/.template.config/dotnetcli.host.json
@@ -4,6 +4,7 @@
       "longName": "framework"
     },
     "EnablePack": {
+      "shortName": "p",
       "longName": "enable-pack"
     }
   }
--- a/template_feed/Microsoft.DotNet.Test.ProjectTemplates.1.x/content/MSTest-CSharp/.template.config/dotnetcli.host.json
+++ b/template_feed/Microsoft.DotNet.Test.ProjectTemplates.1.x/content/MSTest-CSharp/.template.config/dotnetcli.host.json
@@ -2,6 +2,9 @@
   "symbolInfo": {
     "Framework": {
       "longName": "framework"
+    },
+    "EnablePack": {
+      "longName": "enable-pack"
     }
   }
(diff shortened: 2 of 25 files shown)


mlorbetske:
@tintoy would you mind rebasing onto rel/vs2017/post-rtw? Also, from the conversation in #376, it sounds like this will be turned off in the test SDK by default at some point, perhaps this should be included as an option to turn it on instead?

```
<IsPackable Conditon="'$(EnablePack)' == 'true'">true</IsPackable>
```

+a parameter added to the `template.json` files (type: `bool`, default: `false`) and a mapping for the CLI (in `dotnetcli.host.json`) for `-p|--enable-pack`

If there's a reason to not wait until the test SDK update, there's also this option (with the same changes in template.json & the host specific file)
```
<IsPackable Conditon="'$(EnablePack)' != 'true'">false</IsPackable>
<IsPackable Conditon="'$(EnablePack)' == 'true'">true</IsPackable>
```

Thoughts?

tintoy:
Sure, no problem - that makes sense to me :)

I'll give that a go tomorrow morning.

tintoy:
BTW, where does `$(EnablePack)` come from? Is that something we'll define elsewhere, or is that something else from the project system?

tintoy:
Rebasing.

mlorbetske:
@tintoy in MSBuild files, you can refer to symbols defined in `template.json` like any other MSBuild property  in conditions - if the only references a `Condition` contains refer to symbols from `template.json`, they'll be evaluated as part of template generation.

Example:
Project file:
https://github.com/dotnet/templating/blob/master/template_feed/Microsoft.DotNet.Web.ProjectTemplates.1.x/content/StarterWeb-CSharp/Company.WebApplication1.csproj#L4-L6

Declarations in template.json
https://github.com/dotnet/templating/blob/master/template_feed/Microsoft.DotNet.Web.ProjectTemplates.1.x/content/StarterWeb-CSharp/.template.config/template.json#L78-L92 (`Framework`)
https://github.com/dotnet/templating/blob/master/template_feed/Microsoft.DotNet.Web.ProjectTemplates.1.x/content/StarterWeb-CSharp/.template.config/template.json#L42-L61 (`auth` and the computed value `IndividualAuth`)

tintoy:
Interesting! I had no idea it could do that :)

That'll make a lot of things simpler.

tintoy:
Ok - would you mind having a look at d3bda09? If that's what you mean, I'll have a go at applying it to the rest tomorrow morning.

mlorbetske:
Beautiful, jfyi there's a setting for controlling short name generation too (I think it's "shortname" but it's entirely possible that I'm remembering wrong), you could use that to manually control the short switch. As it is now, this will generate -ep|--enable-pack as a command line option

tintoy:
Great! I'll look it up and use it. Are you ok with -p ?

mlorbetske:
Cc @codito

codito:
:+1: thanks!

mlorbetske:
Thanks @tintoy!

PR #383, opened 2017-02-27
Template help group param merging

seancpeters:
Better matching across templates with the same group identity.
Appended ".2.0" to the identity of all the 2.0 templates.

### DIFF
--- a/src/Microsoft.TemplateEngine.Cli/New3Command.cs
+++ b/src/Microsoft.TemplateEngine.Cli/New3Command.cs
@@ -98,11 +98,10 @@ namespace Microsoft.TemplateEngine.Cli
 
         private static bool AreAllTemplatesSameGroupIdentity(IEnumerable<IFilteredTemplateInfo> templateList)
         {
-            Func<IFilteredTemplateInfo, string> groupIdentitySelector = (x) => x.Info.GroupIdentity;
-            return templateList.AllAreTheSame(groupIdentitySelector, StringComparer.OrdinalIgnoreCase);
+            return templateList.AllAreTheSame((x) => x.Info.GroupIdentity, StringComparer.OrdinalIgnoreCase);
         }
 
-        private static IFilteredTemplateInfo HighestPrecedenceTemplateIfAllSameGroupIdentity(IReadOnlyList<IFilteredTemplateInfo> templateList)
+        private static IFilteredTemplateInfo FindHighestPrecedenceTemplateIfAllSameGroupIdentity(IReadOnlyList<IFilteredTemplateInfo> templateList)
         {
             if (!AreAllTemplatesSameGroupIdentity(templateList))
             {
@@ -205,7 +204,7 @@ namespace Microsoft.TemplateEngine.Cli
 
                     IReadOnlyList<IFilteredTemplateInfo> matchesAfterParameterChecks = _matchedTemplatesWithSecondaryMatchInfo.Where(x => x.IsParameterMatch).ToList();
 
-                    if (!matchesAfterParameterChecks.Any())
+                    if (matchesAfterParameterChecks.Count == 0)
                     {   // no param matches, continue additional matching with the list from before param checking.
                         matchesAfterParameterChecks = _matchedTemplatesWithSecondaryMatchInfo;
                     }
@@ -384,8 +383,18 @@ namespace Microsoft.TemplateEngine.Cli
 
         private string GetLanguageMismatchErrorMessage(string inputLanguage)
         {
+            string inputFlagForm;
+            if (_app.RemainingArguments.Contains("-lang"))
+            {
+                inputFlagForm = "-lang";
+            }
+            else
+            {
+                inputFlagForm = "--language";
+            }
+
             string invalidLanguageErrorText = LocalizableStrings.InvalidTemplateParameterValues;
-            invalidLanguageErrorText += Environment.NewLine + string.Format(LocalizableStrings.InvalidParameterDetail, "-lang", inputLanguage, "--language");
+            invalidLanguageErrorText += Environment.NewLine + string.Format(LocalizableStrings.InvalidParameterDetail, inputFlagForm, inputLanguage, "language");
             return invalidLanguageErrorText;
         }
 
@@ -428,7 +437,7 @@ namespace Microsoft.TemplateEngine.Cli
             {
                 IEnumerable<ITemplateInfo> templateList;
 
-                if (UnambiguousTemplateGroupToUse != null && UnambiguousTemplateGroupToUse.Any())
+                if (UnambiguousTemplateGroupToUse != null && UnambiguousTemplateGroupToUse.Count > 0)
                 {
                     templateList = UnambiguousTemplateGroupToUse.Select(x => x.Info);
                 }
@@ -461,7 +470,7 @@ namespace Microsoft.TemplateEngine.Cli
             foreach (IGrouping<string, ITemplateInfo> grouping in grouped)
             {
                 List<string> languageForDisplay = new List<string>();
-                HashSet<string> uniqueLanguages = new HashSet<string>();
+                HashSet<string> uniqueLanguages = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
 
                 foreach (ITemplateInfo info in grouping)
                 {
@@ -469,7 +478,7 @@ namespace Microsoft.TemplateEngine.Cli
                     {
                         foreach (string lang in languageTag.ChoicesAndDescriptions.Keys)
                         {
-                            if (!uniqueLanguages.Contains(lang))
+                            if (!uniqueLanguages.Add(lang))
                             {
                                 if (string.IsNullOrEmpty(Language) && string.Equals(defaultLanguage, lang, StringComparison.OrdinalIgnoreCase))
                                 {
@@ -478,10 +487,7 @@ namespace Microsoft.TemplateEngine.Cli
                                 else
                                 {
                                     languageForDisplay.Add(lang);
-                                    uniqueLanguages.Add(lang);
                                 }
-
-                                uniqueLanguages.Add(lang);
                             }
                         }
                     }
@@ -655,7 +661,7 @@ namespace Microsoft.TemplateEngine.Cli
 
             if (!anyValid)
             {
-                IFilteredTemplateInfo highestPrecedenceTemplate = HighestPrecedenceTemplateIfAllSameGroupIdentity(UnambiguousTemplateGroupToUse);
+                IFilteredTemplateInfo highestPrecedenceTemplate = FindHighestPrecedenceTemplateIfAllSameGroupIdentity(UnambiguousTemplateGroupToUse);
 
                 if (highestPrecedenceTemplate != null)
                 {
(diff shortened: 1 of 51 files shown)


src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -118,33 +118,41 @@ public IFilteredTemplateInfo UnambiguousTemplateToUse
...
-                            if (match == null)
-                            {
-                                match = info;
-                            }
-                            else
-                            {
-                                return UseSecondaryCriteriaToDisambiguateTemplateMatches();
-                            }
-                        }
+                        return _unambiguousTemplateToUse = languageMatchedTemplates.First();
```
mlorbetske: languageMatchedTemplates is an IReadOnlyList<T>, so we can just use `[0]` to get the first item

src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -118,33 +118,41 @@ public IFilteredTemplateInfo UnambiguousTemplateToUse
...
 
                 return UseSecondaryCriteriaToDisambiguateTemplateMatches();
             }
         }
 
+        private static IReadOnlyList<IFilteredTemplateInfo> FindLanguageMatchedTemplates(IEnumerable<IFilteredTemplateInfo> listToFilter, string language)
+        {
+            List<IFilteredTemplateInfo> languageMatches = new List<IFilteredTemplateInfo>();
+
+            if (string.IsNullOrEmpty(language))
```
mlorbetske: Wouldn't this be the other way around? If no language is specified, everything matches?

src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -168,6 +176,14 @@ private IFilteredTemplateInfo UseSecondaryCriteriaToDisambiguateTemplateMatches(
                     {
                         _secondaryFilteredUnambiguousTemplateToUse = matchesIncludingParameterChecks.First();
                     }
+                    else if (EnvironmentSettings.Host.TryGetHostParamDefault("prefs:language", out string defaultLanguage))
+                    {
+                        IReadOnlyList<IFilteredTemplateInfo> languageFiltered = FindLanguageMatchedTemplates(matchesIncludingParameterChecks, defaultLanguage);
+                        if (languageFiltered.Count == 1)
+                        {
+                            _secondaryFilteredUnambiguousTemplateToUse = languageFiltered.First();
```
mlorbetske: .First() again

src/Microsoft.TemplateEngine.Cli/New3Command.cs
seancpeters: The method is used in contexts that want to find explicit language matches.

src/Microsoft.TemplateEngine.Cli/New3Command.cs
seancpeters: 👍

src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -113,65 +146,90 @@ public IFilteredTemplateInfo UnambiguousTemplateToUse
...
+            if (_secondaryFilteredUnambiguousTemplateGroupToUse == null)
             {
                 if (_matchedTemplatesWithSecondaryMatchInfo == null)
                 {
                     _matchedTemplatesWithSecondaryMatchInfo = FilterTemplatesOnParameters(_matchedTemplates).Where(x => x.IsMatch).ToList();
 
-                    IReadOnlyList<IFilteredTemplateInfo> matchesIncludingParameterChecks = _matchedTemplatesWithSecondaryMatchInfo.Where(x => x.IsParameterMatch).ToList();
+                    IReadOnlyList<IFilteredTemplateInfo> matchesAfterParameterChecks = _matchedTemplatesWithSecondaryMatchInfo.Where(x => x.IsParameterMatch).ToList();
+
+                    if (!matchesAfterParameterChecks.Any())
```
mlorbetske: We can just check the count of the list here

src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -362,7 +428,11 @@ private IReadOnlyList<ITemplateInfo> TemplatesToDisplayInfoAbout
             {
                 IEnumerable<ITemplateInfo> templateList;
 
-                if (_matchedTemplatesWithSecondaryMatchInfo != null && _matchedTemplatesWithSecondaryMatchInfo.Count > 0)
+                if (UnambiguousTemplateGroupToUse != null && UnambiguousTemplateGroupToUse.Any())
```
mlorbetske: Can check count

src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -615,14 +780,15 @@ private CreationResultStatus EnterMaintenanceFlow()
         {
             PerformCoreTemplateQuery();
 
-            if (UnambiguousTemplateToUse != null)
+            if (UnambiguousTemplateGroupToUse != null && UnambiguousTemplateGroupToUse.Any())
```
mlorbetske: Count

src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -97,13 +96,47 @@ public New3Command(string commandName, ITemplateEngineHost host, Action<IEngineE
 
         public bool SkipUpdateCheck => _app.InternalParamHasValue("--skip-update-check");
 
-        public IFilteredTemplateInfo UnambiguousTemplateToUse
+        private static bool AreAllTemplatesSameGroupIdentity(IEnumerable<IFilteredTemplateInfo> templateList)
+        {
+            Func<IFilteredTemplateInfo, string> groupIdentitySelector = (x) => x.Info.GroupIdentity;
+            return templateList.AllAreTheSame(groupIdentitySelector, StringComparer.OrdinalIgnoreCase);
```
mlorbetske: Why break out the func to its own line?

src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -97,13 +96,47 @@ public New3Command(string commandName, ITemplateEngineHost host, Action<IEngineE
...
+        }
+
+        private static IFilteredTemplateInfo HighestPrecedenceTemplateIfAllSameGroupIdentity(IReadOnlyList<IFilteredTemplateInfo> templateList)
```
mlorbetske: Action names for methods

src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -323,6 +382,13 @@ private string GetTemplateParameterErrorsMessage(ITemplateInfo templateInfo, out
             return null;
         }
 
+        private string GetLanguageMismatchErrorMessage(string inputLanguage)
+        {
+            string invalidLanguageErrorText = LocalizableStrings.InvalidTemplateParameterValues;
+            invalidLanguageErrorText += Environment.NewLine + string.Format(LocalizableStrings.InvalidParameterDetail, "-lang", inputLanguage, "--language");
```
mlorbetske: Let's use full switch names when mentioning them in output messages (unless we're using the user specified form)

src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -390,27 +460,34 @@ private void DisplayTemplateList(bool showAll = false)
 
             foreach (IGrouping<string, ITemplateInfo> grouping in grouped)
             {
-                List<string> languages = new List<string>();
+                List<string> languageForDisplay = new List<string>();
+                HashSet<string> uniqueLanguages = new HashSet<string>();
```
mlorbetske: Should set a comparer

src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -390,27 +460,34 @@ private void DisplayTemplateList(bool showAll = false)
...
 
                 foreach (ITemplateInfo info in grouping)
                 {
                     if (info.Tags != null && info.Tags.TryGetValue("language", out ICacheTag languageTag))
                     {
                         foreach (string lang in languageTag.ChoicesAndDescriptions.Keys)
                         {
-                            if (string.IsNullOrEmpty(Language) && string.Equals(defaultLanguage, lang, StringComparison.OrdinalIgnoreCase))
+                            if (!uniqueLanguages.Contains(lang))
```
mlorbetske: Couldn't this be converted to `if(uniqueLanuages.Add(lang)` and the call to add at the end of the block be removed?

src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -390,27 +460,34 @@ private void DisplayTemplateList(bool showAll = false)
...
+                                {
+                                    languageForDisplay.Add($"[{lang}]");
+                                }
+                                else
+                                {
+                                    languageForDisplay.Add(lang);
+                                    uniqueLanguages.Add(lang);
+                                }
+
+                                uniqueLanguages.Add(lang);
```
mlorbetske: This add happens both in the else and outside of it - the one in the else can probably be removed, as could this one if the initial contains check is replaced with an Add

src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -521,38 +612,123 @@ private CreationResultStatus EnterMaintenanceFlow()
...
                 }
+            }
 
-                DisplayTemplateList();
-                return CreationResultStatus.Success;
+            DisplayTemplateList();
+            return CreationResultStatus.Success;
+        }
+
+        private CreationResultStatus SingularGroupDisplayTemplateHelp()
```
mlorbetske: Not sure how to read this method name, not sure if it reflects an action or not

src/Microsoft.TemplateEngine.Cli/New3Command.cs
seancpeters: 👍

src/Microsoft.TemplateEngine.Cli/New3Command.cs
seancpeters: Seems more readable to me, but I don't feel particularly strongly about it either way.

src/Microsoft.TemplateEngine.Cli/New3Command.cs
seancpeters: Changing to FindHighestPrecedenceTemplateIfAllSameGroupIdentity

src/Microsoft.TemplateEngine.Cli/New3Command.cs
seancpeters: I did it this way to be more consistent with the output from GetTemplateParameterErrorsMessage(), which lists the canonical (full name) at the end of the message. Perhaps a quick discussion offline is best here.

src/Microsoft.TemplateEngine.Cli/New3Command.cs
seancpeters: 👍

src/Microsoft.TemplateEngine.Cli/New3Command.cs
seancpeters: Yeah, this is a mess, fixing.

src/Microsoft.TemplateEngine.Cli/New3Command.cs
seancpeters: Changing to DisplayTemplateHelpForSingularGroup

PR #393, opened 2017-02-28
Use the new unified Trie and add support for look-arounds

mlorbetske:
Do not merge - these are only here for visibility for now

This PR contains two things:
1) The new Trie that addresses #6 - either the Trie will need to be changed or the IOperation interface will need to absorb the requirement for tracking the absolute position in content streams
2) The new command parser to break the dependency on CLI.Utils (there's also a dependency for console colors & invoking other `dotnet` commands that will need to be dealt with separately)

@seancpeters


mlorbetske:
The remaining work here is to allow the specification of lookarounds in configuration

mlorbetske:
We'll make a new issue for adding full support for lookarounds for each operation (in its matched tokens). This PR is now ready.

src/Microsoft.TemplateEngine.Core/Util/ProcessorState.cs
```diff
@@ -54,31 +74,40 @@ public ProcessorState(Stream source, Stream target, int bufferSize, int flushThr
...
+                byEncoding[encoding] = trie;
             }
 
-            if (bufferSize < _trie.MaxLength && !sizedToStream)
+            _trie = new TrieEvaluator<OperationTerminal>(trie);
+
+            if (bufferSize < _trie.MaxLength + 1)
             {
-                byte[] tmp = new byte[_trie.MaxLength];
+                byte[] tmp = new byte[_trie.MaxLength + 1];
```
seancpeters: The sizedToStrream check was removed with this PR. Wondering if there could be problems if the entire source stream is shorter than the _trie.MaxLength + 1.

src/Microsoft.TemplateEngine.Core/Util/ProcessorState.cs
```diff
@@ -87,154 +116,183 @@ public ProcessorState(Stream source, Stream target, int bufferSize, int flushThr
...
-            }
-
-            while (CurrentBufferLength > 0)
-            {
-                int token;
-                int posedPosition = CurrentBufferPosition;
-
-                if (CurrentBufferLength == CurrentBuffer.Length && CurrentBufferLength - CurrentBufferPosition < _trie.MaxLength)
+                //Loop until we run out of data in the buffer
+                for(; CurrentBufferPosition < CurrentBufferLength;)
```
seancpeters: while loop ?

src/Microsoft.TemplateEngine.Core/Matching/TerminalBase.cs
```diff
@@ -0,0 +1,9 @@
+namespace Microsoft.TemplateEngine.Core.Matching
+{
+    public abstract class TerminalBase
+    {
```
seancpeters: Constructor possibly?

src/Microsoft.TemplateEngine.Core/Util/TokenTrie.cs
```diff
@@ -1,129 +1,89 @@
...
+        public int AddToken(IToken token)
+        {
+            int count = _tokens.Count;
+            AddToken(token, count);
+            return count;
         }
 
-        public bool GetOperation(byte[] buffer, int bufferLength, ref int currentBufferPosition, out int token)
+        public void AddToken(IToken token, int index)
         {
```
seancpeters: I don't see any checking that multiple tokens don't have the same index. This would only potentially be a problem if tokens are added with explicit indexes.
OTOH, I don't have this fully internalized yet, so perhaps duplicated indexes are ok.

src/Microsoft.TemplateEngine.Core/Util/ProcessorState.cs
mlorbetske: This is OK as the current buffer size is tracked differently

src/Microsoft.TemplateEngine.Core/Util/ProcessorState.cs
mlorbetske: Done

PR #401, opened 2017-03-03
Fix all build warnings, add build.proj

mlorbetske:
Also adds the git commit hash and commit count as a class in each project


peterblazejewicz:
Thx folks!

PR #407, opened 2017-03-07
Handling reading different cache formats, and upgrading the cache aut…

seancpeters:
…omatically as needed

### DIFF
--- a/src/Microsoft.TemplateEngine.Abstractions/ISettingsLoader.cs
+++ b/src/Microsoft.TemplateEngine.Abstractions/ISettingsLoader.cs
@@ -28,7 +28,7 @@ namespace Microsoft.TemplateEngine.Abstractions
 
         bool TryGetMountPointInfo(Guid mountPointId, out MountPointInfo info);
 
-        void WriteTemplateCache(IList<ITemplateInfo> templates, string cacheVersion, string locale, bool isCurrentCache);
+        void WriteTemplateCache(IList<ITemplateInfo> templates, string locale, bool isCurrentCache);
 
         IFile FindBestHostTemplateConfigFile(IFileSystemInfo config);
     }
--- a/src/Microsoft.TemplateEngine.Abstractions/ISettingsLoader.cs
+++ b/src/Microsoft.TemplateEngine.Abstractions/ISettingsLoader.cs
@@ -28,7 +28,7 @@ namespace Microsoft.TemplateEngine.Abstractions
 
         bool TryGetMountPointInfo(Guid mountPointId, out MountPointInfo info);
 
-        void WriteTemplateCache(IList<ITemplateInfo> templates, string locale, bool isCurrentCache);
+        void WriteTemplateCache(IList<ITemplateInfo> templates, string cacheVersion, string locale, bool isCurrentCache);
 
         IFile FindBestHostTemplateConfigFile(IFileSystemInfo config);
     }
--- a/src/Microsoft.TemplateEngine.Cli/ExtendedTemplateEngineHost.cs
+++ b/src/Microsoft.TemplateEngine.Cli/ExtendedTemplateEngineHost.cs
@@ -4,9 +4,7 @@ using System.IO;
 using System.Linq;
 using Microsoft.DotNet.Cli.Utils;
 using Microsoft.TemplateEngine.Abstractions;
-using Microsoft.TemplateEngine.Abstractions.Mount;
 using Microsoft.TemplateEngine.Abstractions.PhysicalFileSystem;
-using Microsoft.TemplateEngine.Edge.Settings;
 
 namespace Microsoft.TemplateEngine.Cli
 {
--- a/src/Microsoft.TemplateEngine.Cli/ExtendedTemplateEngineHost.cs
+++ b/src/Microsoft.TemplateEngine.Cli/ExtendedTemplateEngineHost.cs
@@ -4,7 +4,9 @@ using System.IO;
 using System.Linq;
 using Microsoft.DotNet.Cli.Utils;
 using Microsoft.TemplateEngine.Abstractions;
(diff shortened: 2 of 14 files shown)


test/Microsoft.TemplateEngine.Mocks/MockEngineEnvironmentSettings.cs
```diff
@@ -0,0 +1,18 @@
+﻿using Microsoft.TemplateEngine.Abstractions;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.TemplateEngine.Mocks
+{
+    public class MockEngineEnvironmentSettings : IEngineEnvironmentSettings
+    {
+        public ISettingsLoader SettingsLoader { get { throw new NotImplementedException(); } }
```
mlorbetske: Should any of these be implemented?

src/Microsoft.TemplateEngine.Abstractions/ISettingsLoader.cs
```diff
@@ -28,7 +28,7 @@ public interface ISettingsLoader
 
         bool TryGetMountPointInfo(Guid mountPointId, out MountPointInfo info);
 
-        void WriteTemplateCache(IList<ITemplateInfo> templates, string locale, bool isCurrentCache);
+        void WriteTemplateCache(IList<ITemplateInfo> templates, string cacheVersion, string locale, bool isCurrentCache);
```
mlorbetske: Shouldn't the version of the cache the engine requires be an implementation detail of the settings loader or template cache and not need to be exposed?

src/Microsoft.TemplateEngine.Cli/ExtendedTemplateEngineHost.cs
```diff
@@ -144,5 +146,26 @@ private bool GlobalJsonFileExistsInPath
                 return found;
             }
         }
+
+        public void RebuildCacheFromSettingsIfNotCurrent(IEngineEnvironmentSettings environmentSettings)
+        {
+            TemplateCache cache = new TemplateCache(environmentSettings);
+            cache.LoadTemplateCacheForLocale(Locale, out string cacheVersion);
+
+            if (TemplateCache.CheckIfCacheVersionIsCurrent(cacheVersion))
```
mlorbetske: This feels like something that the cache could do internally (`cache.IsCurrentVersion` or some such thing)

src/Microsoft.TemplateEngine.Cli/ExtendedTemplateEngineHost.cs
```diff
@@ -144,5 +146,26 @@ private bool GlobalJsonFileExistsInPath
...
+            {
+                return;
+            }
+
+            cache.CacheVersion = TemplateCache.CurrentCacheVersion;
+            cache.TemplateInfo.Clear();
+
+            foreach (MountPointInfo mountPoint in environmentSettings.SettingsLoader.MountPoints)
```
mlorbetske: Should we read the information from the templates currently in the cache (if we can) so that we don't end up scanning mount points that contain only components for templates?

src/Microsoft.TemplateEngine.Edge/Settings/TemplateCache.cs
```diff
@@ -18,6 +18,8 @@ namespace Microsoft.TemplateEngine.Edge.Settings
 {
     public class TemplateCache
     {
+        public static readonly string CurrentCacheVersion = "1.0.0.0";
```
mlorbetske: This shouldn't really be public, should it?

src/Microsoft.TemplateEngine.Edge/Settings/TemplateCache.cs
```diff
@@ -57,10 +69,23 @@ public TemplateCache(IEngineEnvironmentSettings environmentSettings, JObject par
         [JsonProperty]
         public List<TemplateInfo> TemplateInfo { get; set; }
 
+        [JsonProperty]
+        public string CacheVersion { get; set; }
+
+        public static bool CheckIfCacheVersionIsCurrent(string cacheVersion)
```
mlorbetske: If this were an instance method, it could just be a property & compare to the static/const field indicating the current version number caches would be written with

src/Microsoft.TemplateEngine.Edge/Settings/TemplateCache.cs
```diff
@@ -224,11 +249,17 @@ private void ScanForComponents(IMountPoint mountPoint, string templateDir)
             }
         }
 
-        public List<TemplateInfo> LoadTemplateCacheForLocale(string locale)
+        public List<TemplateInfo> LoadTemplateCacheForLocale(string locale, out string cacheVersion)
```
mlorbetske: Since this seems to ripple out a bit, hopefully we can internalize the version checks rather than having them passed in

src/Microsoft.TemplateEngine.Abstractions/ISettingsLoader.cs
mlorbetske: Thinking more about this - I don't think the version stamp for the cache belongs on the template cache - getting the feeling that this would be more at home on the settings file itself (especially since the template cache & settings file must be coordinated anyway)

test/Microsoft.TemplateEngine.Mocks/MockEngineEnvironmentSettings.cs
seancpeters: I was thinking we could implement them as need for specific tests.

src/Microsoft.TemplateEngine.Edge/Settings/TemplateCache.cs
seancpeters: Ideally yes, but won't work right until the template cache gets refactored.

PR #409, opened 2017-03-07
Allow runtime to be passed in

karajas:
/cc @mlorbetske This is a small adjustment needed for the internal CI.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -134,7 +134,7 @@ $x = PWD
 # Restore
 Write-Host "Restoring dotnet new3..."
 cd "$RepoRoot\src\dotnet-new3"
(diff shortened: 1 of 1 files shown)


mlorbetske:
Looks good, but could you take a look at #401 to see if it'll meet your needs as well? It changes over to using a build.proj rather than orchestrating everything through the scripts

karajas:
I see. Can you allow passing in off other arguments other than the param list at the top so that dotnet msbuild can pick them up?

mlorbetske:
@karajas the properties that are listed there are the only ones that you'd really want to change in this way of building (producing the dotnet-new3 executable, all packages, templates, etc.), is there something in particular you have in mind that you'd want to be able to change that isn't represented in there?

Issue #399, opened 2017-03-02
"How to build and run the latest" - Moving towards an explicit explanation....


chad-ramos:
### Issue
Obviously my position approaching this, is an outsider looking in.  One looking to make contributions to this project would require a ramp-up, of which starts with [How to build and run the latest](https://github.com/dotnet/templating#how-to-build-and-run-the-latest). 

The combination of the four steps that are outlined and how I (an outsider) perceive them, fails to produce a working dev environment.  Of which, I would attribute to implicit explanations. 

### Breakdown

**Prerequisites**

- Though to one it might seem obvious, things like denoting vs 2017 is a requirement is of value to others. 

**Step 1: Get the SDK for your platform from dotnet/cli**

- More explanation is needed as to what version is required and how to maintain version compatibility moving forward. 
- Do we need 1.0.x, 1.1.x, 2.x?  Do we need RC or Alpha? 

**Step 2: Clone this repository to your local machine**

- Simple enough on the surface but details on branching strategies and expectations would go a long way.
- For example, as of the date I am writing this, it appears the default branch is rel/vs2017/post-rtw.
  - If the expectation is for contributes to branch off of branch-x, explicit explanation would go a long way.  

**Step 3: Run the setup script**

- Again, simple enough but results in an unclear outcome.  See next step...

**Step 4: That's it! Now you can run dotnet new3**

![untitled-1](https://cloud.githubusercontent.com/assets/3170200/23528117/7a210a02-ff4d-11e6-8983-cd84aaaef32a.jpg)
 
- Running through the steps as I have viewed them, has resulted in the above. 
  - ```dotnet new --help``` produces expected output while ```dotnet new3 --help``` does not.

### Solution
It would be of value to open up the conversation early about the process I or any other potential new contributor goes through when trying to ramp up the knowledge set needed to contribute to this project.   Personally, I would be more then willing to document the knowledge gained as I progress through this process.  Be that via the wiki or readme.

mlorbetske:
Hi @chad-ramos, answers to your questions below. I totally agree that we should get more crisp about these things in our docs (and yes, if you're up for enhancing our getting started steps, we'd be more than happy to take a PR or a wiki update :) )

For step 1:
The latest 1.1 RC SDK is the one we've been using (build 004771 or higher should work)

For step 2:
Yep, we've got #288 tracking this, trying to get to this soon. The active development branch is rel/vs2017/post-rtw at the moment.

For step 3:
[...]

PR #417, opened 2017-03-10
README updates based around Setup and Debugging

chad-ramos:
Suggested updates to README that adds explicit details about the current setup and debugging process.   Of which applies to #399.

@mlorbetske

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,9 @@ Arguments:
 
 <truncated>
 ```
+### Template Nuget Packages
+- The build the that produces the **template Nuget packages** currently has a dependency on **Nuget.exe**.
+  - Because of this, those that wish to `install` using the **template Nuget packages** will need to be on Windows in order to produce the appropriate assets. 
     
 ## Debugging
 Debugging code requires your current `dotnet new3` session to have its active build session configured to DEBUG, and a debugger from your application of choice to be attached to the current running `dotnet new3` process. The steps required to accomplish this are outlined below.
--- a/README.md
+++ b/README.md
@@ -48,44 +48,27 @@ if you have any difficulties please file a new [issue](https://github.com/dotnet
 
 Now that we've covered the basics of using `dotnew new`, lets move on to info for template authors and contributors.
 
-# How to build and run the latest
+# How to build, run & debug the latest
 
 If you're authoring templates, or interested in contributing to this repo, then you're likely interested in how to use the latest version of this experience.
 The steps required are outlined below.
 
-## Prerequisites
-In order to work with this project, you will need the following.
-- [Visual Studio 2017](https://www.visualstudio.com/)
+## Aquire
 
-## Setup
+- Fork this repository.
+- Clone the forked repository to your local machine.
+  - **master** is a build branch and does not accept contributions directly.
+  - The default branch is the active development branch that accepts contributions and flows to master to produce packages.
 
-Step 1:
-- Get the SDK for your platform from [dotnet/cli](https://github.com/dotnet/cli).
-- Typically, development is  done against the latest RC FTS Release.  
-  - It is best to ask the team to clear up any compatibility concerns before starting.
+## Build & Run
 
-Step 2:
-- Clone this repository to your local machine.
-
-```bash
-git clone https://github.com/dotnet/templating.git
-```
-
-- There is no guarantee that **master** is the current working branch. 
-  - It is recommended that you check with the team to ensure you are developing or building from the correct branch. 
-
-Step 3: 
 - Open up a command prompt and navigation to the root of your source code.
-- Run the setup script.
+- Run the setup script appropriate your environment.
      - **Windows:** [setup.cmd](https://github.com/dotnet/templating/blob/master/setup.cmd)
      - **Mac/Linux**: [setup.sh](https://github.com/dotnet/templating/blob/master/setup.sh) 
 - When running the setup script, the existing built-in command `dotnet new` will be preserved. A new command `dotnet new3` will be enabled which allows you to create
 files with the latest Template Engine.
-- Closing your console window or "session" will destroy your current instance of `dotnet new3`.
-  - The next time a console window is opened, you will again need to run the setup script. 
-
-Step 4: 
- - That's it! Now you can run `dotnet new3`.
+- That's it! Now you can run `dotnet new3`.
 
 For example, here is the result of running `dotnet new3 --help` on a Mac (_truncated to save space here_).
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -53,18 +53,39 @@ Now that we've covered the basics of using `dotnew new`, lets move on to info fo
...
-Step 2: Clone this repository to your local machine
+## Setup
 
-Step 3:  Run the setup script
- - **Windows:** [setup.cmd](https://github.com/dotnet/templating/blob/master/setup.cmd)
- - **Mac/Linux**: [setup.sh](https://github.com/dotnet/templating/blob/master/setup.sh) 
+Step 1:
+- Get the SDK for your platform from [dotnet/cli](https://github.com/dotnet/cli).
+- Typically, development is  done against the latest RC FTS Release.  
+  - It is best to ask the team to clear up any compatibility concerns before starting.
```
mlorbetske: The actual build is done with a CLI that's dynamically acquired. The CLI acquired in this step is used for running the "dotnet new3" command. Without any CLI installed, this command can also be run as "dotnet-new3.exe" (or "dotnet-new3" on non-Windows)

README.md
```diff
@@ -53,18 +53,39 @@ Now that we've covered the basics of using `dotnew new`, lets move on to info fo
...
 
-When running the setup script, the existing built in command `dotnet new` will be preserved. A new command `dotnet new3` will be enabled which allows you to create
+Step 2:
+- Clone this repository to your local machine.
```
mlorbetske: For the purposes of contribution, we'd encourage people to fork the repo & clone their fork instead

README.md
```diff
@@ -53,18 +53,39 @@ Now that we've covered the basics of using `dotnew new`, lets move on to info fo
...
+
+```bash
+git clone https://github.com/dotnet/templating.git
+```
+
+- There is no guarantee that **master** is the current working branch.
```
mlorbetske: Master is the build branch, it does not accept contributions directly. The default branch (rel/vs2017/post-rtw at the moment) is the active dev branch/branch that accepts contributions and flows to master to produce packages

README.md
```diff
@@ -53,18 +53,39 @@ Now that we've covered the basics of using `dotnew new`, lets move on to info fo
...
+
+Step 3: 
+- Open up a command prompt and navigation to the root of your source code.
+- Run the setup script.
+     - **Windows:** [setup.cmd](https://github.com/dotnet/templating/blob/master/setup.cmd)
+     - **Mac/Linux**: [setup.sh](https://github.com/dotnet/templating/blob/master/setup.sh) 
+- When running the setup script, the existing built-in command `dotnet new` will be preserved. A new command `dotnet new3` will be enabled which allows you to create
 files with the latest Template Engine.
+- Closing your console window or "session" will destroy your current instance of `dotnet new3`.
+  - The next time a console window is opened, you will again need to run the setup script.
```
mlorbetske: This is currently true for Windows (until #401 gets merged). It's not currently true for non-Windows.

README.md
```diff
@@ -79,6 +100,56 @@ Arguments:
 
 <truncated>
 ```
+    
+# Debugging
+Debugging code requires your current `dotnet new3` session to have its active build session configured to DEBUG, and a debug process from Visual Studio to be attached to the current running `dotnet new3` session. The steps required to accomplish this are outlined below.
+
+## Setup
+
+Start by opening up the **Microsoft.Templating.sln** in Visual Studio. This solution contains the projects needed to run, modify & debug the application. Once your solution is loaded, open up a new command prompt and navigate to the root of your repository. At this point, you can take one of two paths.
```
mlorbetske: Rather than VS specifically, the application that will be attaching the debugger

README.md
```diff
@@ -79,6 +100,56 @@ Arguments:
...
+
+Path 1:
+- Run the "setup" script that applies to your environment. 
+- Once complete, type "debug" and press enter.
+  - The "debug" command is an alias that kicks off a command to run a script named **dn3buildmode-debug.cmd**. 
+    - Of which sets the build to a DEBUG mode.
+
+Path 2:
```
mlorbetske: I'd skip path 1 and just note that if you've already run setup in that console window, you can just type "debug" instead of the whole script name

README.md
```diff
@@ -79,6 +100,56 @@ Arguments:
...
+
+After your build is set to DEBUG mode, run the "setup" script to build a new `dotnet new3` session. 
+
+## Execution
+
+Once "setup" has completed successfully, run the following command.
+```bash
+dotnet new3 --debug:attach {{additonal args}}
+```
+By supplying the `--debug:attach` argument with any other argument(s) you are running, you are triggering a ` Console.ReadLine();` request which pauses execution of the Template Engine at an early point in its logical loop.
```
mlorbetske: Rather than logical loop, execution perhaps (as this path is not in a user code loop)

README.md
```diff
@@ -79,6 +100,56 @@ Arguments:
...
+
+Once the engine is "paused", you have the opportunity to attach a debugger to the running `dotnet new3` process. 
+
+In Visual studio...
+
+- Open the **Microsoft.TemplateEngine.Cli.New3Command** class and locate the following function.
+  - `New3Command.Run()`
+- Set a breakpoint after the following block of code inside the "Run" function.
```
mlorbetske: The location of the breakpoint doesn't matter, it should be placed in the code the user is interested in rather than immediately after the code that checks for the flag

README.md
chad-ramos: Am I correct in stating that acquiring the SDK is irrelevant to building and running a 'dotnet new3' execution environment? And that acquiring the SDK only comes into play when actually doing development.  Which by way of, would make this step more relevant to the "debugging" section.

README.md
```diff
@@ -53,18 +53,39 @@ Now that we've covered the basics of using `dotnew new`, lets move on to info fo
 If you're authoring templates, or interested in contributing to this repo, then you're likely interested in how to use the latest version of this experience.
 The steps required are outlined below.
 
-Step 1: Get the SDK for your platform from [dotnet/cli](https://github.com/dotnet/cli)
+## Prerequisites
+In order to work with this project, you will need the following.
+- [Visual Studio 2017](https://www.visualstudio.com/)
```
chad-ramos: Can you comment more on you statement of "making the full build Windows only at the moment"?

mlorbetske: The build that produces the template NuGet packages currently only works on Windows (as it depends on the full framework flavor of the NuGet client). This was a larger problem before #401 because setup attempted to install those NuGet packages. Setup now installs the templates from source (in the `template_feed` directory), so there's not a build from source for personal use impact.

For users wishing to install using the template NuGet packages - they'll either need to be on Windows (to be able to produce the packages) or install them from the MyGet feed (which they'd need an appropriate NuGet config).

README.md
mlorbetske: The SDK is tricky in this regard and has a different role to play in different development scenarios.

1) Running setup at the command prompt - the appropriate SDK is downloaded into the `artifacts/.dotnet` directory and is used to do the build. This means that the globally installed SDK (if any) doesn't have influence on the build itself
2) Calling `dotnet new3` - the version of the SDK doesn't matter here, the command resolution behavior that is able to locate `dotnet-new3.exe` (or appropriate variants for other platforms) has been built in to the dotnet CLI for a very long time
3) Working with the source outside of Visual Studio - no particular SDK is required here as `setup` takes care of getting the right SDK for you
4) Working with the source in Visual Studio - this is the place where the version of the SDK matters, but the behavior isn't influenced by the version of the dotnet CLI. A portion of the SDK also ships in Visual Studio (tasks and targets) which governs what files display, when they display, the type of project that each csproj should be interpreted as being, etc. The appropriate guidance here is to use the latest available version of Visual Studio 2017.

sayedihashimi:
I will review and comment this weekend. I have a lot of hi priority items to focus on during the week.

chad-ramos:
@mlorbetske I touched on all your points except the one revolving around "full build" being Windows specific.  You mention **template NuGet packages** and how they can only be built on Windows.  Is my understanding correct when I say **template Nuget packages** equates to the default collection of templates that ship with the Templating Engine?

mlorbetske:
@chad-ramos yes, the template packages that are currently built from this repo (but only on the Windows build) are the ones that ship with the engine in `dotnet new`

Issue #410, opened 2017-03-08
Is it possible to have 2 sources in a template?


javitosanchez:
Hello,

I need to generate 2 pairs of sources (files with names) in one template. I tried to pass parameters, but that doesn't generate a file with the name replaced with the parámeter, only the replacement in inside text. Sample:

```
"sourceName": "Company.Item1"
"symbols": {
    "Item2Parameter": {
      "type": "parameter",
      "replaces": "Company.Item2"
    }
```

My source items:
Company.Item1.txt
Company.Item2.txt

Generation name: MyReplacedItem1
Parameter Item2Parameter: MyReplacedItem2

I expect:

MyReplacedItem1.txt
MyReplacedItem2.txt

I get:

MyReplacedItem1.txt
**_Company.Item2.txt_**

mlorbetske:
It would probably work for now to take off the trailing "1" in source name in the config (as Company.Item2 does not contain the string "Company.Item1" but it does contain "Company.Item"). If this doesn't meet your needs, then we'll probably need the issue about adding value forms to be completed - it'll let you apply different transforms to a value and the value it replaces so that the value in the sources (or source item names) gets replaced with the corresponding value.

javitosanchez:
The names where a sample, they are very different, I cannot apply any generic naming logic...

mlorbetske:
I see - if it's the case where you have two values that don't share a common substring between them, then the automatic renaming won't work even with the different value forms (an additional piece that allows for a file name replacements on arbitrary parameters would be needed instead).

Ex.
Source items:
a.txt
b.txt

sourceName:
a.txt

User supplied name:
Test

Cannot result in:
Test(something).txt
Test(something else).txt

If they do share a common substring, setting `sourceName` to that common substring will work.

Ex.
Source items:
xMyItema.txt
yMyItemb.txt

sourceName:
MyItem

User supplied name:
Test

Would result in:
xTesta.txt
yTestb.txt

mlorbetske:
@javitosanchez did my last reply help? If not, I'm not sure how to proceed here - perhaps we could allow other parameters to participate in rename (adding an option to parameters to allow them to match sections of file names)?

javitosanchez:
Not really. It will be solved if the parameters rename the files name.

We'll need both,  the replacement to text inside files and files name.

Thks!!

mlorbetske:
[...]

Issue #396, opened 2017-03-01
File & Folder re-naming via symbols...


chad-ramos:
I would like to see a way to rename targeted files and/or folders via symbol configuration.  

If we are able to target "variables" inside of files, the addition of targeting asset naming might offer a more rounded experience.  Because the templating engine is as flexible as it is, i.e. it does not matter what you put into your template so long as it meets the required contract, I can envision getting great use running ```dotnet new``` inside an already existing project.  In other words, bring the scope down from the project level to a level that is a component of said project. 

## Examples...

Consider the following file structure in your template...

- foo-folder/foo-controller.ts
- foo-folder/foo-service.ts

Hypothetically we would be able to "```dotnet new```" this template to scaffold a new component inside an existing project. 

###  Target individual assets

This would be the most granular approach...

```json
"symbols": {
  "foo-folder-rename": {
    "type": "folder",
    "datatype": "string",
    "replaces": "foo-folder",
    "defaultValue": "new-compnent",
    "description": "Renames component folder"
  },
  "foo-controller-rename": {
    "type": "file",
    "datatype": "string",
    "replaces": "foo-controller",
    "defaultValue": "new-controller",
    "description": "Renames foo-controller"
  },
  "foo-service-rename": {
    "type": "file",
    "datatype": "string",
    "replaces": "foo-service",
    "defaultValue": "new-service",
    "description": "Renames foo-service"
  }
}
```

### Batch target

More of an aggregated approach....

```json
"symbols": {
  "component-rename" : {
    "type": "folder-file",
    "datatype": "string",
    "replaces": "foo",
    "defaultValue": "new-compnent",
    "description": "Renames folder and files"
  }
}
```

sayedihashimi:
Hi @chad-ramos thanks for the issue. We think this sounds interesting, but we have a lot of work that we need to be focused on in the short term.
For the time being we've added this to our list of potential features for the future. Once we make some progress on features like; install/uninstall/update, sharing templates across cli/vs/vs4mac/yeoman, cleaning up help, and a few other important items.

If you're super interested in this and want to take a crack at the code we will review and discuss in a PR.

chad-ramos:
@sayedihashimi Fully understandable.  I just wanted to get my ideas out there now before I forget about them. In regards to taking a crack at it, I have been reviewing the source recently in hopes to do just that.

sayedihashimi:
@chad-ramos sounds good. We welcome community contributions.

mlorbetske:
[...]

Issue #141, opened 2016-12-24
Provide a way to specify the different forms of a value


mlorbetske:
__Background__
For name, there are currently 4 forms:
- safe_name
 - No non-word characters, trimmed
- lower_safe_name
 - Same as `safe_name` except for all lower case
- safe_namespace
 - No non-word characters except `.`
- lower_safe_namespace
 - Same as `safe_namespace` except for all lower case

__Ask__
These variations are created by composition of macros. We should provide a way to:
- Change the forms of the `name` parameter that are generated
 - These 4 shouldn't be considered "special", but should be the implicit default
- Add forms to any parameter used for replacement
- The set of allowed patterns should be able to be overridden in glob specific configurations

__What's involved__
- Forms should be specifiable as a set of operations to perform.
- Replacements generated by macros (or by this if it's not treated as being generated by a macro) should have IDs that allow them to be turned off as a group

__Algorithm__
- When the replacement map is being built
  - Do not add the original value -> parameter value entry immediately (like we do today)
  - Establish a set of "seen" original forms, make it empty initially
  - Iterate through the set of allowed forms for the value
     - Apply the transform to the original value
     - If the transformed original value has not already been "seen"
        - Add it to the set of "seen" values
        - Apply the transform to the user supplied value
        - Emit a `Replacement` operation configured to replace the augmented original value with the augmented user supplied value
     - Otherwise, continue on to the next allowed form
  - If the original value is not present in the set of "seen" values, emit a `Replacement` operation configured to replace the original value with the user supplied value

PR #444, opened 2017-03-20
Add basic support for parameters in general participating in rename

mlorbetske:
Value transforms are not supported at this point.

Fixes an issue where the end to end test harness was being recompiled on every call (slowing things down substantially)

Adds the list of created files to the output for failed end to end tests

Switches to async reading of the output streams from the end to end test harness

Standardizes all path separators in the physical file system to be forward slashes (to deal with the parent finding logic)

Extends the template parameter interface (with a non-cache intended property) to accommodate the path portion to rename

Moves the calling for the end to end test harness to a base class


mlorbetske:
This adds the basics needed for #410 and #396 but will still need #141 to be finished to be a fully working solution

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/SimpleConfigModel.cs
```diff
@@ -896,6 +866,59 @@ private List<FileSourceMatchInfo> EvaluateSources(IParameterSet parameters, IVar
...
+                        fileRenameMappings.Add(new KeyValuePair<string, string>(p.FileRename, s));
+                    }
+                }
+            }
+
+            foreach (KeyValuePair<string, string> entry in coreRenames)
+            {
+                foreach (KeyValuePair<string, string> rename in fileRenameMappings)
+                {
+                    string outRel = entry.Value.Replace(rename.Key, rename.Value);
```
seancpeters: Nit: I'm not sure what this variable name "outRel" is supposed to mean.

mlorbetske: Yep, these are the "output relative path" and "template relative path" values for a particular piece of content. I'll rename them

test/Microsoft.TemplateEngine.EndToEndTestHarness/Program.cs
```diff
@@ -198,8 +205,9 @@ private static void FirstRun(IEngineEnvironmentSettings environmentSettings, IIn
 
             string packages = Path.Combine(dir, "..", "..", "..", "..", "..", "artifacts", "packages") + Path.DirectorySeparatorChar + "*";
             string templates = Path.Combine(dir, "..", "..", "..", "..", "..", "template_feed") + Path.DirectorySeparatorChar;
+            string testTemplates = Path.Combine(dir, "test_templates") + Path.DirectorySeparatorChar;
             installer.InstallPackages(new[] { packages });
-            installer.InstallPackages(new[] { templates });
+            installer.InstallPackages(new[] { templates, testTemplates });
```
seancpeters: Should we maybe not install the real templates for tests? Changes to them could conceivably cause side-effects in the tests.

PR #470, opened 2017-03-22
More granular control over running post actions. User can be prompted…

seancpeters:
… for each one, if desired

### DIFF
--- a/src/Microsoft.TemplateEngine.Cli/New3Command.cs
+++ b/src/Microsoft.TemplateEngine.Cli/New3Command.cs
@@ -13,6 +13,7 @@ using Microsoft.DotNet.Cli.Utils;
 using Microsoft.Extensions.CommandLineUtils;
 using Microsoft.TemplateEngine.Abstractions;
 using Microsoft.TemplateEngine.Abstractions.Mount;
+using Microsoft.TemplateEngine.Cli.PostActionProcessors;
 using Microsoft.TemplateEngine.Edge;
 using Microsoft.TemplateEngine.Edge.Settings;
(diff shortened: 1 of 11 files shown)


src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -357,12 +362,13 @@ private void ConfigureEnvironment()
             return instantiateResult.Status;
         }
 
-        private string GetTemplateParameterErrorsMessage(ITemplateInfo templateInfo, out IParameterSet allParams, out IReadOnlyList<string> userParamsWithInvalidValues)
+        private string GetTemplateUsageInformation(ITemplateInfo templateInfo, out IParameterSet allParams, out IReadOnlyList<string> userParamsWithInvalidValues, out bool hasPostActionScriptRunner)
         {
             ITemplate template = EnvironmentSettings.SettingsLoader.LoadTemplate(templateInfo);
             ParseTemplateArgs(templateInfo);
-            allParams = _templateCreator.SetupDefaultParamValuesFromTemplateAndHost(template, template.DefaultName, out IList<string> defaultParamsWithInvalidValues);
+            allParams = _templateCreator.SetupDefaultParamValuesFromTemplateAndHost(template, template.DefaultName ?? "testName", out IList<string> defaultParamsWithInvalidValues);
```
mlorbetske: Is the null coalesce required here?

seancpeters: Given your changes in #469 it should no longer be necessary. I was just guaranteeing that some non-null name gets passed in.

PR #515, opened 2017-03-29
Add the restore post-action to all templates it applies to

mlorbetske:
Related to scenario 4 in the doc (kind of related to https://github.com/dotnet/cli/issues/6100)


mlorbetske:
This needs a switch to turn off auto-restore to allow tests to pass & the CLI LZMA archive creation and first-run process to work correctly

PR #518, opened 2017-03-30
Add the requested telemetry points

mlorbetske:


seancpeters:
Would we maybe want to put the telemetry logger on the host object? I'm thinking if we ever want / need to track more detailed info, it'd be easier than passing the logger object all over the place. But having it on the host object also seems a bit dubious, not all hosts would want it. Not certain what's best.

mlorbetske:
@seancpeters I thought about that - I don't think that's something we should do. We're looking for specific pieces of use information rather than internal mechanics (for which the logging mechanisms on the host object are available). We also don't want to bleed the implementation details for the dotnet CLI to other consumers, moving these down into the host object means that we have to standardize a telemetry contract for all consumers on all telemetry pipelines in all environments.

seancpeters:
Makes sense.

PR #519, opened 2017-03-30
Add the ability to display a link to 3PN on help display and create

mlorbetske:


src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -359,6 +359,12 @@ private void ConfigureEnvironment()
             {
                 case CreationResultStatus.Success:
                     Reporter.Output.WriteLine(string.Format(LocalizableStrings.CreateSuccessful, resultTemplateName));
+
```
seancpeters: Nit: This display code appears identically in 2 places, here, and below in ShowTemplateHelp(). It's pretty trivial, but could become more complex at some point. Just concerned with New3Command.vs getting more polluted.

mlorbetske: There are two (or maybe just the one, depending on the resolution of a mail thread) distinct contexts that we'll want to display this information in - Create Successful and Template Help.

PR #520, opened 2017-03-30
fixes choice / bool bug for secondary filtering of templates

seancpeters:
These cases behaved badly before the fix:
dotnet new3 mvc -o Test_C\MvcIndUld1.0 -f netcoreapp1.0 -au Individual -uld -h
dotnet new3 mvc -o Test_C\MvcIndUld1.1 -f netcoreapp1.1 -au Individual -uld -h


seancpeters:
Mike has a better fix for this.

PR #527, opened 2017-04-02
Add the uninstall command

mlorbetske:
Add support for uninstall
Add logic to release mount points (so that uninstalls can still happen after something in the thing being uninstalled has been used)
Add stdout to restore failed output
Skip downloading NuGet.exe if it's already present

### DIFF
--- a/src/Microsoft.TemplateEngine.Cli/New3Command.cs
+++ b/src/Microsoft.TemplateEngine.Cli/New3Command.cs
@@ -60,62 +60,11 @@ namespace Microsoft.TemplateEngine.Cli
             _commandInput = commandInput;
         }
 
+        public static IInstaller Installer { get; set; }
+
         public string CommandName { get; }
 
-        public static IInstaller Installer { get; set; }
-
         public string TemplateName => _commandInput.TemplateName;
-
-        public IList<string> Install => _app.InternalParamValueList("--install");
-
-        public IList<string> Uninstall => _app.InternalParamValueList("--uninstall");
-
-        public static IInstaller Installer { get; set; }
-
-        public bool IsForceFlagSpecified => _app.InternalParamHasValue("--force");
-
-        public bool InstallHasValue => _app.InternalParamHasValue("--install");
-
-        public bool UninstallHasValue => _app.InternalParamHasValue("--uninstall");
-
-        public bool IsHelpFlagSpecified => _app.InternalParamHasValue("--help");
-
-        public bool IsListFlagSpecified => _app.InternalParamHasValue("--list");
-
-        public bool IsQuietFlagSpecified => _app.InternalParamHasValue("--quiet");
-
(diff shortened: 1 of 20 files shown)


sayedihashimi:
+1

src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -662,6 +667,90 @@ private CreationResultStatus EnterInstallFlow()
             return CreationResultStatus.Success;
         }
 
+        private void ProcessUninstallRequests(IEnumerable<string> uninstallRequests, bool warnOnFailure)
```
seancpeters: I'd move this to a separate class - somewhere outside of New3Command.

src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -679,6 +768,18 @@ private CreationResultStatus EnterMaintenanceFlow()
                 return CreationResultStatus.InvalidParamValues;
             }
 
+            if (InstallHasValue &&
+                ((Install.Count > 0) && (Install[0] != null)))
+            {
+                ProcessUninstallRequests(Install.Select(x => x.Split(new[] { "::" }, StringSplitOptions.None)[0]), false);
```
seancpeters: Not sure why uninstalls are happening when the user requested installs.

src/Microsoft.TemplateEngine.Edge/Settings/TemplateCache.cs
```diff
@@ -173,25 +188,48 @@ private void ScanMountPointForTemplatesAndLangpacks(IMountPoint mountPoint, stri
                 {
                     AddTemplateToMemoryCache(template);
                 }
+
+                anythingFound |= templateList.Count > 0;
```
seancpeters: anythingFound doesn't take into account the found langpacks. I think this should be:

anythingFound |= templateList.Count > 0 || localizationInfo.Count > 0

src/Microsoft.TemplateEngine.Cli/New3Command.cs
mlorbetske: To remove anything being overridden (upgrade should remove the old package)

seancpeters: Makes sense.

src/Microsoft.TemplateEngine.Cli/New3Command.cs
mlorbetske: Moving to `Installer`

src/Microsoft.TemplateEngine.Edge/Settings/TemplateCache.cs
mlorbetske: Done

PR #427, opened 2017-03-14
Fixed #425 empty group Ids. Fixed language order in list display

seancpeters:

PR #531, opened 2017-04-03
Small README.md enhancement

JanLenoch:
Added the 'Uninstalling & Cleanup' section to the README.md file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,13 @@ To install a new template use the command:
 
     dotnet new3 -i {the path to the folder containing the templates}
 
+## Uninstalling & Cleanup
+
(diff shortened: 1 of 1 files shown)


mlorbetske:
Thanks for taking the time to figure this all out. Unfortunately, this isn't something we'd want to recommend doing, especially not with #427 being on the horizon (the way to uninstall something will be to pass -u {package name or path}). Moreover, we hope to change the way config is structured very soon (this will be a transparent change for the consumers of it). This may be interesting content for the wiki though - describing how templates are currently discovered from the caches, how contents, localizations and host customizations are found and the relationship between templates and mount points (and/or how it differs from the "component" location strategy).

PR #533, opened 2017-04-03
New command parser

seancpeters:


src/Microsoft.TemplateEngine.Cli/AppExtensions.cs
```diff
@@ -4,22 +4,32 @@
...
-using Microsoft.Extensions.CommandLineUtils;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
 namespace Microsoft.TemplateEngine.Cli
 {
     internal static class AppExtensions
     {
-        public static CommandOption Help(this CommandLineApplication app)
+        public static List<string> CreateArgListFromAdditionalFiles(IList<string> extraArgFileNames)
```
mlorbetske: Should use an interface rather than List<T>

src/Microsoft.TemplateEngine.Cli/AppExtensions.cs
```diff
@@ -4,22 +4,32 @@
...
+                {
+                    argsFlattened.AddRange(oneArg.Value);
+                }
+            }
+
+            return argsFlattened;
         }
 
-        public static IReadOnlyDictionary<string, IList<string>> ParseExtraArgs(this CommandLineApplication app, IList<string> extraArgFileNames)
+        public static Dictionary<string, IList<string>> ParseArgsFromFile(IList<string> extraArgFileNames)
```
mlorbetske: Concretes again

mlorbetske: Also, does the TValue here need to be mutable?

src/Microsoft.TemplateEngine.Cli/CommandParsing/CommandParserSupport.cs
```diff
@@ -0,0 +1,222 @@
...
+        }
+
+        public static Command CreateNewCommandWithoutTemplateInfo(string commandName) => GetNewCommand(commandName, NewCommandVisibleArgs, NewCommandHiddenArgs, DebuggingCommandArgs);
+
+        private static Command GetNewCommand(string commandName, params Option[][] args)
+        {
+            Option[] combinedArgs = CombineArrays(args);
+
+            return Create.Command(commandName,
+                           "Initialize .NET projects.",
```
mlorbetske: Should we make a localizable string out of this?

src/Microsoft.TemplateEngine.Cli/CommandParsing/CommandParserSupport.cs
```diff
@@ -0,0 +1,222 @@
...
+        }
+
+        // Final parser for when there is no template name provided.
+        // Unmatched args are errors.
+        public static Command CreateNewCommandForNoTemplateName(string commandName)
+        {
+            Option[] combinedArgs = CombineArrays(NewCommandVisibleArgs, NewCommandHiddenArgs, DebuggingCommandArgs);
+
+            return Create.Command(commandName,
+                           "Initialize .NET projects.",
```
mlorbetske: Same for the resource string

src/Microsoft.TemplateEngine.Cli/CommandParsing/CommandParserSupport.cs
```diff
@@ -0,0 +1,222 @@
...
+                           true,
+                           combinedArgs);
+        }
+
+        private static Command GetNewCommandForTemplate(string commandName, string templateName, params Option[][] args)
+        {
+            Option[] combinedArgs = CombineArrays(args);
+
+            return Create.Command(commandName,
+                           "Initialize .NET projects.",
```
mlorbetske: Same

src/Microsoft.TemplateEngine.Cli/CommandParsing/CommandParserSupport.cs
```diff
@@ -0,0 +1,222 @@
...
+        }
+
+        private static Option[] NewCommandReservedArgs
+        {
+            get
+            {
+                return new[]
+                {
+                    Create.Option("-up|--update", string.Empty, Accept.ZeroOrOneArgument()),
+                    Create.Option("-u|--uninstall", string.Empty, Accept.ZeroOrOneArgument()),
```
mlorbetske: This one should become public now

src/Microsoft.TemplateEngine.Cli/CommandParsing/CommandParserSupport.cs
```diff
@@ -0,0 +1,222 @@
...
+        {
+            get
+            {
+                return new[]
+                {
+                    Create.Option("-a|--alias", string.Empty, Accept.ExactlyOneArgument()),
+                    Create.Option("-x|--extra-args", string.Empty, Accept.OneOrMoreArguments()),
+                    Create.Option("--locale", string.Empty, Accept.ExactlyOneArgument()),
+                    Create.Option("--quiet", string.Empty, Accept.NoArguments()),
+                    Create.Option("-i|--install", string.Empty, Accept.OneOrMoreArguments()),
```
mlorbetske: This should be public too

src/Microsoft.TemplateEngine.Cli/CommandParsing/CommandParserSupport.cs
```diff
@@ -0,0 +1,222 @@
...
+
+            for (int i = 0; i < options.Length; i++)
+            {
+                variants.UnionWith(options[i].RawAliases);
+            }
+
+            return variants;
+        }
+
+        private static T[] CombineArrays<T>(params T[][] arrayList)
```
mlorbetske: Should we move this in to something in Utils or the shared files?

src/Microsoft.TemplateEngine.Cli/CommandParsing/INewCommandInput.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        IDictionary<string, IList<string>> RemainingParameters { get; }
+
+        bool TemplateParamHasValue(string paramName);
+
+        string TemplateParamValue(string paramName);
+
+        void ParseArgs();
+
+        void ReParseForTemplate(ITemplateInfo templateInfo, HostSpecificTemplateData hostSpecificTemplateData);
```
mlorbetske: Nit: lower case the "p" I think?

src/Microsoft.TemplateEngine.Cli/CommandParsing/NewCommandInputCli2.cs
```diff
@@ -0,0 +1,305 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.DotNet.Cli.CommandLine;
+using Microsoft.TemplateEngine.Abstractions;
+using Microsoft.TemplateEngine.Edge.Template;
+
+namespace Microsoft.TemplateEngine.Cli.CommandParsing
+{
+    public class NewCommandInputCli2 : INewCommandInput
```
mlorbetske: Do we still need the 2?

src/Microsoft.TemplateEngine.Cli/CommandParsing/NewCommandInputCli2.cs
```diff
@@ -0,0 +1,305 @@
...
+            _commandName = commandName;
+            _noTemplateCommand = CommandParserSupport.CreateNewCommandWithoutTemplateInfo(_commandName);
+            _currentCommand = _noTemplateCommand;
+        }
+
+        public bool HasParseError
+        {
+            get
+            {
+                return _parseResult.Errors.Count() > 0;
```
mlorbetske: Can't seem to find the definition of `ParseResult`, can we do this as `.Count` or `.Any()` if it's an IEnumerable?

src/Microsoft.TemplateEngine.Cli/CommandParsing/NewCommandInputCli2.cs
```diff
@@ -0,0 +1,305 @@
...
+                allArgs.AddRange(extraArgs);
+                _args = allArgs;
+                needsReparse = true;
+            }
+
+            IList<string> templateNameList = _parseResult.GetArgumentListAtPath(new[] { _commandName }).ToList();
+            if ((templateNameList.Count > 0) &&
+                !templateNameList[0].StartsWith("-", StringComparison.Ordinal)
+                && (_parseResult.Tokens.Count >= 2)
+                && string.Equals(templateNameList[0], _parseResult.Tokens.ToList()[1], StringComparison.Ordinal))
```
mlorbetske: Instead of `ToList()[x]` we could `.ElementAt(x)` - that's assuming of course that this is `ICollection<T>` or `IEnumerable<T>` and doesn't support indexing directly

src/Microsoft.TemplateEngine.Cli/CommandParsing/NewCommandInputCli2.cs
```diff
@@ -0,0 +1,305 @@
...
+            argsWithCommand.AddRange(_args);
+            ParserConfiguration parseConfig = new ParserConfiguration(new Option[] { _currentCommand }, argumentDelimiters: new[] { '=' }, allowUnbundling: false);
+            Parser parser = new Parser(parseConfig);
+            _parseResult = parser.Parse(argsWithCommand.ToArray());
+            _templateParamCanonicalToVariantMap = null;
+        }
+
+        public void ReParseForTemplate(ITemplateInfo templateInfo, HostSpecificTemplateData hostSpecificTemplateData)
+        {
+            List<ITemplateParameter> filteredParams = templateInfo.Parameters.Where(x => !string.Equals(x.Name, "type") && !string.Equals(x.Name, "language")).ToList();
```
mlorbetske: `StringComparison` in the `string.Equals`

src/Microsoft.TemplateEngine.Cli/AppExtensions.cs
seancpeters: Changed to IReadOnlyList

src/Microsoft.TemplateEngine.Cli/CommandParsing/NewCommandInputCli2.cs
```diff
@@ -0,0 +1,305 @@
...
+    {
+        private ParseResult _parseResult;
+        private IReadOnlyList<string> _args;
+        private string _templateNameArg;
+
+        private Command _currentCommand;
+        private Func<Task<int>> _invoke;
+
+        private IReadOnlyDictionary<string, string> _templateParamValues;
+        private IReadOnlyDictionary<string, IList<string>> _templateParamCanonicalToVariantMap;
```
mlorbetske: Does the `TValue` here need to be mutable?

src/Microsoft.TemplateEngine.Cli/CommandParsing/ParseResultExtensions.cs
```diff
@@ -0,0 +1,111 @@
...
+
+            argValue = null;
+            return false;
+        }
+
+        public static string GetArgumentValueAtPath(this ParseResult parseResult, params string[] optionsPath)
+        {
+            if (parseResult.TryTraversePath(out AppliedOption option, optionsPath) && (option.Arguments.Count > 0))
+            {
+                return option.Arguments.ToList()[0];
```
mlorbetske: Same deal with `ElementAt(x)` or direct indexing

src/Microsoft.TemplateEngine.Cli/AppExtensions.cs
seancpeters: Changed to IReadOnlyDictionary<string, IReadOnlyList<string>>

src/Microsoft.TemplateEngine.Cli/CommandParsing/CommandParserSupport.cs
seancpeters: 👍 Using the existing LocalizableStrings.CommandDescription which is "Template Instantiation Commands for .NET Core CLI"

src/Microsoft.TemplateEngine.Cli/CommandParsing/CommandParserSupport.cs
seancpeters: Not sure why we'd want this public. It's not a parse command on its own, it's just partial info that get used to dynamically build a command in other places in this class.

src/Microsoft.TemplateEngine.Cli/CommandParsing/CommandParserSupport.cs
seancpeters: Same thing as the above - this isn't a command on its own.

src/Microsoft.TemplateEngine.Cli/CommandParsing/CommandParserSupport.cs
seancpeters: Moved to Utils, new class ArrayExtensions

src/Microsoft.TemplateEngine.Cli/CommandParsing/INewCommandInput.cs
seancpeters: Sure, I don't feel strongly about it either way.

src/Microsoft.TemplateEngine.Cli/CommandParsing/NewCommandInputCli2.cs
seancpeters: We can drop it. I had it because the parser repo calls itself "A second generation of command-line parser for .NET Core command line tools" - so the 2 was for "second", but it's immaterial.

src/Microsoft.TemplateEngine.Cli/CommandParsing/NewCommandInputCli2.cs
seancpeters: It's an IEnumerable, changed to Any()

src/Microsoft.TemplateEngine.Cli/CommandParsing/NewCommandInputCli2.cs
seancpeters: Changed to ElementAt() - it's an IReadOnlyCollection

src/Microsoft.TemplateEngine.Cli/CommandParsing/NewCommandInputCli2.cs
seancpeters: D'oh! Fixed.

src/Microsoft.TemplateEngine.Cli/CommandParsing/NewCommandInputCli2.cs
seancpeters: Nope, changed to IReadOnlyDictionary<string, IReadOnlyList<string>>

src/Microsoft.TemplateEngine.Cli/CommandParsing/ParseResultExtensions.cs
seancpeters: Changed to ElementAt()

src/Microsoft.TemplateEngine.Cli/CommandParsing/CommandParserSupport.cs
seancpeters: 👍

src/Microsoft.TemplateEngine.Cli/CommandParsing/CommandParserSupport.cs
seancpeters: 👍

src/Microsoft.TemplateEngine.Cli/CommandParsing/CommandParserSupport.cs
mlorbetske: I got confused then - the collection it appears to belong to is called "NewCommandHiddenArgs", assumed it was hidden

src/Microsoft.TemplateEngine.Cli/CommandParsing/CommandParserSupport.cs
mlorbetske: Same here - this is no longer just "reserved", if the name of the collection is the incorrect thing, we should fix that instead

seancpeters: Ah, the issue is that I didn't make "install" & "uninstall" into generally visible args in this PR. I'll move them to their new (proper) place.

src/Microsoft.TemplateEngine.Cli/CommandParsing/CommandParserSupport.cs
seancpeters: Having the separate Visible & Hidden collections is mainly to keep track of what gets included in the command help output. Having the help text (2nd param) as string.Empty would also keep the hidden ones from showing up. 

The reserved list is only used for preventing those reserved flag values from getting assigned with the template param flags get assigned.

src/Microsoft.TemplateEngine.Cli/CommandParsing/NewCommandInputCli.cs
```diff
@@ -0,0 +1,307 @@
...
+
+        public string Language => _parseResult.GetArgumentValueAtPath(new[] { _commandName, "language" });
+
+        public string Locale => _parseResult.GetArgumentValueAtPath(new[] { _commandName, "locale" });
+
+        public string Name => _parseResult.GetArgumentValueAtPath(new[] { _commandName, "name" });
+
+        public string OutputPath => _parseResult.GetArgumentValueAtPath(new[] { _commandName, "output" });
+
+        public bool SkipUpdateCheck => _parseResult.HasAppliedOption(new[] { _commandName, "skip-update-check" });
```
jonsequitur: Do you have other implementations of `INewCommandInput`? If not, an interface might be unnecessary. 

Also, it's a large interface and is largely not based around behavior but simple state translation, as in the case of these properties. I think you could replace most of these properties with extension methods on `ParseResult` and make the class simpler in the process.

src/Microsoft.TemplateEngine.Utils/ArrayExtensions.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+
+namespace Microsoft.TemplateEngine.Utils
+{
+    public static class ArrayExtensions
+    {
+        public static T[] CombineArrays<T>(params T[][] arrayList)
```
jonsequitur: This method isn't needed. You can do this instead:

```
IEnumerable<string> combined = arrayList.SelectMany(a => a);
```

src/Microsoft.TemplateEngine.Cli/CommandParsing/ParseResultExtensions.cs
```diff
@@ -0,0 +1,111 @@
...
+        {
+            if (parseResult.TryTraversePath(out AppliedOption option, optionPath))
+            {
+                return option.Arguments;
+            }
+
+            return new List<string>();
+        }
+
+        private static bool TryTraversePath(this ParseResult parseResult, out AppliedOption option, params string[] optionPath)
```
jonsequitur: I'm not sure you need this. The extension method `ParseResult.AppliedCommand` will give you the deepest-level command that was applied. There can only be a single level of options under that. So if you're looking for an option on the applied command, you can get it this way:

```
parseResult.AppliedCommand()["option-name"]
```

seancpeters: Unless I'm misunderstanding, this will throw if "option-name"  was not provided. I wrote this little test to check, which throws on getting "thing":
```
        [Fact]
        public void TestAppliedOption()
        {
            Command command = Create.Command("the-command", "",
                Create.Option("--thing", "", Accept.ZeroOrOneArgument()));

            ParserConfiguration parseConfig = new ParserConfiguration(new Option[] { command });
            Parser parser = new Parser(parseConfig);
            var result = parser.Parse("the-command");

            string value = result.AppliedCommand()["thing"].ToString(); // throws here

            Assert.Equal(string.Empty, value);
        }

```

jonsequitur: I see. It looks like there are two different concerns in your function, one of which is drilling down to the command, and the second of which is checking for a specific option.

You could simplify to something like:

```
var appliedCommand = parseResult.AppliedCommand();

return appliedCommand.HasOption(option) 
    ? appliedCommand["option"].Arguments 
    : Array.Empty<string>();
```

src/Microsoft.TemplateEngine.Cli/CommandParsing/NewCommandInputCli.cs
seancpeters: It's currently the only implementation now that I've removed the old, transitional version. But we decided to keep the interface so that if we decide to replace the parser, or try to use this with another front end, it's a pretty painless swap.

src/Microsoft.TemplateEngine.Cli/CommandParsing/ParseResultExtensions.cs
seancpeters: Thanks, I will most likely take on this change at some point in the near future.

PR #537, opened 2017-04-04
Add support for explicitly specifying the usage example to use to a template

mlorbetske:


src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -1458,11 +1470,12 @@ private void ShowInvocationExamples()
 
             List<ITemplateInfo> templateList = _matchedTemplates.Select(x => x.Info).ToList();
             Reporter.Output.WriteLine("Examples:");
+            HashSet<string> usedGroupIds = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
 
             foreach (string preferredName in preferredNameList)
             {
                 ITemplateInfo template = templateList.FirstOrDefault(x => string.Equals(x.ShortName, preferredName, StringComparison.OrdinalIgnoreCase));
-                if (template != null)
+                if (template != null && usedGroupIds.Add(template.GroupIdentity ?? string.Empty))
```
seancpeters: This will cause all templates with no group identity to be eliminated from possible use after the first one is used.

src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -1473,11 +1486,16 @@ private void ShowInvocationExamples()
...
             // show up to 2 examples (total, including the above)
             Random rnd = new Random();
-            for (int i = numShown; i < ExamplesToShow && templateList.Any(); i++)
+            for (int i = numShown; i < ExamplesToShow && templateList.Count > 0; i++)
             {
                 int index = rnd.Next(0, templateList.Count - 1);
                 ITemplateInfo template = templateList[index];
-                GenerateUsageForTemplate(template);
+
+                if (usedGroupIds.Add(template.GroupIdentity ?? string.Empty) && !GenerateUsageForTemplate(template))
```
seancpeters: same as above.

src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -1458,11 +1470,13 @@ private void ShowInvocationExamples()
 
             List<ITemplateInfo> templateList = _matchedTemplates.Select(x => x.Info).ToList();
             Reporter.Output.WriteLine("Examples:");
+            HashSet<string> usedGroupIds = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
 
             foreach (string preferredName in preferredNameList)
             {
                 ITemplateInfo template = templateList.FirstOrDefault(x => string.Equals(x.ShortName, preferredName, StringComparison.OrdinalIgnoreCase));
-                if (template != null)
+                string identity = string.IsNullOrWhiteSpace(template.GroupIdentity) ? string.IsNullOrWhiteSpace(template.Identity) ? string.Empty : template.Identity : template.GroupIdentity;
```
seancpeters: I think this has similar problems to the previous, except with blank template identities. I was thinking something like:

if (template != null && 
    (string.IsNullOrEmpty(template.GroupIdentity) || usedGroupIds.Add(identity)))

mlorbetske: templates aren't allowed to have blank identities, this is an error case

seancpeters: Good point!

PR #560, opened 2017-04-07
Add reference post action implementation

seancpeters:


src/Microsoft.TemplateEngine.Cli/LocalizableStrings.resx
```diff
@@ -434,4 +434,36 @@
     <value>{0} {1}
     The default value '{1}' is not a valid value for {0}.</value>
   </data>
+  <data name="AddRefPostActionFailed" xml:space="preserve">
+    <value>Failed to add reference {0} to project file {1}</value>
+  </data>
+  <data name="AddRefPostActionMultipleProjFiles" xml:space="preserve">
+    <value>Found multiple possible project files in the same directory. Could not determine which to add a reference to. Files:</value>
```
sayedihashimi: I think "possible" isn't adding any value here. Also what do you mean by "same directory" here?

src/Microsoft.TemplateEngine.Cli/LocalizableStrings.resx
```diff
@@ -434,4 +434,36 @@
...
+  </data>
+  <data name="AddRefPostActionNoProjFileFound" xml:space="preserve">
+    <value>Unable to add reference {0}. Could not find an appropriate project file.</value>
```
sayedihashimi: In AddRefPostActionFailed you use "Failed to add...", but "Unable to add..." here. We should be consistent. I think "Unable to add..." is better.

Also I think "Could not find an apporpriate project file" would be better as "Could not find project file".

src/Microsoft.TemplateEngine.Cli/LocalizableStrings.resx
```diff
@@ -434,4 +434,36 @@
...
+  </data>
+  <data name="AddRefPostActionNoRefConfigured" xml:space="preserve">
+    <value>No references configured to be added to a project.</value>
+  </data>
+  <data name="AddRefPostActionAddProjectRef" xml:space="preserve">
+    <value>Adding a project reference. Running dotnet add {0} reference {1}</value>
```
sayedihashimi: Is it possible to run `dotnet add <name> package` as well?

src/Microsoft.TemplateEngine.Cli/LocalizableStrings.resx
```diff
@@ -434,4 +434,36 @@
...
+  </data>
+  <data name="AddRefPostActionSucceeded" xml:space="preserve">
+    <value>Successfully added
+    reference: {0}
+    to project file: {1}</value>
```
sayedihashimi: Is there going to be any line ending issues here? Does it render the same on windows/mac?

src/Microsoft.TemplateEngine.Cli/LocalizableStrings.resx
```diff
@@ -434,4 +434,36 @@
...
+  </data>
+  <data name="PostActionFailedInstructionHeader" xml:space="preserve">
+    <value>Post action failed.</value>
+  </data>
+  <data name="AddRefPostActionAddPackageRef" xml:space="preserve">
+    <value>Adding a package reference. Running dotnet add {0} reference {1}</value>
```
sayedihashimi: Should we have `package` instead of `reference` here?

src/Microsoft.TemplateEngine.Cli/LocalizableStrings.resx
```diff
@@ -434,4 +434,36 @@
...
+  </data>
+  <data name="AddRefPostActionMisconfigured" xml:space="preserve">
+    <value>Add reference action is misconfigured in the template.</value>
```
sayedihashimi: How about "Add reference action is not configured correctly in the template"?

src/Microsoft.TemplateEngine.Cli/LocalizableStrings.resx
mlorbetske: These are handled correctly by the RESX

sayedihashimi: OK thanks.

src/Microsoft.TemplateEngine.Cli/LocalizableStrings.resx
seancpeters: I'll remove possible. 
"Same directory" - the project files found were in the same directory as each other. I could output the directory path, which might make it clearer.

src/Microsoft.TemplateEngine.Cli/LocalizableStrings.resx
seancpeters: I'd contend that we should leave them as-is, they refer to considerably different things:
"Failed to add..." - this occurs when we found the project file and attempted to run 'dotnet add...' but the command failed.
"Unable to add..." - this occurs when a project file wasn't found, and so the 'dotnet add...' command was not run.

seancpeters: I cam get rid of 'an appropriate'.

src/Microsoft.TemplateEngine.Cli/LocalizableStrings.resx
seancpeters: Actually, all of the project files are already output when that string is output.

src/Microsoft.TemplateEngine.Cli/LocalizableStrings.resx
seancpeters: Yes, there's a separate string for that:
```
<data name="AddRefPostActionAddPackageRef" xml:space="preserve">
    <value>Adding a package reference. Running dotnet add {0} reference {1}</value>
</data>
```

src/Microsoft.TemplateEngine.Cli/LocalizableStrings.resx
seancpeters: Sure, changed.

src/Microsoft.TemplateEngine.Cli/LocalizableStrings.resx
seancpeters: Yes, good catch. I should also make another string which includes the version param if it's specified.

src/Microsoft.TemplateEngine.Cli/LocalizableStrings.resx
seancpeters: It might be clearer in the full context of its usage:
addrefmultiprojfiles

seancpeters:
I reworked some of the outputs to be a bit different than before. If the proj file can't be unambiguously determined (either multiple found, or none found), the user will get the same base message of:
`Unable to determine which project file to add the reference to.`
If multiple proj files are found, the user will additionally get the list of them, similar to before
```
Unable to determine which project file to add the reference to.
Project files found:
        c:\temp\Creation\scratch\TestAddRef\MyApp\MyApp.csproj
        c:\temp\Creation\scratch\TestAddRef\MyApp\MyApp2.csproj
```

src/Microsoft.TemplateEngine.Cli/LocalizableStrings.resx
sayedihashimi: ok thanks. looks good.

src/Microsoft.TemplateEngine.Cli/LocalizableStrings.resx
```diff
@@ -434,4 +434,39 @@
     <value>{0} {1}
     The default value '{1}' is not a valid value for {0}.</value>
   </data>
+  <data name="AddRefPostActionFailed" xml:space="preserve">
+    <value>Failed to add reference {0} to project file {1}</value>
+  </data>
+  <data name="AddRefPostActionNoRefConfigured" xml:space="preserve">
+    <value>No references configured to be added to a project.</value>
```
sayedihashimi: Shouldn't this be "No references configured to be added to project." so removing the 'a'. I'll leave this to @seancpeters because he is better with words than I am.

seancpeters: I agree with removing the "a"

src/Microsoft.TemplateEngine.Cli/LocalizableStrings.resx
```diff
@@ -434,4 +434,39 @@
...
+    <value>Adding a package reference. Running dotnet add {0} package {1}</value>
+  </data>
+  <data name="AddRefPostActionMisconfigured" xml:space="preserve">
+    <value>Add reference action is not configured correctly in the template.</value>
+  </data>
+  <data name="AddRefPostActionUnsupportedRefType" xml:space="preserve">
+    <value>Adding reference type {0} is not supported.</value>
+  </data>
+  <data name="AddRefPostActionAddPackageRefWithVersion" xml:space="preserve">
+    <value>Adding a package reference. Running dotnet add {0} package {1} -v {2}</value>
```
sayedihashimi: I'm fine with this string, but I'm curious if there is a long version of `-v` and if so should we be passing that instead of the short form?

seancpeters: Per the help from 'dotnet add package -h', we can use --version. I'll make that change too, it won't leave the user guessing if -v really means version.

sayedihashimi: +1 thanks

seancpeters: ...and I'll change the actual command to use --version, to be consistent.

src/Microsoft.TemplateEngine.Cli/LocalizableStrings.resx
sayedihashimi: sounds good

Issue #402, opened 2017-03-03
sourceName replacement ("." -> "_")


javitosanchez:
I have a problem in the replacement you are doing in sourceName property. The "." replacement has been doing only inside files replacements (text), but not in file names.
My problem is that I'm using the sourceName to create a solution file:

`Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "BlankProject", "BlankProject\BlankProject.csproj", "{7CF740F7-735F-48EA-8B7B-3FFA4902371C}"`

After generation (`sourceName = 'Blank.Project'`) I get:

`Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Blank_Project", "Blank_Project\Blank_Project.csproj", "{7CF740F7-735F-48EA-8B7B-3FFA4902371C}"`

but the folder & project file name is `Blank.Project\Blank.Project.csproj`

Is there a way to not have this replacement?

Thanks!!

sayedihashimi:
@javitosanchez can you change the source project name to include a "." to see if that fixes it? For example instead of the original project being named "BlankProject" try "Blank.Project" and then when creating the template try some different like "Contoso.Project".

RobertVejvoda:
I'm having the same issue. Is it possible not to replace underscores? It's quite common to create projects called My.Company.Business.Web.
Thanks!

javitosanchez:
@sayedihashimi, your solution works, but, do you have plans to change this behavior? It's quite strange to force having a "." in our sources...

sayedihashimi:
@RobertVejvoda if you modify your original source project to have a `.` in the name I believe your scenario should work.

@javitosanchez 
>your solution works, but, do you have plans to change this behavior? It's quite strange to force having a "." in our sources...

You're not strictly required to modify your sources. It's possible to get the replacements to work w/o doing that but you would have to add a bunch of content into the `template.json` file. We have some special logic which determines how to translate the SourceName (namespace value) to the correct representation when it's transformed (i.e. replace special chars with `_` or make all lowercase). It's a much harder task than it seems. To do that right we need a `.` in the value for SourceName. I think it's easier for template authors to just update project names to have a `.` (_and I'll be including that info as a best practice in docs_), but as I said it is possible to

mlorbetske:
[...]

Issue #368, opened 2017-02-24
Encoding Symbol Constants


RehanSaeed:
If I have a `constant` symbol that determines the assembly product in a `.csproj` and I enter an `&`, the ampersand needs to be XML encoded.

mlorbetske:
How do you think would be the best way to do this? Automatic escaping that the engine just does transparently, with a special directive that indicates that replacements in a particular region should be escaped in a certain way (in the content files), an alternate value form (like how name changes behavior between file name, namespace and class) or a special indication in the config?

RehanSaeed:
Not entirely sure. A default of being encoded by default might be good but with a way of turning that off.

mlorbetske:
Assuming that the value that needs to be encoded in the XML is encoded in the XML already, you can use name forms for this.

Ex.
```
{
  "symbols": {
    "x": {
      "type": "parameter",
      "dataType": "string",
      "replaces": "Test\"Value",
      "forms": [ "encode" ]
   },
  "forms": {
    "encode": {
      "identifier": "xmlEncode"
    }
  }
}
```

This will replace the literal `Test"Value` with the literal the user supplies and it will also replace the XML encoded form (`Test&quot;Value`) with the XML encoded form of the user's value.

RehanSaeed:
What if I have the following in my `template.json`:

```
    "Title": {
      "type": "parameter",
      "datatype": "string",
      "defaultValue": "Project Title",
      "replaces": "PROJECT-TITLE",
      "description": "The name of the project which determines the assembly product name. If the Swagger feature is enabled, shows the title on the Swagger UI."
    }
```

And I want to perform a replacement in my csproj:

```
<PropertyGroup Label="Package">
    <Version>1.0.0</Version>
    <Authors>PROJECT-AUTHOR</Authors>
    <Company>PROJECT-AUTHOR</Company>
    <Product>PROJECT-TITLE</Product>
    <Description>PROJECT-DESCRIPTION</Description>
    <Copyright>Copyright © PROJECT-AUTHOR. All rights Reserved</Copyright>
  </PropertyGroup>
```

Using the following command:

> dotnet new api --name Foo --Title "Foo & Bar"

When I add the `forms` from your sample code, the csproj is still output with:

```
<Product>Foo & Bar</Product>
```

seancpeters:
Note: The syntax for defining value forms in template.json has changed since this post. See #1524 for an updated example.

Issue #152, opened 2017-01-02
Investigate adding a macro that gets a free port number


mlorbetske:
Like the "now" macro, we should add a "port" macro with the following characteristics:
- A port number that is not in use by the system at the time of template instantiation should be the result
- Two port number generations cannot have the same result for the same instantiation operation
- The port numbers should be able to be bounded (low/high value)

mlorbetske:
We should be able to do something like this to get our port number, we'll still need to wrap this up in a macro though

``` C#
    private static int GetRandomFreePort(HashSet<int> alreadyReservedPorts, int lowPort = 5000, int highPort = 65536)
    {
        HashSet<int> takenPorts = new HashSet<int>();

        IPGlobalProperties globalProperties = IPGlobalProperties.GetIPGlobalProperties();
        if (globalProperties != null)
        {
            foreach (TcpConnectionInformation info in globalProperties.GetActiveTcpConnections())
            {
                takenPorts.Add(info.LocalEndPoint.Port);
            }

            foreach (IPEndPoint info in globalProperties.GetActiveTcpListeners())
            {
                takenPorts.Add(info.Port);
            }

            foreach (IPEndPoint info in globalProperties.GetActiveUdpListeners())
            {
                takenPorts.Add(info.Port);
            }
        }

        for (int i = lowPort; i <= highPort; ++i)
        {
            if((alreadyReservedPorts == null || !alreadyReservedPorts.Contains(i)) && !takenPorts.Contains(i))
            {
                return i;
            }
        }

        return -1;
    }
```

RehanSaeed:
This is some code I use to get random **free** port numbers. It knows the ports are free by trying to open them. It also supports getting ports for HTTP and HTTPS (HTTPS is within the range 44300 to 44399.

https://github.com/ASP-NET-MVC-Boilerplate/Templates/blob/master/Source/Boilerplate.FeatureSelection/Services/PortService.cs

peterblazejewicz:
I was pleased with Core using conventions: `5000` port by default, debugger with different default port. I use other tools/templates and I know there are: 3000, 4200 or 5000 ports set by default with options like `-p --port` to change this at runtime. Hey, the 44300 is in fact a VS tooling convention printed in books (`By default, IIS Express uses the 44300 port to publish over SSL`) and blogged: http://www.hanselman.com/blog/WorkingWithSSLAtDevelopmentTimeIsEasierWithIISExpress.aspx

mlorbetske:
[...]

PR #561, opened 2017-04-07
The basics of value forms

mlorbetske:


src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/Macros/Config/ProcessValueFormMacro.cs
```diff
@@ -0,0 +1,26 @@
...
+            Forms = forms;
+        }
+
+        public string VariableName { get; }
+
+        public string Type => "processValueForm";
+
+        public string SourceVariable { get; }
+
+        public string Form { get; }
```
seancpeters: Nit: Maybe call this "FormName"? It's a bit confusing having both Form and Forms.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/Macros/ProcessValueFormMacro.cs
```diff
@@ -0,0 +1,58 @@
...
+                {
+                    value = string.Empty;
+                }
+            }
+            else
+            {
+                value = working?.ToString() ?? "";
+            }
+
+            value = realConfig.Forms[realConfig.Form].Process(realConfig.Forms, value);
```
seancpeters: It's not immediately clear that the Forms dictionary will always contain the key realConfig.Form.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/Macros/Config/ProcessValueFormMacro.cs
mlorbetske: Done

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/Macros/ProcessValueFormMacro.cs
mlorbetske: Switching to TryGetValue & emitting a diagnostic message with "authoring" as the category in the case that it's not found

mlorbetske:
Fixes #141 
Fixes #152
Fixes #368 
Fixes #396 
Fixes #402 
Fixes #410

src/Microsoft.TemplateEngine.Core.Contracts/IGlobalRunSpec.cs
```diff
@@ -18,7 +18,7 @@ public interface IGlobalRunSpec
 
         IReadOnlyDictionary<string, IReadOnlyList<IOperationProvider>> LocalizationOperations { get; }
 
-        string PlaceholderFilename { get; }
+        IReadOnlyList<string> IgnoreFileNames { get; }
```
seancpeters: This could potentially be a hashset, since it appears its main use is via Contains(). But in general, it will probably only have a few entries max, so a list is probably as performant.

PR #583, opened 2017-04-13
Add _ViewStart and _ViewImports items

rynowak:


template_feed/Microsoft.DotNet.Common.ItemTemplates/content/ViewImports/.template.config/template.json
```diff
@@ -0,0 +1,27 @@
...
+  "precedence": "100",
+  "identity": "Microsoft.AspNetCore.Mvc.ViewImports",
+  "shortName": "viewimports",
+  "sourceName": "_ViewImports",
+  "primaryOutputs": [
+    {
+      "path": "_ViewImports.cshtml"
+    }
+  ],
+  "defaultName": "_ViewImports",
```
rynowak: /cc @seancpeters - right now this is always outputing the file `content.cshtml`. What did I do wrong here?

seancpeters: I'm a bit confused. The only file in this template (other change the config json files) is called _ViewImports.cshtml. Do you mean that it's getting its name changed to content.cshtml?

Can you share the exact command line that you're using to invoke the template?

rynowak: Yes.
```
dotnet new viewimports -i D:\dotnet\templating\template_feed\Microsoft.DotNet.Common.ItemTemplates\content 
```

Stop by if you want a live demo 😆

rynowak: What would be ideal here is that the name isn't contextual at all unless the user explicitly overrides it.

seancpeters: Ok, if you set the sourceName to some string that doesn't appear anywhere in the template file names or content, it won't have any effect, maybe something like "xxx".

Also, we chatted a bit about the defaultName in templates. That is the final fallback for the "name".
The name resolution is:
1) The vale of the --name input (if it has a value)
2) The directory the template is being made in, which is the last directory in the -o path, or the current directory if no -o flag is given.
3) If #2 doesn't have a value (which can pretty much only happen if creating a template  in the *nix root directory '/'), then the defaultName is used.

rynowak:
@seancpeters updated

mlorbetske:
Rebased & moved the templates into the web item templates

PR #591, opened 2017-04-14
fixed display & filtering issues with type filtering. Fixed a display…

seancpeters:
… issue with invalid language


src/Microsoft.TemplateEngine.Edge/Settings/FilteredTemplateInfo.cs
```diff
@@ -19,7 +19,9 @@ public FilteredTemplateInfo(ITemplateInfo info, IReadOnlyList<MatchInfo> matchDi
 
         public bool IsMatch => MatchDisposition.Count > 0 && !MatchDisposition.Any(x => x.Kind == MatchKind.Mismatch);
 
-        public bool IsPartialMatch => MatchDisposition.Any(x => x.Kind != MatchKind.Mismatch);
+        public bool IsPartialMatch => MatchDisposition.Any(x => x.Kind != MatchKind.Mismatch)
+            && (!MatchDisposition.Any(x => x.Location == MatchLocation.Context)
```
mlorbetske: I think this can be rewritten as `MatchDisposition.All(x => x.Location != MatchLocation.Context || (x.Location == MatchLocation.Context  && x.Kind == MatchKind.Exact))` to avoid iterating the set twice (albeit partially because of the any)

seancpeters: 👍

PR #598, opened 2017-04-16
Add Azure AD B2C auth option to MVC starter web template

danroth27:


danroth27:
@mlorbetske @sayedihashimi @vibronet @gsacavdm Please review

danroth27:
Regarding the individual auth condition change, the individual auth condition now represents one of two possible sub options (local vs B2C) just like the org auth condition does (single vs multitenant). I think it makes sense for these auth conditions to be consistent. I did diff the local individual option with the template output before and after this change to ensure that it's output is the same.

template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/StarterWeb-CSharp/OpenIdConnectOptionsSetup.cs
```diff
@@ -0,0 +1,72 @@
...
+            AzureAdB2COptions = b2cOptions.Value;
+        }
+
+        public AzureAdB2COptions AzureAdB2COptions { get; set; }
+
+        public void Configure(OpenIdConnectOptions options)
+        {
+            options.ClientId = AzureAdB2COptions.ClientId;
+            options.Authority = AzureAdB2COptions.Authority;
+
```
gsacavdm: Can we add UseTokenLifetime = true?

PR #599, opened 2017-04-16
[WIP] BuildWebHost method for tooling

glennc:
@DamianEdwards @bricelam @ajcvickers

We talked about this as a way to allow EF tooling to still function, and even close some of the gaps it has today. How do people feel about it now they see it? @DamianEdwards in particular since you were not here when we talked about it.

### DIFF
--- a/template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/EmptyWeb-CSharp/Program.cs
+++ b/template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/EmptyWeb-CSharp/Program.cs
@@ -10,15 +10,17 @@ namespace Company.WebApplication1
 {
     public class Program
     {
-        public static void Main(string[] args)
-        {
-            var host = new WebHostBuilder()
+        public static IWebHostBuilder BuildWebHost() => new WebHostBuilder()
                 .UseKestrel()
                 .UseContentRoot(Directory.GetCurrentDirectory())
                 .UseIISIntegration()
                 .ConfigureLogging(loggerFactory => loggerFactory.AddConsole())
                 .UseStartup<Startup>()
                 .Build();
+
+        public static void Main(string[] args)
+        {
+            var host = BuildWebHost();
 
             host.Run();
         }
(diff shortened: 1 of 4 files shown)


glennc:
I used the expression methods deliberately to get some feedback. I like how it subtly discourages you from modifying the method outside of changing the builder. Since doing anything other than modifying the builder is going to have side effects when you run tooling that is something we probably should encourage.

But if we think it is too new/strange syntax to put in a template I am ok with removing it.

@davidfowl @eilon do you have any comments here? I forgot to CC you in the initial thing.

If everyone is ok I will merge this and it will be done.

template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/EmptyWeb-CSharp/Program.cs
```diff
@@ -10,16 +10,18 @@ namespace Company.WebApplication1
 {
     public class Program
     {
-        public static void Main(string[] args)
-        {
-            var host = new WebHostBuilder()
+        public static IWebHostBuilder BuildWebHost(string[] args) => new WebHostBuilder()
                 .UseKestrel()
```
Eilon: Can unindent these one level?

template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/EmptyWeb-CSharp/Program.cs
```diff
@@ -10,16 +10,18 @@ namespace Company.WebApplication1
...
                 .UseContentRoot(Directory.GetCurrentDirectory())
                 .UseIISIntegration()
                 .ConfigureLogging(loggerFactory => loggerFactory.AddConsole())
                 .UseStartup<Startup>()
                 .Build();
 
+        public static void Main(string[] args)
+        {
+            var host = BuildWebHost(args);
+
```
Eilon: Remove empty line?

glennc:
Closing as this will be done in https://github.com/dotnet/templating/pull/671 now that everyone is on board.

PR #600, opened 2017-04-16
Add files for 1.x F# version of the empty web template

lambdakris:
Hi, I created a 1.x F# version of the empty web template. 

I was able to manually test installing the template with new3 and test using the template with new3. 

I went for the 1.x version since that seemed like the more immediate need...though I suppose if it makes it's way into the tooling, it will be on the same release as the 2.0 templates. 

If this pr is acceptable, or with feedback I am able to get it to an acceptable state, I can work on the 2.0 version of the template as well as the 1.x and 2.0 version of the web api templates.


mlorbetske:
Looks good, one minor thing is that a couple files don't seem to end with a newline. I don't know F# myself, so I'd be more comfortable if we could get a review from someone with more expertise than me in that area.

sayedihashimi:
Let's see if @cartermp would like to review these

cartermp:
+1, code looks good!

sayedihashimi:
:shipit:

Issue #602, opened 2017-04-17
Add support for the correct form of "else if" (#elif)


mlorbetske:
Per https://msdn.microsoft.com/en-us/library/88td0y52.aspx, what we call `#elseif` should actually be `#elif`. We should _add_ support for this, leaving support for `#elseif` as-is to avoid breaking existing templates.

PR #604, opened 2017-04-17
Added 'elif' as a variant of 'elseif'.

seancpeters:
This adds more internal support for having multiple variants of each conditional keyword.
Not done: allowing template.json configurations to specify multiple forms of the same keyword in a single configuration.


mlorbetske:
Fixes #602

PR #608, opened 2017-04-18
Add back options namespace for individual user accounts template

danroth27:


danroth27:
@mlorbetske Oops! Sorry!

PR #631, opened 2017-04-21
Enable existing B2C for MVC & WebAPI

mlorbetske:


template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/StarterWeb-CSharp/.template.config/vs-2017.3.host.json
```diff
@@ -24,7 +24,8 @@
     },
     {
       "auth": "Individual",
-      "authenticationType": "IndividualAuth"
+      "authenticationType": "IndividualAuth",
```
danroth27: should this be IndividualB2CAuth?

mlorbetske: Nope, IndividualAuth gets repurposed when one or more B2C options are present.

PR #654, opened 2017-04-23
Add support for the Framework parameter from the CLI to the 1.x F# Class Library template

lambdakris:
This pull request adds a Framework parameter to template.json and dotnetcli.host.json files under the 1.x ClassLibrary-FSharp template


mlorbetske:
Thanks!

PR #659, opened 2017-04-24
Update WebAPI template to Auth 2.0

HaoK:
@mlorbetske @danroth27

### DIFF
--- a/template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/WebApi-CSharp/Startup.cs
+++ b/template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/WebApi-CSharp/Startup.cs
@@ -27,6 +27,10 @@ namespace Company.WebApplication1
         {
             // Add framework services.
             services.AddMvc();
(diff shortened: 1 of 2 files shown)


danroth27:
Could you please also add the missing namespace in `JwtBearerOptionsSetup` so that the Web API templates compile (my bad)?

```c#
using Microsoft.AspNetCore.Authentication.JwtBearer;
```

template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/WebApi-CSharp/Startup.cs
```diff
@@ -27,6 +27,10 @@ public void ConfigureServices(IServiceCollection services)
         {
             // Add framework services.
             services.AddMvc();
+#if (OrganizationalAuth || IndividualB2CAuth)
+            services.AddJwtBearerAuthentication();
+
```
danroth27: The blank line should be before the call to `services.AddJwtBearerAuthentication();`, not after.

HaoK:
Fixed, I think we are good to go, @mlorbetske can you merge these changes for us?

Issue #625, opened 2017-04-20
dotnet new3 Error on Linux/Mac


rschiefer:
I'm trying to test "dotnet new3" on Linux/Mac but received the following error on both:

> $ dotnet new3 -l
> 
> Unhandled Exception: System.ComponentModel.Win32Exception: No such file or directory
>    at Interop.Sys.ForkAndExecProcess(String filename, String[] argv, String[] envp, String cwd, Boolean redirectStdin, Boolean redirectStdout, Boolean redirectStderr, Int32& lpChildPid, Int32& stdinFd, Int32& stdoutFd, Int32& stderrFd)
>    at System.Diagnostics.Process.StartCore(ProcessStartInfo startInfo)
>    at System.Diagnostics.Process.Start()
>    at Microsoft.DotNet.Cli.Utils.Command.Execute()
>    at Microsoft.DotNet.Cli.Program.ProcessArgs(String[] args, ITelemetry telemetryClient)
>    at Microsoft.DotNet.Cli.Program.Main(String[] args)
> Aborted (core dumped)

I can run "dotnet new" without issue:

> $ dotnet new -l
> Template Instantiation Commands for .NET Core CLI.
> 
> Usage: dotnet new [arguments] [options]
> 
> Arguments:
>   template  The template to instantiate.
> 
> 
> Templates          Short Name       Language      Tags
> ----------------------------------------------------------
> Nuget Config       nugetconfig                    Config
> Web Config         webconfig                      Config
> Solution File      sln                            Solution

I first ran setup.sh to setup the new3 command from a customized fork, then pulled from the main repo and reran setup.sh.  Received the above error for both branches.  It may be that I'm not properly setting up the new3 command.  I did try the harderreset.sh script but no change.

My fork - https://github.com/rschiefer/templating
Main repo - https://github.com/dotnet/templating

mlorbetske:
It looks like this trace is from the dotnet cli itself, which build of the cli are you running? The difference between new and new3 from the cli perspective is that new calls into these assemblies directly, while new3 must be started as an external process.

rschiefer:
"dotnet --version" returns 1.0.1 on Mac and Linux.  My Windows machine says "1.0.0-rc4-004917".  

What version do I need?

rschiefer:
Not sure if this helps but I ran dotnet from the ".dotnet" folder which is the same version as is working in Windows and got the same error on Linux.

> $ .dotnet/dotnet new3 -l
> 
[...]

PR #660, opened 2017-04-24
Added support for creating templates from Git URL

rschiefer:
Haven't been able to verify on Mac/Linux yet but works well on Windows.

### DIFF
--- a/src/Microsoft.TemplateEngine.Cli/GitSource.cs
+++ b/src/Microsoft.TemplateEngine.Cli/GitSource.cs
@@ -28,18 +28,18 @@ namespace Microsoft.TemplateEngine.Cli
             }
             else
             {
-                var index = gitIndex + 4;
-                var indexOfLastSlashBeforeGit = -1;
-                var indexOfSlash = -1;
+                int index = gitIndex + 4;
+                int indexOfLastSlashBeforeGit = -1;
+                int indexOfSlash = -1;
                 while ((indexOfSlash = spec.IndexOf('/', indexOfLastSlashBeforeGit + 1)) < index && indexOfSlash != -1)
                 {
                     indexOfLastSlashBeforeGit = indexOfSlash;
                 }
 
-                var gitUrl = spec.Substring(0, index);
-                var subFolder = spec.Substring(index);
+                string gitUrl = spec.Substring(0, index);
+                string subFolder = spec.Substring(index);
                 subFolder = subFolder.Trim('/');
-                var repoName = gitUrl.Substring(indexOfLastSlashBeforeGit + 1).Replace(".git", string.Empty);
+                string repoName = gitUrl.Substring(indexOfLastSlashBeforeGit + 1).Replace(".git", string.Empty);
                 package = new GitSource(gitUrl, subFolder, repoName);
                 return true;
             }
--- a/src/Microsoft.TemplateEngine.Cli/GitSource.cs
+++ b/src/Microsoft.TemplateEngine.Cli/GitSource.cs
@@ -1,0 +1,48 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.TemplateEngine.Cli
+{
+    internal class GitSource
+    {
+        public string GitUrl { get; set; }
+        public string SubFolder { get; set; }
+        public string RepositoryName { get; set; }
+
+        public GitSource(string gitUrl, string subFolder, string repoName)
+        {
+            GitUrl = gitUrl;
+            SubFolder = subFolder;
+            RepositoryName = repoName;
+        }
+
+        public static bool TryParseGitSource(string spec, out GitSource package)
+        {
+            package = null;
+            int gitIndex = -1;
+
+            if (string.IsNullOrEmpty(spec) || (gitIndex = spec.IndexOf(".git", StringComparison.OrdinalIgnoreCase)) < 0)
+            {
+                return false;
+            }
+            else
+            {
+                var index = gitIndex + 4;
+                var indexOfLastSlashBeforeGit = -1;
+                var indexOfSlash = -1;
+                while ((indexOfSlash = spec.IndexOf('/', indexOfLastSlashBeforeGit + 1)) < index && indexOfSlash != -1)
+                {
+                    indexOfLastSlashBeforeGit = indexOfSlash;
+                }
+
+                var gitUrl = spec.Substring(0, index);
+                var subFolder = spec.Substring(index);
+                subFolder = subFolder.Trim('/');
+                var repoName = gitUrl.Substring(indexOfLastSlashBeforeGit + 1).Replace(".git", string.Empty);
+                package = new GitSource(gitUrl, subFolder, repoName);
+                return true;
+            }
+        }
+    }
+}
--- a/src/Microsoft.TemplateEngine.Cli/GitSource.cs
+++ b/src/Microsoft.TemplateEngine.Cli/GitSource.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Text;
+using System.Text.RegularExpressions;
 
 namespace Microsoft.TemplateEngine.Cli
 {
@@ -26,7 +27,8 @@ namespace Microsoft.TemplateEngine.Cli
             Uri uri = null;
 
             if (string.IsNullOrEmpty(spec) 
-                || Uri.TryCreate(spec, UriKind.Absolute, out uri) == false
+                || (Uri.TryCreate(spec, UriKind.Absolute, out uri) == false
+                    && Regex.IsMatch(spec, ".*@.*:.*") == false)
                 || (gitIndex = spec.IndexOf(_gitExtension, StringComparison.OrdinalIgnoreCase)) < 0)
             {
                 return false;
--- a/src/Microsoft.TemplateEngine.Cli/GitSource.cs
+++ b/src/Microsoft.TemplateEngine.Cli/GitSource.cs
@@ -21,8 +21,11 @@ namespace Microsoft.TemplateEngine.Cli
         {
             package = null;
             int gitIndex = -1;
+            Uri uri = null;
 
-            if (string.IsNullOrEmpty(spec) || (gitIndex = spec.IndexOf(".git", StringComparison.OrdinalIgnoreCase)) < 0)
+            if (string.IsNullOrEmpty(spec) 
+                || Uri.TryCreate(spec, UriKind.Absolute, out uri) == false
+                || (gitIndex = spec.IndexOf(".git", StringComparison.OrdinalIgnoreCase)) < 0)
             {
                 return false;
             }
--- a/src/Microsoft.TemplateEngine.Cli/GitSource.cs
+++ b/src/Microsoft.TemplateEngine.Cli/GitSource.cs
@@ -6,6 +6,8 @@ namespace Microsoft.TemplateEngine.Cli
 {
     internal class GitSource
     {
+        static readonly string _gitExtension = ".git";
+
         public string GitUrl { get; set; }
         public string SubFolder { get; set; }
         public string RepositoryName { get; set; }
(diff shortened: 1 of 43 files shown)


dansiegel:
just tested this on the Mac, it still has the same problem as issue #625 with the install, however I am able to install a template from a git url with these changes.

test/Microsoft.TemplateEngine.Core.UnitTests/StubSettingsLoader.cs
```diff
@@ -0,0 +1,81 @@
+﻿using Microsoft.TemplateEngine.Abstractions;
+using System;
+using System.Collections.Generic;
+using Microsoft.TemplateEngine.Abstractions.Mount;
+
+namespace Microsoft.TemplateEngine.Core.UnitTests
+{
+    internal class StubSettingsLoader : ISettingsLoader
```
mlorbetske: Is there already one of these in the mocks project?

src/Microsoft.TemplateEngine.Cli/Installer.cs
```diff
@@ -189,6 +200,23 @@ private void InstallRemotePackages(List<Package> packages)
...
+        private void InstallGitSources(List<GitSource> gitSources)
+        {
+            _paths.CreateDirectory(_paths.User.ScratchDir);
+
+            List<string> newLocalPackages = new List<string>();
+
+            foreach (GitSource source in gitSources)
+            {
+                string targetPath = $"{_paths.User.ScratchDir}/{source.RepositoryName}";
+                ExecuteProcess("git", "clone", source.GitUrl, targetPath);
```
mlorbetske: Not a big fan of this, if you don't have git on the path, you can't use the feature. The alternative would be to embed a git client, but that's not very attractive either. I'd really rather get to a place where you can install installers, where this would light up for people who explicitly acquire it

src/Microsoft.TemplateEngine.Cli/GitSource.cs
```diff
@@ -0,0 +1,48 @@
...
+            package = null;
+            int gitIndex = -1;
+
+            if (string.IsNullOrEmpty(spec) || (gitIndex = spec.IndexOf(".git", StringComparison.OrdinalIgnoreCase)) < 0)
+            {
+                return false;
+            }
+            else
+            {
+                int index = gitIndex + 4;
```
mlorbetske: In cases like these, I prefer declaring a local const for the string (in this case ".git") and accessing the length instead of having the value of the length appearing as a magic number

src/Microsoft.TemplateEngine.Cli/GitSource.cs
```diff
@@ -0,0 +1,48 @@
...
+            SubFolder = subFolder;
+            RepositoryName = repoName;
+        }
+
+        public static bool TryParseGitSource(string spec, out GitSource package)
+        {
+            package = null;
+            int gitIndex = -1;
+
+            if (string.IsNullOrEmpty(spec) || (gitIndex = spec.IndexOf(".git", StringComparison.OrdinalIgnoreCase)) < 0)
```
mlorbetske: Feels like a good first check would be with Uri.TryParse & UriKind.Absolute

src/Microsoft.TemplateEngine.Cli/Installer.cs
mlorbetske: The exit code of the process isn't checked, though it's returned by the helper

src/Microsoft.TemplateEngine.Cli/Installer.cs
```diff
@@ -237,5 +265,21 @@ private void InstallLocalPackages(IReadOnlyList<string> packageNames)
                 }
             }
         }
+
+
+        internal virtual bool ExecuteProcess(string command, params string[] args)
```
mlorbetske: There's a helper for executing stuff at the command line that we should use instead. It handles the redirection of the output streams and all that.

src/Microsoft.TemplateEngine.Utils/PhysicalFileSystem.cs
```diff
@@ -44,6 +44,10 @@ public void FileCopy(string source, string target, bool overwrite)
 
         public void DirectoryDelete(string path, bool recursive)
         {
+            foreach(string file in Directory.GetFiles(path, "*", SearchOption.AllDirectories))
```
mlorbetske: This doesn't seem like a good thing to do, if the user marked a file as readonly or If it's marked as system, I'd say that failure is probably correct.

test/Microsoft.TemplateEngine.Cli.UnitTests/InstallerTests.cs
```diff
@@ -0,0 +1,89 @@
+﻿using Microsoft.TemplateEngine.Abstractions;
+using Microsoft.TemplateEngine.Core.UnitTests;
+using Microsoft.TemplateEngine.Utils;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+using System.Linq;
+using Xunit;
+using Microsoft.TemplateEngine.Abstractions.PhysicalFileSystem;
+using System;
```
mlorbetske: Nit: usings are out of order

test/Microsoft.TemplateEngine.Cli.UnitTests/Microsoft.TemplateEngine.Cli.UnitTests.csproj
```diff
@@ -16,6 +16,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\Microsoft.TemplateEngine.Core.UnitTests\Microsoft.TemplateEngine.Core.UnitTests.csproj" />
```
mlorbetske: Assuming this is for the mock if settings loader, if there isn't one in mocks already, it should be added there & shared through the reference to mocks

test/Microsoft.TemplateEngine.Core.UnitTests/StubSettingsLoader.cs
```diff
@@ -0,0 +1,81 @@
...
+    {
+        public IComponentManager Components => throw new NotImplementedException();
+
+        public IEngineEnvironmentSettings EnvironmentSettings => throw new NotImplementedException();
+
+        public IEnumerable<MountPointInfo> MountPoints => throw new NotImplementedException();
+
+        public void AddMountPoint(IMountPoint mountPoint)
+        {
+            throw new NotImplementedException();
```
mlorbetske: If we're adding a new mock, we should go all the way here, others that are trying to test extensions to the engine wouldn't have a good experience with all the throws

test/Microsoft.TemplateEngine.Core.UnitTests/StubSettingsLoader.cs
rschiefer: I only see two implementations of the ISettingsLoader interface the real implementation and this stub I added.

test/Microsoft.TemplateEngine.Core.UnitTests/StubSettingsLoader.cs
rschiefer: I had thought that if no one has needed it thus far they probably wouldn't going forward and YAGNI.  Is it acceptable to have empty methods or return default values for all methods?  The other reason I didn't fully implement is I'm less familiar with this class and how to fully mock it out for testing purposes.

src/Microsoft.TemplateEngine.Cli/Installer.cs
rschiefer: Yeah, I had the same thought but figured most folks will have git so we are incrementally adding value here.  I like the Installer idea but its a larger design/development effort.  Do you think this is a must have at this point?

I will add a check for the exit code.

src/Microsoft.TemplateEngine.Cli/Installer.cs
rschiefer: I saw that but wasn't sure if I should use it.  I will swap over to this.

src/Microsoft.TemplateEngine.Cli/GitSource.cs
rschiefer: Great idea.  I will add it.

src/Microsoft.TemplateEngine.Cli/GitSource.cs
rschiefer: Yes, I like that much better. Will change it.

src/Microsoft.TemplateEngine.Utils/PhysicalFileSystem.cs
rschiefer: Okay.  I only added this because it wasn't letting me delete my scratch folder with the cloned git repo.

I will find another solution.

test/Microsoft.TemplateEngine.Cli.UnitTests/InstallerTests.cs
rschiefer: Will fix.

test/Microsoft.TemplateEngine.Cli.UnitTests/Microsoft.TemplateEngine.Cli.UnitTests.csproj
rschiefer: Understood, will fix.

rschiefer:
Thanks for the feedback Mike!  This week is slammed but I will try to have these fixed by EOD Friday or sooner.

src/Microsoft.TemplateEngine.Cli/Installer.cs
rschiefer: Well I went back and found 2 options in Dotnet.cs and ProcessStartPostActionProcessor.cs.  Neither of these appear to be generic implementations that I could use from the Installer class.  Is there another?

rschiefer:
All changes complete except for those related to outstanding questions above:

-I didn't see a generic Execute method I could reuse.
-Is the refactor/redesign to create Installers required in the first release?

mlorbetske:
@rschiefer we're currently transitioning branches, please retarget this against `rel/vs2017/3-Preview2`. Thanks

rschiefer:
Rebased to rel/vs2017/3-Preview2.

mlorbetske:
Thanks!

motowilliams:
I assumed the syntax for SSH was just to swap out the URI.
```
 .\dotnet-new3.exe -i git@github.com:sayedihashimi/dotnet-new-samples.git
Error: Could not install "git@github.com:sayedihashimi/dotnet-new-samples.git".
```
But that doesn't seem to be the case.

rschiefer:
I think I can fix that tonight.  Stay tuned!

rschiefer:
SSH connections are now supported.

motowilliams:
Any update / progress on this item?

mlorbetske:
Unfortunately, no. The main issue here is that there's an implicit dependency added to git being present and on the path. Ultimately, this means that something that would ship with the engine potentially doesn't work by default - maybe that's OK if the error case is well documented (not currently the case) and has instructions for how to appropriately configure the system to make it work. 

The thinking on this when last it was visited was to componentize the installer & being able to add the ability to install from git (or other sources) by installing a NuGet package carrying that support. However, the installer is not properly extensible right now & the work to make it be has not yet been prioritized. #1335 may also be required for that work to be completed.

dansiegel:
@mlorbetske just a thought here, but it seems to me that if you don't have git installed then you are far less likely to try installing via git. I think we both agree someone is bound to attempt it and get an error. IMO this could be handled by simply displaying an error message stating the obvious...

mlorbetske:
@dansiegel I don’t necessarily disagree with anything you’ve said, but there are a fair number of people who’ll copy & paste a command from a blog post & expect things to just work. While instructions to install git being present in the error (and having them vary appropriately by platform) would be helpful, I’m still not a fan of something that could generate an error by default. On the other hand, doing the work to fix installer to allow proper extension and creating an extension that uses the logic in this PR gets kind of the best of both worlds - the ability to install from git repos and explicit knowledge that there are additional installation requirements before that functionality can be used (and presumably a error message/instructions in the case that git cannot be found). I’ve got a branch with the installer work in progress and with these changes ported in to an extension, it does still need some cleanup before it’s ready to go though.

motowilliams:
Is this feature request dead then?

vijayrkn:
Not necessarily. There were a lot of pending PRs against a branch that is not actively maintained. if this PR can be re-targeted to master, then some one can take a look.

PR #661, opened 2017-04-24
Adding netcore class lib templates

jmarolf:
As part of making templates in dotnet/templating the one-true-source for templates we need to add coreclr class library templates to match what was shipped in VS.

related PR that depends on this one: https://github.com/dotnet/project-system/pull/2053

### DIFF
--- a/template_feed/Microsoft.DotNet.Common.ProjectTemplates.1.x/content/ClassLibrary-CoreClr-CSharp/.template.config/template.json
+++ b/template_feed/Microsoft.DotNet.Common.ProjectTemplates.1.x/content/ClassLibrary-CoreClr-CSharp/.template.config/template.json
@@ -1,0 +1,60 @@
+{
+  "author": "Microsoft",
+  "classifications": ["NetCore", "Library"],
+  "name": "Class library",
+  "generatorVersions": "[1.0.0.0-*)",
+  "groupIdentity": "Microsoft.NetCore.Library",
+  "precedence": "100",
+  "identity": "Microsoft.NetCore.Library.CSharp",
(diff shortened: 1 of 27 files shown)


template_feed/Microsoft.DotNet.Common.ProjectTemplates.1.x/content/ClassLibrary-CoreClr-CSharp/.template.config/template.json
```diff
@@ -0,0 +1,60 @@
+{
+  "author": "Microsoft",
+  "classifications": ["NetCore", "Library"],
+  "name": "Class library",
```
mlorbetske: This will be confusing to have two class library entries in the list at the CLI. We're working on a feature that should make this set of variations unnecessary. @seancpeters

mlorbetske:
The engine side of this is coming with #663, I'll add the VSTemplate handling tonight

mlorbetske:
Added the baselines for class libraries in #667 

In the VSTemplate, you'll be able to pass a wizard data entry now that specifies which baseline to use, like
```
<CustomParameter Name="$set:baseline$" Value="app" />             <-- Targets netcoreapp

- or -

<CustomParameter Name="$set:baseline$" Value="standard" />        <-- Targets netstandard
```

jmarolf:
@mlorbetske That is music to my ears.  I will close this in anticipation if that change getting merged in.  I assume I can depend on that being in the 15.3 preview?

mlorbetske:
Yep, we'll be preparing the insertions today

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #678, opened 2017-04-26
Post action: add projects to sln file

seancpeters:
Also added a placeholder for the baseline param to TemplateEngineWizard.cs
It'll need to be wired up to acquire its value from the VS side.
The bootstrapper, which is called by the wizard, was already setup to use the passed in baseline.


seancpeters:
We may want to hold off on merging this due to the bug in dotnet sln add [#6413](https://github.com/dotnet/cli/issues/6413)

mlorbetske:
Let's resolve the merge conflict & then merge to `rel/vs2017/3-Preview2`, the add to referenced issue can be sorted out after the fact

PR #705, opened 2017-04-28
Update ApplicationDbContextFactory

bricelam:
Depends on aspnet/EntityFramework#8314

@Eilon @DamianEdwards @ajcvickers

### DIFF
--- a/template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/StarterWeb-CSharp/Data/ApplicationDbContextFactory.cs
+++ b/template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/StarterWeb-CSharp/Data/ApplicationDbContextFactory.cs
@@ -9,7 +9,7 @@ namespace Company.WebApplication1.Data
 {
     public class ApplicationDbContextFactory : IDbContextFactory<ApplicationDbContext>
     {
-        public ApplicationDbContext Create(DbContextFactoryOptions options) =>
-            Program.BuildWebHost(Array.Empty<string>()).Services.GetRequiredService<ApplicationDbContext>();
+        public ApplicationDbContext Create(string[] args) =>
+            Program.BuildWebHost(args).Services.GetRequiredService<ApplicationDbContext>();
     }
 }


bricelam:
@Eilon So nice that I'm starting to second guess if we even need to do this automatically...

Thoughts, @DamianEdwards?

Eilon:
Hmmm @DamianEdwards ?

Why wouldn't we want this?

prafullbhosale:
:eyes:

template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/StarterWeb-CSharp/Data/ApplicationDbContextFactory.cs
```diff
@@ -9,7 +9,7 @@ namespace Company.WebApplication1.Data
 {
     public class ApplicationDbContextFactory : IDbContextFactory<ApplicationDbContext>
     {
-        public ApplicationDbContext Create(DbContextFactoryOptions options) =>
-            Program.BuildWebHost(Array.Empty<string>()).Services.GetRequiredService<ApplicationDbContext>();
+        public ApplicationDbContext Create(string[] args) =>
+            Program.BuildWebHost(args).Services.GetRequiredService<ApplicationDbContext>();
```
bricelam: May change our signature to `IEnumerable<string>`. Should I call `.ToArray()` here or change the signature of `BuildWebHost` too?

bricelam: Given that both [WebHost.CreateDefaultBuilder](https://github.com/aspnet/MetaPackages/blob/rel/2.0.0-preview1/src/Microsoft.AspNetCore/WebHost.cs#L150) and [IConfigurationBuilder.AddCommandLine](https://github.com/aspnet/Configuration/blob/rel/2.0.0-preview1/src/Microsoft.Extensions.Configuration.CommandLine/CommandLineConfigurationExtensions.cs#L21) (which it calls) are `string[]`, we should keep it `string[]`.

bricelam:
@mlorbetske upstream PRs have been merged. This one is good to go.

mlorbetske:
@bricelam These require new packages, right?

danroth27:
Yup, the latest release build has these changes

PR #706, opened 2017-04-28
Fix indenting in Program.cs

danroth27:
@DamianEdwards @mlorbetske


danroth27:
@mlorbetske ready to merge

PR #727, opened 2017-05-03
Adds FSharp WebAPI Template

odytrice:
Adds FSharp WebAPI Template


mlorbetske:
Thanks @odytrice!

mlorbetske:
It looks like this may be impacted by #760

Issue #730, opened 2017-05-03
Re-add VB templates


mlorbetske:

mlorbetske:
Not needed, the PR was abandoned

PR #731, opened 2017-05-03
Temporarily remove VB templates

mlorbetske:
@livarcocc @srivatsn 

Filed https://github.com/dotnet/templating/issues/730 to track re-adding when the support is there


srivatsn:
Will this remove them from VS as well?

mlorbetske:
VS would no longer install them, but the VSTemplate file would also need to be removed

srivatsn:
No we need this in VS. VS fully supports building VB projects. The CLI doesn't have the VB compiler yet. So we need to only selectively remove these templates from the CLI - not from VS.

mlorbetske:
OK, closing this PR & associated bug. We need to add an engine feature to allow hiding these. @seancpeters is starting on that now.

mlorbetske:
Bug for the new feature is #734

PR #754, opened 2017-05-05
Exclude urlRewrite.config when not doing auth

danroth27:


danroth27:
should be ready to merge

Issue #729, opened 2017-05-03
Xamarin Templates


dansiegel:
I was working on some Xamarin templates and was running into a couple of issues that hopefully someone might have some ideas on how to best handle. The core project is converted over to NetStandard1.4 using the new csproj format, but the Android and iOS projects are still using the older format. I'm sure I'm missing something simple. 

1. I want the template to generate an `#if DEBUG` in the iOS template to enable testing. but obviously this gets stripped out by the templating engine. Is there a way to get the templating engine to ignore that directive?

2. I have a XAML file that needs to change a namespace import depending on which choice they made but I haven't been able to get any conditional lines to work. I had been trying something like the following but it gets totally ignored and the generated project looks exactly the same. 

``` xml
<prism:PrismApplication xmlns="http://xamarin.com/schemas/2014/forms"
             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
<!--if UseAutofac -->
             xmlns:prism="clr-namespace:Prism.Autofac;assembly=Prism.Autofac.Forms"
<!--elif UseDryIoc -->
             xmlns:prism="clr-namespace:Prism.DryIoc;assembly=Prism.DryIoc.Forms"
<!--elif UseNinject -->
             xmlns:prism="clr-namespace:Prism.Ninject;assembly=Prism.Ninject.Forms"
<!--else -->
             xmlns:prism="clr-namespace:Prism.Unity;assembly=Prism.Unity.Forms"
<!--endif -->
             x:Class="MobileApp.App">
``` 

3. Similar to the XAML issue is for the packages.config on the platform project

sayedihashimi:
For the #if debug it sounds like the same issue as https://github.com/dotnet/templating/issues/587

dansiegel:
@sayedihashimi that was the missing link... since the issue is already being tracked I'm going to close this issue.

dansiegel:
@sayedihashimi it looks like I'm still running into a couple issues with my templates.

1) I haven't been able to prevent the templating engine from removing elements or blocks in my csproj file that require conditional statements. See the [iOS csproj](https://github.com/dansiegel/Prism-Templates/blob/master/content/MobileApp-CSharp/src/MobileApp.iOS/MobileApp.iOS.csproj#L4) for an example. I have tried wrapping these areas with `-:cnd` and `+:cnd` without any change in behavior

[...]

PR #762, opened 2017-05-06
Add app.config, web.config, web.*.config, packages.config and *.sh

mlorbetske:


mlorbetske:
Related to #282 and #729

PR #787, opened 2017-05-13
Constrain routes

TerribleDev:
Since the id paramter is a value type, and has a default value. Those actions will invoke even when the route contains a string, such as `/api/values/yo`. To the average user, its not obvious this behavior will occur. 

This also outlines an extra feature to users, that may not realize you can constrain routes on types.


glennc:
In general I think we prefer letting the actions be called and getting a model state error or something of that sort rather than the action just not even being called. That might not be super applicable for the web api template though. @DamianEdwards @rynowak for comment.

rynowak:
My opinion, a 400 is better than a 404. The constraints are only good for disambiguation, a 404 in this case doesn't tell a client what they did wrong.

TerribleDev:
@rynowak @glennc I think you have convinced me, and I think that actually makes more sense.  The one thing that I don't like is `/api/values/yo` would end up being treated the same as `/api/values/0` when semantically it is not the same, especially if you used a `0`based index for your identifiers.

Issue #698, opened 2017-04-27
Enhance EntityFramework Provider options in Templates


dansiegel:
The templates currently provide a possible outcome for generating a template using either SqlServer or Sqlite. I think that this could be further enhanced by providing choices for the EntityFramework Provider. 

- SqlServer => Microsoft.EntityFrameworkCore.SqlServer
- SQLite => Microsoft.EntityFrameworkCore.SQLite
- MySQL => MySql.Data.EntityFrameworkCore
- PostgreSQL => Npgsql.EntityFrameworkCore.PostgreSQL
- In Memory => Microsoft.EntityFrameworkCore.InMemory

sayedihashimi:
I would like to expand the list of options as well. I would love some community help in this area.

sayedihashimi:
Any community members interested in sending a PR?

dansiegel:
I'm happy to take a look at this over the weekend.

sayedihashimi:
@dansiegel thanks! I'll mark the item as assigned. Let me know if you need any info or decide to move on from the feature.

sayedihashimi:
@dansiegel it looks like I can assign this item to you. I'll assign it to you, so that we know someone is looking at it.

dansiegel:
For MySql there are actually three different providers. The Official MySql package is also the only one that has yet to release a stable package. Is there a preference to only support the Official MySql provider, all three, or .....

dansiegel:
@sayedihashimi another question for you, and I suppose this could be an issue of it's own as well.

Will the `2017.3.host.json` need updating so that you can choose the EF Provider from within Visual Studio? From looking at the spec it seems like the host file is pretty much set for custom template dialogs centered around web projects. I don't know if maybe this would be an area where you could work with the VS team to expand the spec so Templates can provide a more customized new Project Dialog based on the ever changing needs of the templates.

mlorbetske:
@dansiegel we are the VS team for that too 😉 We're doing a bunch of work right now to provide a UI that can adapt to the options the template has available & some of those capabilities will flow back in to the web project creation (OneASP.NET) dialog. This could certainly be one of those capabilities.

sayedihashimi:
FYI we are having discussions with the team that owns the broader File->New Project experience to get 1st class support for Template Engine templates. Those discussions are more long term discussions. The discussions recently got started. I will keep you posted on progress.

dansiegel:
[...]

PR #788, opened 2017-05-13
add EFProviders #698

dansiegel:
This isn't ready to be merged. @sayedihashimi @mlorbetske please let me know if there is anything you think should be changed.

- [x] Add Ef Provider Symbol
- [x] Set the proper provider when configuring the ApplicationDbContext
- [ ] Validate the migrations
- [ ] Test

### DIFF
--- a/template_feed/Microsoft.DotNet.Web.ProjectTemplates.1.x/content/StarterWeb-CSharp/.template.config/template.json
+++ b/template_feed/Microsoft.DotNet.Web.ProjectTemplates.1.x/content/StarterWeb-CSharp/.template.config/template.json
@@ -30,7 +30,7 @@
           ]
         },
         {
-          "condition": "(UseLocalDB || !IndividualAuth)",
+          "condition": "(!UseSQLiteProvider || !IndividualAuth)",
           "exclude": [
             "*.db"
           ]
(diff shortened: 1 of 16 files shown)


dansiegel:
@danroth27 @phenning please let me know if the naming looks good, or if there are any changes that you want. Currently I have all three MySQL providers in there, I'm not sure if that would be too confusing... since the official package is only pre release, I'm not sure if that's something that people want verses Sapient or Pomelo that have "stable" packages

danroth27:
@divega

divega:
@danroth27 thanks. I don't have enough context to do a good job reviewing or understanding what this does, but maybe @bricelam can.

dansiegel:
@divega the goal is to provide an option of what EF Provider to use when creating a new project instead of making an assumption that you want SQLServer or SQLite. Hope that helps

template_feed/Microsoft.DotNet.Web.ProjectTemplates.1.x/content/StarterWeb-CSharp/.template.config/template.json
```diff
@@ -106,6 +106,42 @@
...
+        {
+          "choice": "PomeloMySql",
+          "description": "This database provider allows Entity Framework Core to be used with MySQL. The provider is maintained as part of the Pomelo Foundation Project."
+        },
+        {
+          "choice": "SapientGuardian",
+          "description": "This database provider allows Entity Framework Core to be used with MySQL. The provider is maintained as part of the Sapient Guardian Project."
+        }
+      ],
+      "defaultValue": "InMemory",
```
bricelam: Feels like SQLite might be a better default since it...
1. Actually persists data
2. Doesn't require a database server
3. Runs everywhere

template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/StarterWeb-CSharp/.template.config/dotnetcli.host.json
```diff
@@ -6,8 +6,9 @@
       "longName": "target-framework-override",
       "shortName": ""
     },
-    "UseLocalDB": {
-      "longName": "use-local-db"
+    "EFProvider": {
+      "longName": "ef-provider",
+      "shortName": "ef"
```
bricelam: Feels like these should be something like `db-provider` and `db` so they can be reused across templates regardless of data access technology.

template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/StarterWeb-CSharp/.template.config/template.json
```diff
@@ -267,11 +267,61 @@
...
+    "EFProvider": {
       "type": "parameter",
-      "datatype": "bool",
-      "defaultValue": "false",
-      "description": "Whether or not to use LocalDB instead of SQLite"
+      "datatype": "choice",
+      "choices": [
+        {
+          "choice": "InMemory",
+          "description": "This database provider allows Entity Framework Core to be used with an in-memory database. This is useful when testing code that uses Entity Framework Core. The provider is maintained as part of the EntityFramework GitHub project."
```
bricelam: I'd assert stronger that this provider is only intended for testing (and even then, we recommend using SQLite in-memory databases). Maybe it shouldn't even be an option...

template_feed/Microsoft.DotNet.Web.ProjectTemplates.1.x/content/StarterWeb-CSharp/.template.config/template.json
dansiegel: I can do that. Was thinking In Memory since it's quick, requires no setup and will just work regardless of the platform.

template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/StarterWeb-CSharp/.template.config/dotnetcli.host.json
dansiegel: I had that originally, then changed it since there are three providers for MySQL I thought that might be confusing.

template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/StarterWeb-CSharp/.template.config/template.json
dansiegel: Maybe `This should ONLY be used for testing`?

seancpeters:
@danroth27 - The changes in this PR would now need to be applied to the templates in the aspnet/templating repo. So if they're still desired, this PR will need to be ported there. 

I'm closing this PR in the templating repo, but it'll still be available for reference if needed.

danroth27:
@divega @ajcvickers @Eilon Should this change be ported to the https://github.com/aspnet/templating repo?

ajcvickers:
@danroth27 See discussion on https://github.com/aspnet/templating/issues/26. I think this is basically the same idea, so I don't think we need to port it.

PR #802, opened 2017-05-16
Removed the logger factory in Startup.Configure

davidfowl:
It's not in the other templates and isn't used. Looks like left overs.


mlorbetske:
Thanks!

PR #803, opened 2017-05-16
Remove user secrets ID from empty template

davidfowl:
Do we need it? We don't/can't even include the `Microsoft.Extensions.SecretManager.Tools` cli tool.

/cc @glennc @DamianEdwards


mlorbetske:
This was here to prevent warnings that were generated during build looking for the attribute. We can remove it if that's no longer happening.

PR #805, opened 2017-05-16
Added missing "AccessDenied" view for account controller in StarterWe…

thrundi:
There was missing a view for account controller. It's basicly the same view as in Areas/IdentityService/Views/Account. 
I'm not sure if those views should be shared (and how) because they belong to different areas.


sayedihashimi:
@danroth27?

danroth27:
LGTM

sayedihashimi:
@mlorbetske can you merge when ready?

PR #812, opened 2017-05-17
Add Uninstall to bootstrapper

phenning:
@mlorbetske


phenning:
@mlorbetske take another look with the readonly changes please.

Issue #633, opened 2017-04-21
Add Remove Unused Using Statements Post-Action


RehanSaeed:
Right now, I have to add `#if` pre-processor directives around my using statements to add or remove certain features. In the example below, `System.Linq` could be used by multiple conditional symbols and it's very hard to work that out without generating the template using every combination of symbol values and then trying to build the project. 

```
#if (Versioning)
    using System.Linq;
#endif
#if (CORS)
    using ApiTemplate.Constants;
#endif
    using Boilerplate.AspNetCore;
    using Microsoft.AspNetCore.Builder;
    using Microsoft.AspNetCore.Hosting;
    using Microsoft.AspNetCore.Http;
    using Microsoft.AspNetCore.Mvc;
#if (Versioning)
    using Microsoft.AspNetCore.Mvc.ApiExplorer;
#endif
    using Microsoft.AspNetCore.Mvc.Infrastructure;
    using Microsoft.AspNetCore.Mvc.Routing;
#if (HttpsEverywhere)
    using Microsoft.AspNetCore.Rewrite;
#endif
    using Microsoft.Extensions.Configuration;
    using Microsoft.Extensions.DependencyInjection;
    using Microsoft.Extensions.Logging;
    using Newtonsoft.Json;
    using Newtonsoft.Json.Converters;
```

In my experience with my own templating engine, this is the easist way to make mistakes. A good way around this problem is to get rid of these `#if` directives and use a post-action to remove unused using statements.

sayedihashimi:
In the asp.net templates we purposefully leave some extra using statements. I don't think this would be something that we would work on, since I don't think we would use it directly. Let's see if @mlorbetske would be interested in taking a PR for this.

mlorbetske:
I'm interested, obviously it'd have to be an opt-in behavior. If we get a PR for this that'd be fantastic.

dansiegel:
@mlorbetske I think it would be safe as a default if duplicates were automatically removed. That would seem to resolve the issue that @RehanSaeed and I both have with our templates. Only removing unused namespaces should be an opt in feature... and I would say more often than not I'm adding some extra namespaces to make life easier when you start coding.

sayedihashimi:
For VS and `dotnet new` we leave extra using statements so that users can use types without needing to add the usage. Since this is by design, this isn't an item that the team at Microsoft will be working on. If there is a community member, or team, that want's to implement this as a Post Action that can be added to templates we will review that.
[...]

PR #848, opened 2017-05-23
Proof of concept for remove duplicated usings and sort usings post-actions

mlorbetske:
Related to https://github.com/dotnet/templating/issues/633

Open issues:
* This won't work at all for VB or F# in its current state
* Existing and newly created content isn't differentiated between
* There's a ton of redundancy in the operations
* Since post-actions can't currently be parameterized, niceties like flowing the environment's "place system usings first" setting into the action won't work
* The post-action processing interface isn't exposed in abstractions leading to this awkward workaround with creating a component to expose it & consuming it in the post-action processor (this also means that post-actions can't be generator specific)

### DIFF
--- a/src/Microsoft.TemplateEngine.Core/RemoveDuplicateUsingsPostAction.cs
+++ b/src/Microsoft.TemplateEngine.Core/RemoveDuplicateUsingsPostAction.cs
@@ -1,0 +1,60 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using Microsoft.TemplateEngine.Abstractions;
+using Microsoft.TemplateEngine.Core.Util;
+
+namespace Microsoft.TemplateEngine.Core
+{
+    public class RemoveDuplicateUsingsPostAction : ICommonPostActionHandler
+    {
+        public static readonly Guid ActionProcessorId = new Guid("C405DDB2-4F64-4A2D-8005-1243D5C55EAD");
+
+        public Guid Id => ActionProcessorId;
+
+        public bool Process(IEngineEnvironmentSettings settings, IPostAction actionConfig, ICreationResult templateCreationResult, string outputBasePath)
+        {
+            foreach(string file in Directory.EnumerateFiles(outputBasePath, "*.cs", SearchOption.AllDirectories))
+            {
+                byte[] fileData = File.ReadAllBytes(file);
+                Encoding fileEncoding = EncodingUtil.Detect(fileData, fileData.Length, out byte[] bom);
+                string allText = fileEncoding.GetString(fileData.Skip(bom.Length).ToArray());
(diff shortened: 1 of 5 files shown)


src/Microsoft.TemplateEngine.Core/RemoveDuplicateUsingsPostAction.cs
```diff
@@ -0,0 +1,60 @@
...
+            }
+
+            return true;
+        }
+
+        private static bool IsUsingDirective(string line)
+        {
+            string trimmed = line.Trim();
+            return trimmed.StartsWith("using ", StringComparison.Ordinal) && trimmed.EndsWith(";", StringComparison.Ordinal);
+        }
```
seancpeters: Duplicates could be missed if they only differ in the amount of whitespace between the keyword 'using' and the thing being used.

seancpeters: This could be gotten around by having the seenUsings only store / check the thing being used, without the literal 'using' or intermediate whitespace.

src/Microsoft.TemplateEngine.Core/SortUsingsPostAction.cs
```diff
@@ -0,0 +1,107 @@
...
+            }
+
+            return true;
+        }
+
+        private static bool IsUsingDirective(string line)
+        {
+            string trimmed = line.Trim();
+            return trimmed.StartsWith("using ", StringComparison.Ordinal) && trimmed.EndsWith(";", StringComparison.Ordinal);
+        }
```
seancpeters: This method is duplicated in RemoveDuplicateUsingsPostAction. In general we may want to abstract it to be based on the file type.

seancpeters: This may also want to exclude anything with parentheses, for odd cases like:
`using (Foo f = bar.GetFoo()) do { <stuff-with-f> } while(<condition>);`

mlorbetske:
@seancpeters the reason for the new post-action interface is called out in the description of the PR. There's no way currently of having generator specific post-actions because there's no post action handling interface in the abstractions project & in this particular case, the desire was to use some stuff defined in core (that can't be referenced directly higher up the chain). The interface is defined at a much higher level in the stack.

seancpeters:
Ah, thanks, I got it now.

RehanSaeed:
@vijayrkn Any changes of making this real?

vijayrkn:
> @vijayrkn Any changes of making this real?

Most of the pending PRs were either work in progress or against the stabilize branch. Wanted to make sure we had an accurate list of pending PRs and PRs against master.  This specific work-item is not currently being worked on but if we get a completed PR against master, we can get it reviewed and merged.

Issue #381, opened 2017-02-27
"dotnet new" can generate invalid code if directory name is not a valid C# namespace identifier


TheRealPiotrP:
_From @mikeharder on February 23, 2017 21:52_

## Steps to reproduce
```
docker run -it --rm microsoft/dotnet:1.0.3-sdk-msbuild-rc4
mkdir new
cd new
dotnet new console
dotnet restore
dotnet build
```

## Expected  behavior
Builds successfully.

## Actual behavior
```
Program.cs(3,11): error CS1001: Identifier expected [/new/new.csproj]
Program.cs(3,11): error CS1514: { expected [/new/new.csproj]
Program.cs(3,11): error CS0116: A namespace cannot directly contain members such as fields or methods [/new/new.csproj]
Program.cs(4,1): error CS1022: Type or namespace definition, or end-of-file expected [/new/new.csproj]
```

The root cause is `dotnet new` uses the directory name as the namespace in the generated code, **without** first checking if the directory name is a valid namespace identifier.

```
using System;

namespace new
{
    class Program
    {
        static void Main(string[] args)
        {
            Console.WriteLine("Hello World!");
        }
    }
}
```

## Environment data
`dotnet --info` output:

```
.NET Command Line Tools (1.0.0-rc4-004883)

Product Information:
 Version:            1.0.0-rc4-004883
 Commit SHA-1 hash:  fa2c9e025c

Runtime Environment:
 OS Name:     debian
 OS Version:  8
 OS Platform: Linux
 RID:         debian.8-x64
 Base Path:   /usr/share/dotnet/sdk/1.0.0-rc4-004883
```

_Copied from original issue: dotnet/cli#5825_

mlorbetske:
As a workaround, you can specify the -n switch to give a name that isn't a C# keyword. We'll beef up the name cleansing logic.

donJoseLuis:
This issue was last touched some years ago.  We are working on a new delivery road map.  Please reopen if this is something we want & we'll properly assess its' priority compared to other work aimed at improving the overall templating UX.

PR #867, opened 2017-05-31
Value forms for name are less special, can be overridden

seancpeters:
Primary changes:
1) The default name is setup as a symbol in SimpleConfigModel.FromJObject(), and used unless explicitly overridden from template configuration. If the template config provides name symbol configuration, without value forms, the default name forms are used. But if the explicitly defined name symbol defines values forms, those are used exclusively for name.
2) Direct ParameterSymbol replacements are now setup via the identity value form (internal to SimpleConfigModel).
3) Parameters with value forms explicitly defined in the template are not given an identity replacement. This is a break from past behavior, in which a direct replacement was setup, in addition to any explicitly defined value forms.

Ancillary changes:
1) Fixed a bug in TrieEvaluator.cs so that if multiple terminals exist as matches to the same pattern, the effective terminal is the last-in, instead of the first in.
2) Modified a unit test to deal with the fact that parameters with value forms explicitly defined no longer get an identity replacement.
3) Modified TemplateCreator.InstantiateAsync so that separate IParameterSets are used for GetCreationEffects() and CreateAsync(). Using the same set was causing duplication of macros produced.

### DIFF
--- a/src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/ParameterSymbol.cs
+++ b/src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/ParameterSymbol.cs
@@ -3,11 +3,54 @@ using System.Collections.Generic;
 using Microsoft.TemplateEngine.Abstractions;
 using Microsoft.TemplateEngine.Utils;
 using Newtonsoft.Json.Linq;
+using Microsoft.TemplateEngine.Orchestrator.RunnableProjects.ValueForms;
 
 namespace Microsoft.TemplateEngine.Orchestrator.RunnableProjects
 {
     public class ParameterSymbol : ISymbolModel
     {
+        // Used when the template explicitly defines the symbol "name".
+        // The template definition is used exclusively, except for the case where it doesn't define any value forms.
+        // When that is the case, the default value forms are used.
+        //
+        // When we add file-specific forms, this'll need some work.
+        public static ParameterSymbol ExplicitNameSymbolMergeWithDefaults(ISymbolModel templateDefinedName, ISymbolModel defaultDefinedName)
+        {
+            if (!(templateDefinedName is ParameterSymbol templateSymbol))
+            {
+                throw new InvalidCastException("templateDefinedName is not a ParameterSymbol");
+            }
+
+            if (!(defaultDefinedName is ParameterSymbol defaultSymbol))
+            {
+                throw new InvalidCastException("defaultDefinedName is not a ParameterSymbol");
+            }
+
+            if (templateSymbol.Forms.GlobalForms.Count > 0)
+            {   // template symbol has forms, use them
+                return templateSymbol;
+            }
+
+            ParameterSymbol mergedSymbol = new ParameterSymbol()
+            {
+                Binding = templateSymbol.Binding,
+                DefaultValue = templateSymbol.DefaultValue,
+                Description = templateSymbol.Description,
+                Forms = defaultSymbol.Forms,    // this is the only thing that gets replaced from the default
+                IsRequired = templateSymbol.IsRequired,
+                Type = templateSymbol.Type,
+                Replaces = templateSymbol.Replaces,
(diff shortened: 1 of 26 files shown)


mlorbetske:
@dotnet-bot test OSX10.12 x64 Release Build please

mlorbetske:
Is this good to go? We haven't talked about this one in a while. I seem to recall that there's a behavior change in here, let's talk through it again tomorrow - at this point, I think we'd want to keep the behavior with what's currently out there the same.

seancpeters:
After talking offline, I'm going to change the default setup to include the identity form as the first value form for name - to avoid the breaking change.

seancpeters:
The second commit deals with the issues of breaking behavior:
1) Trie creation will now filter duplicated match patterns, with a last in wins behavior. This does the same thing as filtering at match time, but is a slight performance increase since the filtering only needs to occur once.
2) All symbols with value forms defined (including name) get the "identity" value form prepended to their list. This can be overridden with the `"addIdentity": false` configuration option when using the extended value form configuration format.

Notable change:
The syntax of value form configuration in template.json now has 2 separate syntaxes:

Note: in the examples below, "global" is used. But we'll be extending this to allow
conditional forms, which will have other names.
The same format will be used for other named form definitions.

```
Simple:
"forms": {
   "global": [ <strings representing value form names> ]
}

Detailed:
"forms" {
   "global": {
     "forms": [ <strings representing value form names> ],
     "addIdentity": <bool default true>,
     // other future extensions, e.g. conditionals
   },
```

If the symbol doesn't include an "identity" form and the addIdentity flag isn't false,
an identity value form is added to the beginning of the symbol's value form list.
If there is an identity form listed, its position remains intact irrespective of the addIdentity flag.

This setup allows for backwards compatibility with existing configuration processing prior to this change.

mlorbetske:
Should we make the fix for #381 as part of this or in a subsequent PR?

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/SimpleConfigModel.cs
```diff
@@ -925,7 +925,12 @@ private static ISymbolModel SetupDefaultNameSymbol(string sourceName)
   ""description"": ""The default name symbol"",
   ""datatype"": ""string"",
   ""forms"": {
-    ""global"": [ ""identity"", ""safe_name"", ""lower_safe_name"", ""safe_namespace"", ""lower_safe_namespace""]
+    ""global"": [ """ + IdentityValueForm.FormName
```
seancpeters: I went back & forth on how to best format this too. Offline we decided to keep as is.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/ParameterSymbol.cs
```diff
@@ -1,13 +1,56 @@
...
+        //
+        // When we add file-specific forms, this'll need some work.
+        public static ParameterSymbol ExplicitNameSymbolMergeWithDefaults(ISymbolModel templateDefinedName, ISymbolModel defaultDefinedName)
+        {
+            if (!(templateDefinedName is ParameterSymbol templateSymbol))
+            {
+                throw new InvalidCastException("templateDefinedName is not a ParameterSymbol");
+            }
+
+            if (!(defaultDefinedName is ParameterSymbol defaultSymbol))
```
seancpeters: Yes, but what it's cast to is used later in the method.

seancpeters:
I think it'd be best to work on #381 in a separate PR, so this doesn't get too big.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/ParameterSymbol.cs
mlorbetske: Ah, didn't see the symbol after the is, oops

PR #868, opened 2017-05-31
Update to allow uppercase policies

jaxxon33:
As per https://github.com/Azure-Samples/active-directory-b2c-dotnetcore-webapp/commit/445d3d5ea0cdf9534f26f68fe65fed8753621bd0


danroth27:
We're going to move to using separate auth schemes for each policy instead of trying to manipulate the issuer.

jaxxon33:
Thank you

PR #904, opened 2017-06-07
removed lead dot-slashes from db files in appsettings

seancpeters:


seancpeters:
Fixes https://github.com/aspnet/Templates/issues/840

PR #912, opened 2017-06-07
Update to SetupCrossgen 167 - Coherence 25532

mlorbetske:
@mikeharder


mikeharder:
What's the purpose of this PR?  For getting ASP.NET into the CLI, we should wait until the next ASP.NET build based on the latest shared framework, which hasn't built yet.

PR #930, opened 2017-06-12
Remove MvcRazorCompileOnPublish

davidfowl:
Why isn't it on by default in the web SDK?

/cc @pranavkm @DamianEdwards @mlorbetske @rynowak


rynowak:
We wanted to enable it for new projects, but without rolling it out to a bunch of existing projects that upgrade to 2.0.0. Talk to @DamianEdwards about it

davidfowl:
Yea I think we should just default it. Will discuss with @DamianEdwards

mlorbetske:

PR #935, opened 2017-06-13
Remove size for form inputs

Petermarcu:
Making this PR based on comments from Damian. @DamianEdwards


DamianEdwards:
This is going to require some reactive changes to size the fields using bootstrap grid classes instead. I actually think we should change to using the standard form layout instead of horizontal too to simplify it even further.

Petermarcu:
Ok, we can close this PR while you guys figure out what the desired outcome is :)

PR #953, opened 2017-06-15
the wildcard remove of F# Compile items is not need.

enricosada:
the F# sdk will disable implicit wildcard include of Compile items for F#

@mlorbetske i dont know if this the right target branch for the PR. my idea is for the current sdk 2.0 (preview2) and next vs (15.3? please @KevinRansom  @cartermp to correct me about this)

/cc @KevinRansom @cartermp


enricosada:
as a note, the error with implicit included wildcard, was an issue in `2.0.0-preview2-006127` but is ok in >=  `2.0.0-preview2-006391`

cartermp:
cc @brettfo who is working in this area while Kevin is out

Pilchie:
Also tagging @srivatsn and @livarcocc

PR #960, opened 2017-06-16
Allow other operators to process unhandled condition expressions

### LINKED ISSUE #952: csproj import condition not replaced.
When using .NET templating with a solution containing a shared project, all references to the template app are scrubbed on a `dotnet new`, except for conditional imports in csproj files.

## Reproduction Steps

1. Create a new console app project.
2. Add a shared project to the solution.
3. Reference the shared project from the console app.
4. Add generic `.template.config` to replace namespaces and such.
5. Install template.
6. `dotnet new template-name`.
7. The reference in the console app's csproj is updated to import the correct project, but the [`Condition`](https://msdn.microsoft.com/en-us/library/92x05xfs.aspx) attribute is not.

Example, where BlankApp is the template app, and HeyThere is the app created via `dotnet new`:

```
 <Import Project="..\HeyThere\HeyThere.projitems" Label="Shared" Condition="Exists('..\BlankApp\BlankApp.projitems')" />
```

This causes apps not to build, even though a reference is properly there.

mlorbetske:
Fixes #952 

The current behavior for "rejected"/"failed" (not evaluated to false - but things like encountering unknown symbols) situations for conditions simply takes the bytes that were collected in gathering the condition contents and emits them directly.

With this change, the collected condition byte are "injected" as a chunk just prior to the current position of the source stream. Since the source stream shouldn't be considered to be mutable, a stream that combines streams has been added to allow reading the chunk the condition had read from the buffer/combined from one or more portions of buffers before continuing to read from the original source stream.

Currently, after the combined stream is applied, it's never removed - meaning that if there were multiple conditions rejected, there'd be one combining stream for each rejected condition (and associated MemoryStreams holding the chunk to be read) - this should be changed such that after a read that moves `_source1` to the depleted state, the combining stream is removed and the original source stream (`_source2`) is restored on the caller.

### DIFF
--- a/src/Microsoft.TemplateEngine.Core.Contracts/IProcessorState.cs
+++ b/src/Microsoft.TemplateEngine.Core.Contracts/IProcessorState.cs
@@ -33,6 +33,6 @@ namespace Microsoft.TemplateEngine.Core.Contracts
 
         void SeekBackWhile(ITokenTrie match);
 
-        void Inject(Stream staged, int sequenceNumberEffect);
+        void Inject(Stream staged);
     }
 }
--- a/src/Microsoft.TemplateEngine.Core.Contracts/IProcessorState.cs
+++ b/src/Microsoft.TemplateEngine.Core.Contracts/IProcessorState.cs
@@ -1,3 +1,4 @@
+using System.IO;
 using System.Text;
 
 namespace Microsoft.TemplateEngine.Core.Contracts
@@ -31,5 +32,7 @@ namespace Microsoft.TemplateEngine.Core.Contracts
         void SeekBackUntil(ITokenTrie match, bool consume);
 
         void SeekBackWhile(ITokenTrie match);
+
+        void Inject(Stream staged, int sequenceNumberEffect);
     }
 }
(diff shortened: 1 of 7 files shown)


mlorbetske:

mlorbetske:

src/Microsoft.TemplateEngine.Core.Contracts/IProcessorState.cs
```diff
@@ -31,5 +32,7 @@ public interface IProcessorState
         void SeekBackUntil(ITokenTrie match, bool consume);
 
         void SeekBackWhile(ITokenTrie match);
+
+        void Inject(Stream staged, int sequenceNumberEffect);
```
seancpeters: I believe this is what is going on:
Using Inject() is logically similar to seeking backwards in a stream - but we use forward only streams. 

The specific use in this PR is:
- The InlineMarkupConditional read some data to process.
- After examining the data, InlineMarkupConditional decides it shouldn't be processing that data.
- Prior to this PR, the unprocessed data would be written to the output stream, and processing of the input data would continue after that data.
- With this PR, the unprocessed source data is "injected" back into the source stream, so it can be read again, and processed by other operations, if applicable.

mlorbetske:

PR #1060, opened 2017-07-11
Make B2C issuer update logic more robust

danroth27:
In the future B2C will support allowing the user to control the domain in the issuer, so we need to make the issuer update logic more robust.


danroth27:
@mlorbetske Should be ready to go

PR #1062, opened 2017-07-11
Update to SetupCrossgen 289 - Coherence 26210

mlorbetske:
Moves to 2.0.0 branded packages

### DIFF
--- a/tools/DependencyUpdater/Program.cs
+++ b/tools/DependencyUpdater/Program.cs
@@ -34,7 +34,7 @@ dotnet DependencyUpdater.dll [path to templates:string] [infer no-timestamp pack
 
                     foreach (string line in sem.Split('\n'))
                     {
-                    	string trimmed = line.Trim();
+                        string trimmed = line.Trim();
                         string share = trimmed.Substring(trimmed.LastIndexOf(':') + 1).Trim();
                         string source = Path.Combine(share, @"Signed\Packages") + (NoTimestamp ? "-NoTimeStamp" : "");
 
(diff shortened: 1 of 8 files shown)


tools/DependencyUpdater/Program.cs
```diff
@@ -31,8 +31,17 @@ static int Main(string[] args)
                 else if (File.Exists(arg))
                 {
                     string sem = File.ReadAllText(arg);
-                    string share = sem.Substring(sem.LastIndexOf(':') + 1).Trim();
-                    sources.Add(Path.Combine(share, @"Signed\Packages") + (NoTimestamp ? "-NoTimeStamp" : ""));
+
+                    foreach (string line in sem.Split('\n'))
+                    {
+                        string share = line.Trim().Substring(line.LastIndexOf(':') + 1).Trim();
```
seancpeters: I think you need to get rid of the first Trim(), otherwise the substring index could be off.
`line.LastIndexOf(':')`
operates on the untrimmed version of line. But the substring operates on 
`line.Trim()`

One case which will not be correctly processed is a line with leading space, e.g ` blah :foo ` (with a leading space before blah).

tools/DependencyUpdater/Program.cs
```diff
@@ -34,7 +34,8 @@ static int Main(string[] args)
 
                     foreach (string line in sem.Split('\n'))
                     {
-                        string share = line.Trim().Substring(line.LastIndexOf(':') + 1).Trim();
+                    	string trimmed = line.Trim();
```
seancpeters: Nit: spacing (could be a GitHub display issue)

PR #1065, opened 2017-07-11
Use consistent modern casing across all page titles

danroth27:


danroth27:
@mlorbetske Ready to go

PR #1073, opened 2017-07-12
Fixup comment in EmailSender

danroth27:


danroth27:

PR #1074, opened 2017-07-12
Remove unused rememberMe parameter

danroth27:


danroth27:

PR #1070, opened 2017-07-12
Numerous fixes for Razor Pages project template

DamianEdwards:
Many fixes to the Razor Pages project template, both functional (fixing bugs) and aesthetic (changing style).

PR #1078, opened 2017-07-12
Confirm the user's already saved email

danroth27:
@HaoK Please review

### DIFF
--- a/template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/RazorPagesWeb-CSharp/Pages/Account/Manage/Index.cshtml.cs
+++ b/template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/RazorPagesWeb-CSharp/Pages/Account/Manage/Index.cshtml.cs
@@ -53,18 +53,19 @@ namespace Company.WebApplication1.Pages.Account.Manage
 
         public async Task<IActionResult> OnGetAsync()
         {
-            var user = await _userManager.GetUserAsync(HttpContext.User);
+            var user = await _userManager.GetUserAsync(User);
             if (user == null)
             {
                 return RedirectToPage("/Error");
             }
 
-            Username = user.UserName;
+            Username = await _userManager.GetUserNameAsync(user);
 
             Input = new InputModel()
             {
-                Email = user.Email,
-                PhoneNumber = user.PhoneNumber
+                Email = await _userManager.GetEmailAsync(user);
+                PhoneNumber = 
+                GetPhoneNumberAsync(user);
             };
 
             IsEmailConfirmed = await _userManager.IsEmailConfirmedAsync(user);
@@ -79,13 +80,14 @@ namespace Company.WebApplication1.Pages.Account.Manage
                 return Page();
             }
 
-            var user = await _userManager.GetUserAsync(HttpContext.User);
+            var user = await _userManager.GetUserAsync(User);
             if (user == null)
             {
                 return RedirectToPage("/Error");
             }
 
-            if (Input.Email != user.Email)
+            var email = await _userManager.GetEmailAsync(user);
+            if (Input.Email != email)
             {
                 var setEmailResult = await _userManager.SetEmailAsync(user, Input.Email);
                 if (!setEmailResult.Succeeded)
@@ -94,7 +96,8 @@ namespace Company.WebApplication1.Pages.Account.Manage
                 }
             }
 
-            if (Input.PhoneNumber != user.PhoneNumber)
+            var phoneNumber = await _userManager.GetPhoneNumber(user);
+            if (Input.PhoneNumber != phoneNumber)
             {
                 var setPhoneResult = await _userManager.SetPhoneNumberAsync(user, Input.PhoneNumber);
                 if (!setPhoneResult.Succeeded)
@@ -113,15 +116,17 @@ namespace Company.WebApplication1.Pages.Account.Manage
                 return Page();
             }
 
-            var user = await _userManager.GetUserAsync(HttpContext.User);
+            var user = await _userManager.GetUserAsync(User);
             if (user == null)
             {
                 return RedirectToPage("/Error");
             }
 
             var code = await _userManager.GenerateEmailConfirmationTokenAsync(user);
-            var callbackUrl = Url.EmailConfirmationLink(user.Id, code, Request.Scheme);
-            await _emailSender.SendEmailConfirmationAsync(user.Email, callbackUrl);
+            var userId = await _userManager.GetUserIdAsync(user);
+            var callbackUrl = Url.EmailConfirmationLink(userId, code, Request.Scheme);
+            var email = await _userManager.GetEmailAsync(user);
+            await _emailSender.SendEmailConfirmationAsync(email, callbackUrl);
 
             StatusMessage = "Verification email sent. Please check your email.";
             return RedirectToPage();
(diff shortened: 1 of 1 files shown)


danroth27:
@phenning

template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/RazorPagesWeb-CSharp/Pages/Account/Manage/Index.cshtml.cs
```diff
@@ -121,7 +121,7 @@ public class InputModel
 
             var code = await _userManager.GenerateEmailConfirmationTokenAsync(user);
             var callbackUrl = Url.EmailConfirmationLink(user.Id, code, Request.Scheme);
-            await _emailSender.SendEmailConfirmationAsync(Input.Email, callbackUrl);
+            await _emailSender.SendEmailConfirmationAsync(user.Email, callbackUrl);
```
HaoK: Typically we don't access user's properties like this directly, instead to get the user's email, you go through identity: `await _userManager.GetEmailAsync(user)`

HaoK: Same issue exists for user.Id in the line above, that should probably be `await _userManager.GetUserIdAsync(user)`

danroth27: @DamianEdwards FYI - recommendation is to not access properties on ApplicationUser directly, but to go through UserManager instead.

danroth27:
@HaoK @DamianEdwards @phenning Updated - please review.

template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/RazorPagesWeb-CSharp/Pages/Account/Manage/Index.cshtml.cs
```diff
@@ -113,15 +116,17 @@ public class InputModel
...
+            var user = await _userManager.GetUserAsync(User);
             if (user == null)
             {
                 return RedirectToPage("/Error");
             }
 
             var code = await _userManager.GenerateEmailConfirmationTokenAsync(user);
-            var callbackUrl = Url.EmailConfirmationLink(user.Id, code, Request.Scheme);
-            await _emailSender.SendEmailConfirmationAsync(Input.Email, callbackUrl);
+            var userId = await _userManager.GetUserIdAsync(user);
```
danroth27: @HaoK Would it be better to use the user manager to pull userId and username out of the principal? I.e. `_userManager.GetUserName(User)` and `_userManager.GetUserId(User)`

HaoK: Nah those are put in the principal on purpose as claims (NameIdentifier/Name by default).

template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/RazorPagesWeb-CSharp/Pages/Account/Manage/Index.cshtml.cs
```diff
@@ -53,18 +53,19 @@ public class InputModel
...
-            Username = user.UserName;
+            Username = await _userManager.GetUserNameAsync(user);
 
             Input = new InputModel()
             {
-                Email = user.Email,
-                PhoneNumber = user.PhoneNumber
+                Email = await _userManager.GetEmailAsync(user);
+                PhoneNumber = 
+                GetPhoneNumberAsync(user);
```
HaoK: should this be _userManager.GetPhoneNumberAsync?

template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/RazorPagesWeb-CSharp/Pages/Account/Manage/Index.cshtml.cs
```diff
@@ -53,18 +53,19 @@ public class InputModel
...
         {
-            var user = await _userManager.GetUserAsync(HttpContext.User);
+            var user = await _userManager.GetUserAsync(User);
             if (user == null)
             {
                 return RedirectToPage("/Error");
             }
 
-            Username = user.UserName;
+            Username = await _userManager.GetUserNameAsync(user);
```
HaoK: Why do we need username on the model?  This should already be in the ClaimsPrincipal (User.Identity.Name) for the manage page, we aren't allowing this to be changed are we?

template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/RazorPagesWeb-CSharp/Pages/Account/Manage/Index.cshtml.cs
HaoK: That said, the principal claims are configurable via: IdentityOptions.ClaimsIdentity, so maybe its a bit safer to go through the manager.

template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/RazorPagesWeb-CSharp/Pages/Account/Manage/Index.cshtml.cs
danroth27: This page should reflect what is stored in the database. The username is not editable. It's just displayed as part of the user profile.

template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/RazorPagesWeb-CSharp/Pages/Account/Manage/Index.cshtml.cs
danroth27: I assume the user ID never changes, so I think calling `_userManager.GetUserId(User)` should be fine. For the username, I think it really depends if you are ok with username stored in the cookie, or if you want the most up to date username from the identity store. @HaoK Does that sound right?

HaoK: Yeah its not a big deal either way, you'd assume any app that let them change the username, would also clear the cookie/sign them in again as part of the change name flow.  That's what our manage actions do already...

template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/RazorPagesWeb-CSharp/Pages/Account/Manage/Index.cshtml.cs
HaoK: Yeah so if its read only data on the ManagePage, you probably could just use User.Identity.Name in the view in this case.

HaoK: Yeah that's what we do already for the "Hello, DanRoth" when you are signed in:

That view does: Hello, @User.Identity.Name!

No reason not to do the same

danroth27:
@DamianEdwards is handling these changes as part of https://github.com/dotnet/templating/pull/1070

PR #1087, opened 2017-07-14
https ports are configured to be in the proper range (44300-44399)

seancpeters:


template_feed/Microsoft.DotNet.Web.ProjectTemplates.1.x/content/StarterWeb-CSharp/.template.config/template.json
```diff
@@ -157,7 +157,11 @@
     },
     "HttpsPortGenerated": {
       "type": "generated",
-      "generator": "port"
+      "generator": "port",
+      "parameters": {
+        "low": 44300,
+        "high": 44400
```
phenning: "high" being set to 44000 here would indicate to a user that I may get a port numbered 44000, which won't be the case.

We know that the low is inclusive and the high is exclusive in the implementation, but I don't think we should flow those semantics out to the configuration. 

It seems more intuitive if both low and high in the configuration were inclusive.

phenning: I should have mentioned this in the other PR, but it didn't sink in until I saw the configuration in the json.

PR #1103, opened 2017-07-18
Fix typo

danroth27:


danroth27:
@phenning Should be ready to go

PR #1112, opened 2017-07-19
Remove enable 2fa button

danroth27:
@HaoK @DamianEdwards @blowdart Please review

### DIFF
--- a/template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/RazorPagesWeb-CSharp/Pages/Account/Manage/TwoFactorAuthentication.cshtml
+++ b/template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/RazorPagesWeb-CSharp/Pages/Account/Manage/TwoFactorAuthentication.cshtml
@@ -5,7 +5,7 @@
 }
 
 <h4>@ViewData["Title"]</h4>
-@if (Model.HasAny2faProviders && Model.Is2faEnabled)
+@if (Model.Is2faEnabled)
 {
     if (Model.RecoveryCodesLeft == 0)
     {
--- a/template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/RazorPagesWeb-CSharp/Pages/Account/Manage/TwoFactorAuthentication.cshtml
+++ b/template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/RazorPagesWeb-CSharp/Pages/Account/Manage/TwoFactorAuthentication.cshtml
@@ -5,44 +5,32 @@
 }
 
 <h4>@ViewData["Title"]</h4>
-@if (Model.HasAny2faProviders)
+@if (Model.HasAny2faProviders && Model.Is2faEnabled)
 {
-    if (Model.Is2faEnabled)
+    if (Model.RecoveryCodesLeft == 0)
     {
-        if (Model.RecoveryCodesLeft == 0)
-        {
-            <div class="alert alert-danger">
-                <strong>You have no recovery codes left.</strong>
-                <p>You must <a asp-page="./GenerateRecoveryCodes">generate a new set of recovery codes</a> before you can log in with a recovery code.</p>
-            </div>
-        }
-        else if (Model.RecoveryCodesLeft == 1)
-        {
-            <div class="alert alert-danger">
-                <strong>You have 1 recovery code left.</strong>
-                <p>You can <a asp-page="./GenerateRecoveryCodes">generate a new set of recovery codes</a>.</p>
-            </div>
-        }
-        else if (Model.RecoveryCodesLeft <= 3)
-        {
-            <div class="alert alert-warning">
-                <strong>You have @Model.RecoveryCodesLeft recovery codes left.</strong>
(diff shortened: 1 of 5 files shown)


DamianEdwards:
So is this basically removing the ability to manually enable 2FA, and rather enabling it automatically **only** when an authenticator app is verified?

danroth27:
Yes

HaoK:
Yeah I can't really review this without seeing the new flow... I think this means that the only way to enableTFA now is via the enable authenticator post?

danroth27:
@HaoK Yup

template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/RazorPagesWeb-CSharp/Pages/Account/Manage/TwoFactorAuthentication.cshtml
```diff
@@ -5,44 +5,32 @@
 }
 
 <h4>@ViewData["Title"]</h4>
-@if (Model.HasAny2faProviders)
+@if (Model.HasAny2faProviders && Model.Is2faEnabled)
```
HaoK: Should we just rename things to be clearer, 2fa => Authenticator?  And then this could just be a single Model.IsAuthenticatorEnabled check and we can do away with HasAny2faProviders entirely

HaoK: Basically move away from the old attempt to make it easy to add arbitrary 2fa providers, and just tailor the template for only the authenticator flow.  They could still add more 2fa providers and additional flows, they would just need to do that all themselves.

danroth27: My preference is to leave the generic 2FA stuff wherever possible. I'm guessing there will be folks that want to add back the SMS or email based 2FA support

HaoK: Yeah but I think we are in an in between state now, where the 2fa flow basically is very targetted towards the authenticator.  They will likely still need to do roughly what we did to enable authenticator to make SMS/email work again (unlike previous templates which had more of the SMS/email stuff baked in to attempt to make it more straightforward to drop in additional TFA providers).  I thought we eliminated most of the code/comments that were there to try and make it easy to add back other 2FA providers?

HaoK: Plus I thought the long term direction was to try and guide everyone to use only the authenticator flow as SMS/email are less secure right @blowdart ?

DamianEdwards: I tend to agree with @HaoK on this.

blowdart: Yes but people may want fallbacks. Or it's simply not suitable for their audience. Take the US social security web site, it's 2fa is SMS based, because really, you can't assume every user is going to have a smartphone.

HaoK: Yeah so maybe we have the template only support authenticator, and then have a longer doc/sample/walkthrough that shows how to add SMS support

HaoK: That will be even more complicated since it will bring back the whole EnableTFA button/issue again

danroth27: I also think renaming things to make them sound authenticator specific is misleading. There is no notion of an authenticator app being "enabled" in the runtime. Currently the names directly reflect the runtime state, which is clear and correct.

We could however just remove the HasAny2faProviders property. I wasn't sure if it was needed anymore. @HaoK is checking `Is2faEnabled` sufficient in this case?

HaoK: Yeah for this app, Is2faEnabled == AuthenticatorEnabled so that probably good enough

mlorbetske:
Is this ready to go?

danroth27:
@HaoK I updated as discussed. Approve?

mlorbetske:

HaoK:
Yep looks good

PR #1113, opened 2017-07-19
Remove default controller route value in razor pages

danroth27:
@javiercn Please review


mlorbetske:
Should any templates apart from Razor pages get this change as well?

danroth27:
@mlorbetske No, this change is specific to Razor Pages

PR #1134, opened 2017-07-21
Fixes the bugs discovered during the verification

### LINKED ISSUE #1115: Windows auth does not work on web app template when running standalone
Create a web app from the template and select Windows auth. When running behind IIS Express, you can see your Windows identity in the top-right corner of the page. It is not there, however, when running standalone.

Consider fixing the template or removing the launch setting from `launchSettings.json`.

### LINKED ISSUE #1124: (dotnet new) `-au Windows` should imply --use-launch-settings
Currently `dotnet new mvc -au Windows` does not add launchsettings.json to the project. As a result when starting the application (e.g. from VS) the user is not being authenticated even if Windows auth was supposed to be turned on (VS will add its own launchsettings.json with `windowsAuthentication` set to `false`). 

Adding web.config with forwardWindowsAuthToken=true should also be considered - launchsettings.json is only being used by VS (dotnet new does not count since Kestrel does not support Windows auth), so when the application is run in production it has to use ANCM/IIS and without web.config the application will not run at all. If web.config exists but does not have forwardWindowsAuthToken=true Windows auth will not work. 

Right now -au Windows create an app that will not authenticate users at all and is a pit of failure given how much the user needs to know to make it work properly.

### LINKED ISSUE #1125: --aad-b2c-instance does not seem to work
Even though I used `--aad-b2c-instance https://login1.microsoftonline.com/tfp1/` the Instance in appsetting.json is `https://login.microsoftonline.com/tfp`. 

```
PS C:\temp\test> dotnet new mvc -au IndividualB2C --aad-b2c-instance https://login1.microsoftonline.com/tfp1/ --no-restore
The template "ASP.NET Core Web App (Model-View-Controller)" was created successfully.
This template contains technologies from parties other than Microsoft, see https://aka.ms/template-3pn for details.
PS C:\temp\test> cat .\appsettings.json
{
  "AzureAdB2C": {
    "Instance": "https://login.microsoftonline.com/tfp",
    "ClientId": "11111111-1111-1111-11111111111111111",
    "CallbackPath": "/signin-oidc",
    "Domain": "qualified.domain.name",
    "SignUpSignInPolicyId": "",
    "ResetPasswordPolicyId": "",
    "EditProfilePolicyId": ""
  },
  "Logging": {
    "IncludeScopes": false,
    "LogLevel": {
      "Default": "Warning"
    }
  }
}
```

Another way to check is to compare differences from running
`dotnet new mvc -au IndividualB2C`
and
`dotnet new mvc -au IndividualB2C --aad-b2c-instance https://login1.microsoftonline.com/tfp1/`

You will see that the only difference is UserSecretsId

### LINKED ISSUE #1126: Incorrect description for --client-id switch
The description for the --client-id switch says:
```
  --client-id                     The Client ID for this project (use with individual and organizational auth types).
                                  string - Optional
                                  Default: 11111111-1111-1111-11111111111111111
```

I believe it should be `... with IndividualB2C and organizational...` and not `... with individual and organizational ...`. The Individual auth type (as opposed to IndividualB2C) is Identity based and I don't think it uses the client id in any way.

### LINKED ISSUE #1127: Typo in --tenant-id description
The description of --tenant-id switch reads:

```
 --tenant-id                     The TenantId ID of the directory to connect to (use with Single-org auth types).
                                 string - Optional
                                 Default: 22222222-2222-2222-2222-222222222222
```

It should read:
`(use with Single-org auth type)` as opposed to `(use with Single-org auth types)` - (plural types).

### LINKED ISSUE #1128: Inconsistency with SingleOrg and MultiOrg spelling
Some switches (e.g. `-r`) use `SingleOrg` while other (e.g. `--callback-path`) use `Single-org`. I believe the correct spelling is `SingleOrg` because it is the name used to choose the type of authentication.

Same with MultiOrg

### LINKED ISSUE #1132: `--audience` switch has no effect on generated project
There is no difference in output between:

`dotnet new webapi -au SingleOrg --no-restore --audience AUDIENCE --force`
and
`dotnet new webapi -au SingleOrg --no-restore`

(less than UserSecretsId)

mlorbetske:
Fixes #1132

Fixes #1128

Fixes #1127

Fixes #1126

Fixes #1125

Fixes #1124

Fixes #1115

### DIFF
--- a/template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/StarterWeb-CSharp/.template.config/template.json
+++ b/template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/StarterWeb-CSharp/.template.config/template.json
@@ -61,7 +61,7 @@
           ]
         },
         {
-          "condition": "(!IncludeLaunchSettings)",
+          "condition": "(!IncludeLaunchSettings && !WindowsAuth)",
           "exclude": [
             "Properties/launchSettings.json"
           ]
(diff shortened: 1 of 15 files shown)


template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/StarterWeb-CSharp/.template.config/template.json
```diff
@@ -121,66 +121,66 @@
...
       "replaces": "https:////login.microsoftonline.com/",
-      "description": "The Azure Active Directory instance to connect to (use with Single-org or Multi-org auth types)."
+      "description": "The Azure Active Directory instance to connect to (use with SingleOrg or MultiOrg auth)."
     },
     "ClientId": {
       "type": "parameter",
       "datatype": "string",
       "replaces": "11111111-1111-1111-11111111111111111",
-      "description": "The Client ID for this project (use with individual and organizational auth types)."
+      "description": "The Client ID for this project (use with IndividualB2C, SingleOrg or MultiOrg auth)."
```
seancpeters: Nit: I think we should use the Oxford comma here (https://en.wikipedia.org/wiki/Serial_comma), it's less ambiguous, e.g.:
`use with IndividualB2C, SingleOrg, or MultiOrg auth`
...adding a comma after "SingleOrg"
I don't know if our doc standards say anything about using or not using it.

PR #1138, opened 2017-07-21
Set tailcalls to true in project file

rodrigovidal:


mlorbetske:
@enricosada @KevinRansom any thoughts on this change?

mlorbetske:
@enricosada @KevinRansom friendly ping

KevinRansom:
Hi @rodrigovidal,

This is already taken care of here:  the latest builds of the dotnet/cli and netsdk should have this change.

https://github.com/Microsoft/visualfsharp/blob/master/src/fsharp/FSharp.Build/Microsoft.FSharp.NetSdk.props#L40

This file deploys with compiler, because it sets values that are generally applicable.

@mlorbetske 
This PR should be closed.

Kevin

mlorbetske:
Thanks!

PR #1144, opened 2017-07-24
Update helpers for individual auth

javiercn:
Makes helpers only apply to individual local auth.
Add identity only on the local case.


javiercn:
@mlorbetske

PR #1166, opened 2017-07-28
util to remove bin/ obj/ and .vs/ directories - for template output comparison

seancpeters:


tools/ComparisonCleanup/Program.cs
```diff
@@ -0,0 +1,71 @@
...
+            if (args.Length == 0)
+            {
+                UsageMessage();
+                return;
+            }
+
+            string baseDir = args[0];
+            if (!Directory.Exists(baseDir))
+            {
+                Console.WriteLine($"directory '{baseDir}' doesn't exist");
```
mlorbetske: We should return here if the directory doesn't exist I think

seancpeters: ah yes, definitely.

PR #1175, opened 2017-08-02
Support for derived symbols

seancpeters:
Derived symbols work like parameter symbols in most regards. The difference is that value of a derived symbol is based on the value of an existing parameter symbol, transformed by a value form.


seancpeters:
I'm going to split these changes into 2 PR's: one for the refactoring, and one for the functional changes. Closing this.

PR #1190, opened 2017-08-08
Add LESS, SCSS and TypeScript item templates

mlorbetske:


mlorbetske:
Since there won't be variations on say the LESS template (as LESS is the language, which is part of the name), adding the language pivot to allow selection of other languages for the target template wouldn't make sense. Updated the classifications & tags as appropriate.

PR #1194, opened 2017-08-08
DerivedSymbol implementation

seancpeters:
symbol whose value is a value-form transform of another symbol value

### DIFF
--- a/src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/Macros/ProcessValueFormMacro.cs
+++ b/src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/Macros/ProcessValueFormMacro.cs
@@ -43,11 +43,21 @@ namespace Microsoft.TemplateEngine.Orchestrator.RunnableProjects.Macros
             {
                 value = form.Process(realConfig.Forms, value);
 
-                Parameter p = new Parameter
+                Parameter p;
+
+                if (parameters.TryGetParameterDefinition(config.VariableName, out ITemplateParameter existingParam))
+                {   // Derived parameters already have a definition, but need value form processing.
+                    // When the param already exists, use its definition.
+                    p = (Parameter)existingParam;
+                }
+                else
                 {
-                    IsVariable = true,
-                    Name = config.VariableName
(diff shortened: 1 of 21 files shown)


src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/Macros/ProcessValueFormMacro.cs
```diff
@@ -43,11 +43,21 @@ public void EvaluateConfig(IEngineEnvironmentSettings environmentSettings, IVari
             {
                 value = form.Process(realConfig.Forms, value);
 
-                Parameter p = new Parameter
+                Parameter p;
+
+                if (parameters.TryGetParameterDefinition(config.VariableName, out ITemplateParameter existingParam))
+                {   // Derived parameters already have a definition, but need value form processing.
+                    // When the param already exists, use its definition.
+                    p = (Parameter)existingParam;
```
mlorbetske: Let's set `IsVariable = true` on the existing parameter for consistency (even though it's never used right now) and make all of the other macros behave this way too.

We should probably make this comment a little more clear as well (since we're in the value forms macro) that the existing parameter must be reused to allow the processing of value forms declared on the derived symbol type later in the thing that deals with value forms for parameter type symbols.

We should also make sure that the processing of the value forms macro for the derived symbol happens before value forms for parameters/derived symbols happen

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/SimpleConfigModel.cs
```diff
@@ -571,6 +577,14 @@ private IGlobalRunConfig ProduceOperationSetup(SpecialOperationConfigParams defa
                             }
                         }
                     }
+
+                    if (symbol.Value is DerivedSymbol derivedSymbol)
```
mlorbetske: This one should move to above the replaces check (to set up the map between the value of the source symbol and the value of the derived symbol), then the block for handling the value forms of parameter symbols should be made to handle BaseSomethingSymbol

PR #1199, opened 2017-08-10
[WIP] Refactored template resolution out of New3Command

seancpeters:


src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -525,25 +375,26 @@ private IReadOnlyList<ITemplateInfo> TemplatesToDisplayInfoAbout
...
                 IEnumerable<ITemplateInfo> templateList;
 
-                if (UnambiguousTemplateGroupToUse != null && UnambiguousTemplateGroupToUse.Count > 0)
+                if (_templateResolutionResult.UnambiguousTemplateGroupToUse != null && _templateResolutionResult.UnambiguousTemplateGroupToUse.Count > 0)
                 {
-                    templateList = UnambiguousTemplateGroupToUse.Select(x => x.Info);
+                    templateList = _templateResolutionResult.UnambiguousTemplateGroupToUse.Select(x => x.Info);
                 }
-                else if (!string.IsNullOrEmpty(TemplateName) && _matchedTemplatesWithSecondaryMatchInfo != null && _matchedTemplatesWithSecondaryMatchInfo.Count > 0)
+                else if (!string.IsNullOrEmpty(TemplateName) && _templateResolutionResult.MatchedTemplatesWithSecondaryMatchInfo != null && _templateResolutionResult.MatchedTemplatesWithSecondaryMatchInfo.Count > 0)
```
mlorbetske: Should we expose a property on the resolution result that is the result of the last two parts of the expression? Something like HasMatchedTemplatesWithSecondaryMatchInfo?

src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -525,25 +375,26 @@ private IReadOnlyList<ITemplateInfo> TemplatesToDisplayInfoAbout
...
                 {   // without template name, it's not reasonable to do secondary matching
-                    templateList = _matchedTemplatesWithSecondaryMatchInfo.Select(x => x.Info);
+                    templateList = _templateResolutionResult.MatchedTemplatesWithSecondaryMatchInfo.Select(x => x.Info);
                 }
-                else if (_matchedTemplates != null && _matchedTemplates.Any(x => x.IsMatch))
+                else if (_templateResolutionResult.CoreMatchedTemplates != null && _templateResolutionResult.CoreMatchedTemplates.Any(x => x.IsMatch))
```
mlorbetske: Same sort of "Has" bool suggestion

src/Microsoft.TemplateEngine.Cli/New3Command.cs
seancpeters: My first thought on this was "yes, absolutely!". But we sometimes store "matches" in the result lists that are only partial matches, or even have some mismatch info, e.g. invalid parameters or invalid parameter values. So I'm a bit wary that the "Has" properties could be misleading... then again the lists can be misleading too. I want to think about it a bit more, but the current situation is definitely a bit kludgy.

src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -61,8 +59,10 @@ public New3Command(string commandName, ITemplateEngineHost host, ITelemetryLogge
             CommandName = commandName;
             _paths = new Paths(EnvironmentSettings);
             _onFirstRun = onFirstRun;
-
+            _hostDataLoader = new HostSpecificDataLoader(EnvironmentSettings.SettingsLoader);
             _commandInput = commandInput;
+
+            EnvironmentSettings.Host.TryGetHostParamDefault("prefs:language", out _defaultLanguage);
```
mlorbetske: Explicitly set _defaultLanguage to an appropriate value (null) when TryGetHostParamDefault returns false

mlorbetske: We can also reference host directly here

src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -17,31 +17,29 @@
...
     {
-        private static readonly IReadOnlyCollection<MatchLocation> NameFields = new HashSet<MatchLocation> { MatchLocation.Name, MatchLocation.ShortName, MatchLocation.Alias };
         private HostSpecificTemplateData _hostSpecificTemplateData;
-        private IReadOnlyList<IFilteredTemplateInfo> _matchedTemplates;
         private readonly ITelemetryLogger _telemetryLogger;
         private readonly TemplateCreator _templateCreator;
         private readonly SettingsLoader _settingsLoader;
         private readonly AliasRegistry _aliasRegistry;
         private readonly Paths _paths;
         private readonly ExtendedTemplateEngineHost _host;
```
mlorbetske: Looks like this is no longer used

src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -61,8 +59,10 @@ public New3Command(string commandName, ITemplateEngineHost host, ITelemetryLogge
             CommandName = commandName;
```
mlorbetske: Looks like the `host = _host = new ExtendedTemplateEngineHost(host, this);` no longer requires the `_host =` portion either

src/Microsoft.TemplateEngine.Edge/Settings/TemplateCache.cs
```diff
@@ -47,38 +47,6 @@ public TemplateCache(IEngineEnvironmentSettings environmentSettings, JObject par
         [JsonProperty]
         public IReadOnlyList<TemplateInfo> TemplateInfo { get; set; }
 
-        public IReadOnlyCollection<IFilteredTemplateInfo> List(bool exactMatchesOnly, params Func<ITemplateInfo, MatchInfo?>[] filters)
```
mlorbetske: Let's preserve this signature, call the new location of this method & mark it as obsolete (maybe?)

src/Microsoft.TemplateEngine.Cli/New3Command.cs
seancpeters: This turned out to be not so bad. I added the commonly used patterns as properties on the result class.

PR #1212, opened 2017-08-16
Update template for latest release, fixing #1211

### LINKED ISSUE #1211: dotnet new mvc --auth Individual CreateIdentitySchema class breaks Sqlite migrations
Looks like the generated CreateIdentitySchema class needs to be updated to match the current models used for identity so as to not create a migration requiring an `AddForeignKey` operation which isn't supported by Sqlite.

A simple way to repro the problem:
```
$ dotnet --version
2.0.0
$ dotnet new mvc --auth Individual
$ dotnet ef migrations add InitialMigration
$ dotnet ef database update
```
Produces the error:
```
Applying migration '20170815210528_InitialMigration'.
System.NotSupportedException: SQLite does not support this migration operation ('AddForeignKeyOperation'). For more information, see http://go.microsoft.com/fwlink/?LinkId=723262.
   at Microsoft.EntityFrameworkCore.Migrations.SqliteMigrationsSqlGenerator.Generate(AddForeignKeyOperation operation, IModel model, MigrationCommandListBuilder builder)
   at Microsoft.EntityFrameworkCore.Migrations.MigrationsSqlGenerator.<>c.<.cctor>b__63_1(MigrationsSqlGenerator g, MigrationOperation o, IModel m, MigrationCommandListBuilder b)
   at Microsoft.EntityFrameworkCore.Migrations.MigrationsSqlGenerator.Generate(MigrationOperation operation, IModel model, MigrationCommandListBuilder builder)
   at Microsoft.EntityFrameworkCore.Migrations.MigrationsSqlGenerator.Generate(IReadOnlyList`1 operations, IModel model)
   at Microsoft.EntityFrameworkCore.Migrations.SqliteMigrationsSqlGenerator.Generate(IReadOnlyList`1 operations, IModel model)
   at Microsoft.EntityFrameworkCore.Migrations.Internal.Migrator.GenerateUpSql(Migration migration)
   at Microsoft.EntityFrameworkCore.Migrations.Internal.Migrator.<>c__DisplayClass13_2.<GetMigrationCommandLists>b__2()
   at Microsoft.EntityFrameworkCore.Migrations.Internal.Migrator.Migrate(String targetMigration)
   at Microsoft.EntityFrameworkCore.Design.Internal.MigrationsOperations.UpdateDatabase(String targetMigration, String contextType)
   at Microsoft.EntityFrameworkCore.Design.OperationExecutor.UpdateDatabase.<>c__DisplayClass0_1.<.ctor>b__0()
   at Microsoft.EntityFrameworkCore.Design.OperationExecutor.OperationBase.Execute(Action action)
SQLite does not support this migration operation ('AddForeignKeyOperation'). For more information, see http://go.microsoft.com/fwlink/?LinkId=723262.
```

The `dotnet db database update` will fail because an `AddForeignKey` call is required to get the schema up to date and the default Sqlite provider doesn't support that operation.

A workaround is to create the migration, then copy the contents of the newly created *.Designer.cs file to the *CreateIdentitySchema.Designer.cs file, and manually resolve the differences in the new db migration *.cs file with the old *CreateIdentitySchemaDesigner.cs, effectively folding the migration up into the initial schema creation. Then delete the db file so a new one is created.

baumatron:
Here's my attempt to fix #1211.

I basically did this:
```
$ dotnet new mvc --auth Individual
$ dotnet ef migrations add InitialMigration
```
Then incorporated the changes from the generated InitialMigration into the initial `00000000000000_CreateIdentitySchema.*cs` files in the templates here.

I'm not familiar with this codebase or how templates are constructed, so someone who knows this area well should really have a good look at this and make sure it works for other scenarios.

### DIFF
--- a/template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/StarterWeb-CSharp/Data/Migrations/00000000000000_CreateIdentitySchema.Designer.cs
+++ b/template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/StarterWeb-CSharp/Data/Migrations/00000000000000_CreateIdentitySchema.Designer.cs
@@ -17,17 +17,85 @@ namespace Company.WebApplication1.Data.Migrations
     {
         protected override void BuildTargetModel(ModelBuilder modelBuilder)
         {
+#pragma warning disable 612, 618
             modelBuilder
 #if (UseLocalDB)
-                .HasAnnotation("ProductVersion", "1.0.0-rc3")
+                .HasAnnotation("ProductVersion", "2.0.0-rtm-26452")
                 .HasAnnotation("SqlServer:ValueGenerationStrategy", SqlServerValueGenerationStrategy.IdentityColumn);
 #else
-                .HasAnnotation("ProductVersion", "1.0.2");
+                .HasAnnotation("ProductVersion", "2.0.0-rtm-26452");
 #endif
+
+            modelBuilder.Entity("Company.WebApplication1.Models.ApplicationUser", b =>
+                {
+                    b.Property<string>("Id")
+                        .ValueGeneratedOnAdd();
+
+                    b.Property<int>("AccessFailedCount");
+
+                    b.Property<string>("ConcurrencyStamp")
+                        .IsConcurrencyToken();
+
+                    b.Property<string>("Email")
+#if (UseLocalDB)
+                        .HasAnnotation("MaxLength", 256);
+#else
+                        .HasMaxLength(256);
+#endif
+
+                    b.Property<bool>("EmailConfirmed");
+
+                    b.Property<bool>("LockoutEnabled");
+
+                    b.Property<DateTimeOffset?>("LockoutEnd");
+
+                    b.Property<string>("NormalizedEmail")
+#if (UseLocalDB)
+                        .HasAnnotation("MaxLength", 256);
+#else
+                        .HasMaxLength(256);
+#endif
+
+                    b.Property<string>("NormalizedUserName")
+#if (UseLocalDB)
+                        .HasAnnotation("MaxLength", 256);
+#else
+                        .HasMaxLength(256);
+#endif
+
+                    b.Property<string>("PasswordHash");
+
+                    b.Property<string>("PhoneNumber");
+
+                    b.Property<bool>("PhoneNumberConfirmed");
+
+                    b.Property<string>("SecurityStamp");
+
+                    b.Property<bool>("TwoFactorEnabled");
+
+                    b.Property<string>("UserName")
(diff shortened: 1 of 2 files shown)


mlorbetske:
Thanks @baumatron! We'll review

danroth27:
@bricelam @HaoK Can you guys take a look at this?

template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/StarterWeb-CSharp/Data/Migrations/00000000000000_CreateIdentitySchema.Designer.cs
```diff
@@ -17,17 +17,85 @@ partial class CreateIdentitySchema
...
+
+                    b.Property<string>("ConcurrencyStamp")
+                        .IsConcurrencyToken();
+
+                    b.Property<string>("Email")
+#if (UseLocalDB)
+                        .HasAnnotation("MaxLength", 256);
+#else
+                        .HasMaxLength(256);
+#endif
```
bricelam: Weird. Are they really generated differently between the two? AFAIK those two lines are equivalent.

ajcvickers:
See also the discussion here: https://github.com/aspnet/Identity/issues/1361 /cc @blowdart

HaoK:
There weren't any schema changes but we did refactor the poco base classes and remove navigation properties. That might have caused migration changes?

ajcvickers:
@HaoK So a database created for Identity in 1.1 should still work for 2.0 without needing to be migrated?

HaoK:
Yes @ajcvickers the database shouldn't need to be migrated at all

ajcvickers:
@HaoK Okay, thanks.

HaoK:
But it appears the initial migration is now different even though there were no functional schema changes?  Removing the navigation properties seems to be the root cause?

ajcvickers:
I think there may be other causes--for example, we didn't explicitly make some FKs in 1-1 relationships unique in 1.0. The updated schema here appears to better match the model due to improvements in EF, but it seems to me like the old schema should still work fine with the new model. At least, if the model has not been tweaked. If the model is tweaked in application code, then there may need to be a migration, but that seems okay too, @bricelam Any thoughts?

bricelam:
Agree, it all looks compatible. Getting a few extra operations the first time you evolve the model in your application after upgrading to 2.0.0 seems fine to me too.

mlorbetske:
@bricelam looks like this is good to go then?

bricelam:
I'll let @HaoK or @ajcvickers give the final singnoff since it's more of an Identity change.

PR #1224, opened 2017-08-25
Make the MVC content verification test explicitly pass the framework

mlorbetske:


test/Microsoft.TemplateEngine.Cli.UnitTests/PrecedenceSelectionTests.cs
```diff
@@ -5,8 +5,8 @@ namespace Microsoft.TemplateEngine.Cli.UnitTests
     public class PrecedenceSelectionTests : EndToEndTestBase
     {
         [Theory(DisplayName = nameof(VerifyTemplateContent))]
-        [InlineData("mvc", "MvcNoAuthTest.json", "MvcFramework20Test.json")]
-        [InlineData("mvc -au individual", "MvcIndAuthTest.json", "MvcFramework20Test.json")]
+        [InlineData("mvc -f netcoreapp2.0", "MvcNoAuthTest.json", "MvcFramework20Test.json")]
+        [InlineData("mvc -au individual -f netcoreapp2.0", "MvcIndAuthTest.json", "MvcFramework20Test.json")]
```
seancpeters: These test cases are meant to show, among other things, that the correct template will be selected without explicitly passing in the framework. With these changes, these first 2 tests are identical to the final 2 tests in the list.

PR #1231, opened 2017-08-30
More template resolution rework

seancpeters:
1) Greatly simplifies the template resolution code, much easier to understand.
2) IFilteredTemplateInfo now stores more information about command parse results in the context of the template it's for. This avoids the need to reparse, and also avoids having to rely on the command input object to be in the state of a particular template.
3) Invocations that would cause errors are now prevented by checking for the same errors pre-invocation.
4) Some reworking of help / error output when not invoking templates, just enough to deal with the resolution changes. There is still considerable help rework to be done, but the scope on this PR is already creeping.
5) Changes to project test runner tests - to explicitly invoke the 2.0 templates, instead of the default (which is now 2.1)

### DIFF
--- a/src/Microsoft.TemplateEngine.Cli/HelpForTemplateResolution.cs
+++ b/src/Microsoft.TemplateEngine.Cli/HelpForTemplateResolution.cs
@@ -1,0 +1,147 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.TemplateEngine.Edge.Template;
+using Microsoft.TemplateEngine.Abstractions;
+
+namespace Microsoft.TemplateEngine.Cli
+{
+    internal static class HelpForTemplateResolution
+    {
+        public static void DisplayInvalidParameters(IReadOnlyList<string> invalidParams)
+        {
+            if (invalidParams.Count > 0)
+            {
+                Reporter.Error.WriteLine(LocalizableStrings.InvalidInputSwitch.Bold().Red());
+                foreach (string flag in invalidParams)
+                {
+                    Reporter.Error.WriteLine($"  {flag}".Bold().Red());
+                }
+            }
+        }
+
+        public static void DisplayParametersInvalidForSomeTemplates(IReadOnlyList<string> invalidParams)
+        {
+            if (invalidParams.Count > 0)
+            {
+                Reporter.Error.WriteLine(LocalizableStrings.PartialTemplateMatchSwitchesNotValidForAllMatches.Bold().Red());
+                foreach (string flag in invalidParams)
+                {
+                    Reporter.Error.WriteLine($"  {flag}".Bold().Red());
+                }
+            }
+        }
+
+        // get a better name for this, it's unclear what it's doing.
+        public static bool ShowTemplateNameMismatchHelp(string templateName, string context, TemplateListResolutionResult templateResolutionResult)
+        {
+            GetContextBasedAndOtherPartialMatches(templateResolutionResult, out IReadOnlyList<IReadOnlyList<IFilteredTemplateInfo>> contextProblemMatchGroups, out IReadOnlyList<IReadOnlyList<IFilteredTemplateInfo>> remainingPartialMatchGroups);
+            return DisplayPartialNameMatchAndContextProblems(templateName, context, contextProblemMatchGroups, remainingPartialMatchGroups);
+        }
+
+        private static void GetContextBasedAndOtherPartialMatches(TemplateListResolutionResult templateResolutionResult, out IReadOnlyList<IReadOnlyList<IFilteredTemplateInfo>> contextProblemMatchGroups, out IReadOnlyList<IReadOnlyList<IFilteredTemplateInfo>> remainingPartialMatchGroups)
+        {
+            Dictionary<string, List<IFilteredTemplateInfo>> contextProblemMatches = new Dictionary<string, List<IFilteredTemplateInfo>>();
+            Dictionary<string, List<IFilteredTemplateInfo>> remainingPartialMatches = new Dictionary<string, List<IFilteredTemplateInfo>>();
+
+            // this filtering / grouping ignores language differences.
+            foreach (IFilteredTemplateInfo template in templateResolutionResult.CoreMatchedTemplates)
+            {
+                if (template.MatchDisposition.Any(x => x.Location == MatchLocation.Context && x.Kind != MatchKind.Exact))
+                {
+                    if (!contextProblemMatches.TryGetValue(template.Info.GroupIdentity, out List<IFilteredTemplateInfo> templateGroup))
+                    {
+                        templateGroup = new List<IFilteredTemplateInfo>();
+                        contextProblemMatches.Add(template.Info.GroupIdentity, templateGroup);
+                    }
+
+                    templateGroup.Add(template);
+                }
+                else if (template.MatchDisposition.Any(t => t.Location != MatchLocation.Context && t.Kind != MatchKind.Mismatch && t.Kind != MatchKind.Unspecified))
+                {
+                    if (!remainingPartialMatches.TryGetValue(template.Info.GroupIdentity, out List<IFilteredTemplateInfo> templateGroup))
+                    {
+                        templateGroup = new List<IFilteredTemplateInfo>();
+                        remainingPartialMatches.Add(template.Info.GroupIdentity, templateGroup);
+                    }
+
+                    templateGroup.Add(template);
+                }
+            }
+
+            contextProblemMatchGroups = contextProblemMatches.Values.ToList();
+            remainingPartialMatchGroups = remainingPartialMatches.Values.ToList();
+        }
+
+        private static bool DisplayPartialNameMatchAndContextProblems(string templateName, string context, IReadOnlyList<IReadOnlyList<IFilteredTemplateInfo>> contextProblemMatchGroups, IReadOnlyList<IReadOnlyList<IFilteredTemplateInfo>> remainingPartialMatchGroups)
+        {
+            bool anythingReported = false;
+            if (contextProblemMatchGroups.Count + remainingPartialMatchGroups.Count > 1)
+            {
+                Reporter.Error.WriteLine(string.Format(LocalizableStrings.AmbiguousInputTemplateName, templateName));
+                anythingReported = true;
+            }
+            else if (contextProblemMatchGroups.Count + remainingPartialMatchGroups.Count == 0)
+            {
+                Reporter.Error.WriteLine(string.Format(LocalizableStrings.NoTemplatesMatchName, templateName));
+                Reporter.Error.WriteLine();
+                return true;
+            }
+
+            // scp (2017-08-28): not certain anything in this loop can happen anymore, due to the way filtering occurs.
+            foreach (IReadOnlyList<IFilteredTemplateInfo> templateGroup in contextProblemMatchGroups)
+            {
+                // all templates in a group should have the same context & name
+                if (templateGroup[0].Info.Tags != null && templateGroup[0].Info.Tags.TryGetValue("type", out ICacheTag typeTag))
+                {
+                    MatchInfo? matchInfo = WellKnownSearchFilters.ContextFilter(context)(templateGroup[0].Info);
+                    if ((matchInfo?.Kind ?? MatchKind.Mismatch) == MatchKind.Mismatch)
+                    {
+                        Reporter.Error.WriteLine(string.Format(LocalizableStrings.TemplateNotValidGivenTheSpecifiedFilter, templateGroup[0].Info.Name).Bold().Red());
+                        anythingReported = true;
+                    }
+                }
+                else
+                {   // this really shouldn't ever happen. But better to have a generic error than quietly ignore the partial match.
+                    Reporter.Error.WriteLine(string.Format(LocalizableStrings.GenericPlaceholderTemplateContextError, templateGroup[0].Info.Name).Bold().Red());
+                    anythingReported = true;
+                }
+            }
+
+            if (remainingPartialMatchGroups.Count > 0)
+            {
+                Reporter.Error.WriteLine(LocalizableStrings.TemplateMultiplePartialNameMatches.Bold().Red());
+                anythingReported = true;
+            }
+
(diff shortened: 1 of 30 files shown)


seancpeters:

seancpeters:
@dotnet-bot test OpenSUSE42.1 x64 Debug Build please
@dotnet-bot CentOS7.1 x64 Debug Build please

seancpeters:

seancpeters:
@dotnet-bot test OpenSUSE42.1 x64 Debug Build please
@dotnet-bot CentOS7.1 x64 Debug Build please

seancpeters:

seancpeters:
I found a bug with how context (type) matches are handled. It was introduced in April 2017 with this PR https://github.com/dotnet/templating/pull/591 which changed the definition of IsPartialMatch. Going to fix it as part of this PR.

src/Microsoft.TemplateEngine.Edge/Settings/FilteredTemplateInfo.cs
```diff
@@ -10,27 +13,88 @@ public class FilteredTemplateInfo : IFilteredTemplateInfo
...
+                _matchDisposition = new List<MatchInfo>();
+            }
+
+            _dispositionOfDefaults = new List<MatchInfo>();
         }
 
         public ITemplateInfo Info { get; }
 
-        public IReadOnlyList<MatchInfo> MatchDisposition { get; set; }
+        public IReadOnlyList<MatchInfo> MatchDisposition
```
mlorbetske: Setter removed from a public member of a public type

src/Microsoft.TemplateEngine.Edge/Settings/FilteredTemplateInfo.cs
```diff
@@ -10,27 +13,88 @@ public class FilteredTemplateInfo : IFilteredTemplateInfo
...
+                                        )
+                                    );
+
+        // a list of the invalid parameter names for the template, if any.
+        // analogous to INewCommandInput.RemainingArguments
+        public IReadOnlyList<string> InvalidParameterNames => MatchDisposition.Where(x => x.Kind == MatchKind.InvalidParameterName)
+                                                                            .Select(x => x.ChoiceIfLocationIsOtherChoice).ToList();
 
-        // There is at least one parameter match And all parameter matches are exact
-        public bool IsParameterMatch => !HasParameterMismatch && MatchDisposition.Any(x => x.Location == MatchLocation.OtherParameter);
```
mlorbetske: Looks like some public members have also been removed

src/Microsoft.TemplateEngine.Edge/Template/IFilteredTemplateInfo.cs
```diff
@@ -9,16 +9,29 @@ public interface IFilteredTemplateInfo
...
+
+        void AddDefaultDisposition(MatchInfo newDisposition);
+
         bool IsMatch { get; }
 
         bool IsPartialMatch { get; }
 
         bool HasParameterMismatch { get; }
 
-        bool IsParameterMatch { get; }
```
mlorbetske: Interface modification needs either a breaking change announce or a "2" flavor

src/Microsoft.TemplateEngine.Cli/HelpForTemplateResolution.cs
```diff
@@ -0,0 +1,147 @@
...
+            {
+                Reporter.Error.WriteLine(LocalizableStrings.PartialTemplateMatchSwitchesNotValidForAllMatches.Bold().Red());
+                foreach (string flag in invalidParams)
+                {
+                    Reporter.Error.WriteLine($"  {flag}".Bold().Red());
+                }
+            }
+        }
+
+        // get a better name for this, it's unclear what it's doing.
```
mlorbetske: Is this something we want to address now or in the future? I'm OK either way - if it's for the future, let's prefix with `TODO :` so it's easier to find in the future

src/Microsoft.TemplateEngine.Cli/HelpForTemplateResolution.cs
```diff
@@ -0,0 +1,147 @@
...
+                }
+            }
+
+            if (remainingPartialMatchGroups.Count > 0)
+            {
+                Reporter.Error.WriteLine(LocalizableStrings.TemplateMultiplePartialNameMatches.Bold().Red());
+                anythingReported = true;
+            }
+
+            Reporter.Error.WriteLine();
```
mlorbetske: It'd be good to only write the empty line if we wrote anything to stderr for cases where consumers look for non-empty stderr to determine failure

mlorbetske: I guess that we may be guaranteed to write something here anyway, so it might be OK

src/Microsoft.TemplateEngine.Cli/HelpForTemplateResolution.cs
seancpeters: It should be a "TODO:", I'll add it.

src/Microsoft.TemplateEngine.Cli/HelpForTemplateResolution.cs
```diff
@@ -0,0 +1,147 @@
...
+            {
+                foreach (string paramName in template.GetInvalidParameterNames())
+                {
+                    if (!invalidCounts.ContainsKey(paramName))
+                    {
+                        invalidCounts[paramName] = 1;
+                    }
+                    else
+                    {
+                        invalidCounts[paramName]++;
```
mlorbetske: nit `++invalidCounts[paramName]` as we don't need the non-incremented value

src/Microsoft.TemplateEngine.Cli/HelpForTemplateResolution.cs
```diff
@@ -0,0 +1,147 @@
...
+                    }
+                }
+            }
+
+            invalidForAllTemplates = invalidCounts.Where(x => x.Value == templateList.Count).Select(x => x.Key).ToList();
```
mlorbetske: I think there's an extension somewhere for doing bucketing - we're iterating the set twice for matching counts vs non-matching counts, which we could actually do in one pass

src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -466,14 +477,31 @@ private CreationResultStatus EnterAmbiguousTemplateManipulationFlow(TemplateList
                 return CreationResultStatus.NotFound;
             }
 
-            if (!TemplateListResolver.ValidateRemainingParameters(_commandInput, out IReadOnlyList<string> invalidParams) || (!_commandInput.IsListFlagSpecified && !string.IsNullOrEmpty(TemplateName)))
+            // handling when there are parameter mismatches. There may be other issues too.
+            IReadOnlyList<ITemplateMatchInfo> templatesForDisplay = GetTemplatesToDisplayInfoAbout(templateResolutionResult);
+            HelpForTemplateResolution.GetParametersInvalidForTemplatesInList(templatesForDisplay, out IReadOnlyList<string> invalidForAllTemplates, out IReadOnlyList<string> invalidForSomeTemplates);
+            if (invalidForAllTemplates.Any() || invalidForSomeTemplates.Any())
```
mlorbetske: These should both probably just be `.Count > 0`

src/Microsoft.TemplateEngine.Cli/TemplateListResolutionResult.cs
```diff
@@ -10,34 +10,106 @@ namespace Microsoft.TemplateEngine.Cli
...
             }
+
+            invokableTemplates = null;
+            return false;
+        }
+
+        public bool TryGetSingularInvokableMatch(out ITemplateMatchInfo template)
+        {
+            IReadOnlyList<ITemplateMatchInfo> invokableMatches = CoreMatchedTemplates.Where(x => x.IsInvokableMatch()).ToList();
+            if (invokableMatches.Count() == 1)
```
mlorbetske: `.Count` rather than `.Count()`

src/Microsoft.TemplateEngine.Cli/TemplateListResolutionResult.cs
```diff
@@ -10,34 +10,106 @@ namespace Microsoft.TemplateEngine.Cli
...
+            {
+                template = invokableMatches.First();
```
mlorbetske: `[0]` rather than `.First()`

src/Microsoft.TemplateEngine.Cli/TemplateListResolver.cs
```diff
@@ -321,23 +260,20 @@ private static IReadOnlyList<IFilteredTemplateInfo> FilterTemplatesOnParameters(
...
                         else
                         {   // the name is not known
-                            dispositionForTemplate.Add(new MatchInfo { Location = MatchLocation.OtherParameter, Kind = MatchKind.InvalidParameterName, ChoiceIfLocationIsOtherChoice = unmatchedParamName });
+                            // TODO: reconsider storing the canonical in this situation. It's not really a canonical since the param is unknown.
+                            template.AddDisposition(new MatchInfo { Location = MatchLocation.OtherParameter, Kind = MatchKind.InvalidParameterName, ChoiceIfLocationIsOtherChoice = unmatchedParamName });
                         }
                     }
                 }
-                catch
+                catch (CommandParserException ex)
```
mlorbetske: Are there any other cases that could throw here? Should we add a general catch in addition to the specific one that has something to the effect of "unexpected error"? We could end up dumping a stack here where we wouldn't have before

src/Microsoft.TemplateEngine.Edge/Settings/TemplateMatchInfo.cs
```diff
@@ -0,0 +1,75 @@
...
+        public TemplateMatchInfo(ITemplateInfo info)
+        {
+            Info = info;
+            _matchDisposition = new List<MatchInfo>();
+            _dispositionOfDefaults = new List<MatchInfo>();
+        }
+
+        public ITemplateInfo Info { get; }
+
+        public IReadOnlyList<MatchInfo> MatchDisposition
```
mlorbetske: Nit: we can simplify the these bodies to just be expressions:

```
public IReadOnlyList<MatchInfo> MatchDisposition => _matchDisposition.ToList();
```

src/Microsoft.TemplateEngine.Edge/Settings/TemplateMatchInfo.cs
```diff
@@ -0,0 +1,75 @@
...
+            {
+                _dispositionOfDefaults.Add(newDisposition);
+            }
+            else
+            {
+                _matchDisposition.Add(newDisposition);
+            }
+        }
+
+        public bool IsMatch => MatchDisposition.Count > 0 && !MatchDisposition.Any(x => x.Kind == MatchKind.Mismatch);
```
mlorbetske: `!MatchDisposition.Any(x => x.Kind == MatchKind.Mismatch);`
vs
`MatchDisposition.All(x => x.Kind != MatchKind.Mismatch);`

minimal difference in this case (as success iterates the whole set), but trades inverting the result with eq checks inside with no outer inversion but neq checks inside. Not worth changing if you feel the current way is more readable

src/Microsoft.TemplateEngine.Edge/Template/ITemplateMatchInfo.cs
```diff
@@ -0,0 +1,23 @@
+using Microsoft.TemplateEngine.Abstractions;
+using System;
+using System.Collections.Generic;
+using System.Text;
```
mlorbetske: Nit: sort usings

src/Microsoft.TemplateEngine.Edge/Template/TemplateMatchInfoEqualityComparer.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Microsoft.TemplateEngine.Edge.Template
+{
+    public class TemplateMatchInfoEqualityComparer : IEqualityComparer<ITemplateMatchInfo>
+    {
+        public static IEqualityComparer<ITemplateMatchInfo> Default { get; } = new TemplateMatchInfoEqualityComparer();
+
+        public bool Equals(ITemplateMatchInfo x, ITemplateMatchInfo y)
+        {
+            return ReferenceEquals(x.Info, y.Info) || (x != null && y != null && x.Info != null && y.Info != null && string.Equals(x.Info.Identity, y.Info.Identity, StringComparison.Ordinal));
```
mlorbetske: if x or y were null, this reference equals would have thrown before making it to the null check. Null propagation may make things cleaner here

src/Microsoft.TemplateEngine.Edge/Template/MatchInfo.cs
```diff
@@ -1,11 +1,19 @@
-﻿namespace Microsoft.TemplateEngine.Edge.Template
+namespace Microsoft.TemplateEngine.Edge.Template
 {
     public struct MatchInfo
```
mlorbetske: We should explicitly mark the layout as sequential here & verify that the additional fields don't break against clients that haven't been rebuilt

src/Microsoft.TemplateEngine.Cli/HelpForTemplateResolution.cs
seancpeters: Adding the check, it's possible nothing was written.

src/Microsoft.TemplateEngine.Edge/Settings/TemplateMatchInfo.cs
seancpeters: I like what you suggested better.

src/Microsoft.TemplateEngine.Cli/TemplateListResolver.cs
seancpeters: Adding a general catch, but it's going to do the same thing as CommandParserException for now. Not much else we can do  in this situation. We could create another MatchKind and/or MatchLocaltion in addition to "Unspecified", have the caller of this check for that sort of disposition, and bail with an appropriate message. Not sure what's best, at this point, a template that catches and adds the unspecified disposition merely gets treated as a non-match.

seancpeters:

mlorbetske:

seancpeters:

seancpeters:

seancpeters:

seancpeters:

seancpeters:

PR #1245, opened 2017-09-19
File rename from multiple parameters

seancpeters:
Allows multiple parameter-based renames to be applied to the same file. Matching & replacements occur left-to-right, and are not allowed to overlap. This means that a replaced substring will not be considered for additional replacements (but the remainder of the string is still eligible for replacements).


seancpeters:
@dotnet-bot test Ubuntu x64 Release Build  please

chris-cummings:
Is this only going to be fixed in 2.1? No fix for 2.0?

PR #1255, opened 2017-09-28
Parameter value standardization for telemetry

seancpeters:


seancpeters:
Derp, this has my usual line ending changes in additional to the real changes.

src/Microsoft.TemplateEngine.Cli/TelemetryHelper.cs
```diff
@@ -0,0 +1,43 @@
...
+using Microsoft.TemplateEngine.Abstractions;
+
+namespace Microsoft.TemplateEngine.Cli
+{
+    public static class TelemetryHelper
+    {
+        // Checks if the input parameter value is a valid choice for the parameter, and returns the canonical value, or defaultValue if there is no appropriate canonical value.
+        // If the parameter or value is null, return defaultValue.
+        // For this to return other than the defaultValue, one of these must occur:
+        //  - the input value must either exactly match one of the choices (case-sensitive)
```
seancpeters: self review: to be consistent with parameter handling when matching templates, this should be case insensitive comparisons on the parameter values.

mlorbetske:
Let's update the event names (names go to all lower case, words separated by '-', create gets a '-' between the command name and 'create'), add the flag indicating whether the template is 3rd party & send hashes of the template name so we can get an idea of the number of different kinds of templates being made.

seancpeters:
The 4th commit standardizes the event names to dash-separated-words. It also reworks where the help flag telemetry event is emitted - ensures it's always emitted when the help flag is specified.

Also added hashing of 3rd party template names. But it's currently using GetHashCode() which gives varying results on the same value over separate runs. We need something better.

PR #1259, opened 2017-10-02
Add missing @ symbol to AccessDenied.cshtml

scottaddie:
Add a missing `@` symbol to the `AccessDenied.cshtml` view, as reported in https://github.com/aspnet/Docs/pull/4476.


scottaddie:
The ASP.NET templates have moved to the aspnet/templating repo. I will open a PR over there instead.

PR #1268, opened 2017-10-05
fixed a bug with installs that dumped a stack

seancpeters:
Wildcards in local package specifications was causing installation to throw.


seancpeters:

PR #1273, opened 2017-10-07
Initial work for semantic version datatype and tests

mlorbetske:
Needs more comments and probably some more tests still

### DIFF
--- a/src/Microsoft.TemplateEngine.Utils/SemanticVersion.cs
+++ b/src/Microsoft.TemplateEngine.Utils/SemanticVersion.cs
@@ -3,10 +3,22 @@ using System.Globalization;
 
 namespace Microsoft.TemplateEngine.Utils
 {
+    /// <summary>
+    /// Wrapper for working with Semantic Versioning v2.0 (http://semver.org/)
+    /// </summary>
     public class SemanticVersion : IEquatable<SemanticVersion>, IComparable
     {
         private readonly int _hashCode;
 
+        /// <summary>
+        /// Creates a new semantic version object
+        /// </summary>
+        /// <param name="originalText">The original text that had been parsed</param>
+        /// <param name="major">The interpreted major version number</param>
+        /// <param name="minor">The interpreted minor version number</param>
+        /// <param name="patch">The interpreted patch number</param>
+        /// <param name="prereleaseInfo">The prerelease information section, if any (should be null if none was found)</param>
+        /// <param name="buildMetadata">The build metadata section, if any (should be null if none was found)</param>
         private SemanticVersion(string originalText, int major, int minor, int patch, string prereleaseInfo, string buildMetadata)
         {
             OriginalText = originalText;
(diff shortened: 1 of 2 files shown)


src/Microsoft.TemplateEngine.Utils/SemanticVersion.cs
```diff
@@ -0,0 +1,438 @@
...
+            if (majorEnd == 0 || (source[0] == '0' && majorEnd > 1))
+            {
+                version = null;
+                return false;
+            }
+
+            if (majorEnd == source.Length || source[majorEnd] != '.')
+            {
+                tail = majorEnd;
+                goto HandlePostVersion;
```
mlorbetske: thinking about doing nested `else`s instead of the goto, but it's prettier flat

src/Microsoft.TemplateEngine.Utils/SemanticVersion.cs
```diff
@@ -0,0 +1,438 @@
...
+            if (minorEnd == majorEnd + 1 || (source[majorEnd + 1] == '0' && minorEnd - majorEnd - 1 > 1))
+            {
+                version = null;
+                return false;
+            }
+
+            if (minorEnd == source.Length || source[minorEnd] != '.')
+            {
+                tail = minorEnd;
+                goto HandlePostVersion;
```
mlorbetske: This goto too

src/Microsoft.TemplateEngine.Utils/SemanticVersion.cs
```diff
@@ -0,0 +1,438 @@
...
+
+                if (tail == end - 1)
+                {
+                    version = null;
+                    return false;
+                }
+
+                for (int i = tail + 1; i < end; ++i)
+                {
+                    if (!(
```
mlorbetske: Should probably factor this out into its own method

src/Microsoft.TemplateEngine.Utils/SemanticVersion.cs
```diff
@@ -0,0 +1,438 @@
...
+            {
+                if (tail == source.Length - 1)
+                {
+                    version = null;
+                    return false;
+                }
+
+                for (int i = tail + 1; i < source.Length; ++i)
+                {
+                    if (!(
```
mlorbetske: Would refer to the same method as above

src/Microsoft.TemplateEngine.Utils/SemanticVersion.cs
```diff
@@ -0,0 +1,438 @@
...
+                        source[i] == '-' || source[i] == '.' ||
+                        (source[i] >= 'A' && source[i] <= 'Z') ||
+                        (source[i] >= 'a' && source[i] <= 'z') ||
+                        (source[i] >= '0' && source[i] <= '9')
+                        ))
+                    {
+                        version = null;
+                        return false;
+                    }
+                }
```
mlorbetske: Per the spec, there should really be a check here that disallows period separated segments that consist only of digits that begin with a 0 that are more than one character in length

src/Microsoft.TemplateEngine.Utils/SemanticVersion.cs
```diff
@@ -0,0 +1,438 @@
+using System;
+using System.Globalization;
+
+namespace Microsoft.TemplateEngine.Utils
+{
+    public class SemanticVersion : IEquatable<SemanticVersion>, IComparable
```
mlorbetske: Add a link to http://semver.org/ in a comment on the class

test/Microsoft.TemplateEngine.Utils.UnitTests/SemanticVersionTests.cs
```diff
@@ -0,0 +1,207 @@
...
+        [InlineData("1.0.0-alpha1+build42", "1.0.0-alpha1+build42", null, true)]
+
+        [InlineData("1.0.0-alpha1+build42", "1.0.0-beta1+build42", false, false)]
+        [InlineData("1.0.0-alpha1+build42", "1.0.0-beta1+build42", false, true)]
+        [InlineData("1.0.0-alpha1+build42", "1.0.0-beta1+build42", null, false)]
+
+        [InlineData("1.0.0-alpha1+build42", "1.0.0-alpha1+build43", false, false)]
+        [InlineData("1.0.0-alpha1+build42", "1.0.0-alpha1+build43", false, true)]
+        [InlineData("1.0.0-alpha1+build42", "1.0.0-alpha1+build43", null, false)] // This is expected to differ from the Operators test - as build metada has to be accounted for when performing a comparison sort
+        public void SemanticVersionCompareTo(string ver1, string ver2, bool? greater, bool equal)
```
mlorbetske: Correct the failure messages to indicate the the comparisons that are actually being done, the current messages imply usage of the operators rather than checking the numeric result of `CompareTo(SemanticVersion)`

test/Microsoft.TemplateEngine.Utils.UnitTests/SemanticVersionTests.cs
```diff
@@ -0,0 +1,207 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Xunit;
+
+namespace Microsoft.TemplateEngine.Utils.UnitTests
+{
+    public class SemanticVersionTests
```
mlorbetske: Add a few tests that check:
* `bool Equals(object)`
* `int CompareTo(object)`
* `int GetHashCode()`
* `int CompareTo(SemanticVersion, out bool)` - this is being verified indirectly, but a test or two the checks the build metadata flag explicitly would be good
* `string ToString()`
* Tests verifying the behavior of null values with `bool TryParse(string, out SemanticVersion)` `int CompareTo(object)`/`int CompareTo(SemanticVersion)`, `int CompareTo(SemanticVersion, out bool)` and operators

mlorbetske:

src/Microsoft.TemplateEngine.Utils/SemanticVersion.cs
```diff
@@ -0,0 +1,438 @@
...
+                return false;
+            }
+
+            source = source.Trim();
+            int majorEnd = 0,
+                minorEnd = 0,
+                patchEnd = 0,
+                tail = 0;
+
+            for(majorEnd = 0; majorEnd < source.Length && source[majorEnd] >= '0' && source[majorEnd] <= '9'; ++majorEnd)
```
mlorbetske: indentation's messed up here - format

src/Microsoft.TemplateEngine.Utils/SemanticVersion.cs
```diff
@@ -0,0 +1,438 @@
...
+            tail = patchEnd;
+
+HandlePostVersion:
+            int major = int.Parse(source.Substring(0, majorEnd), NumberStyles.None, CultureInfo.InvariantCulture);
+            int minor = minorEnd > majorEnd ? int.Parse(source.Substring(majorEnd + 1, minorEnd - majorEnd - 1), NumberStyles.None, CultureInfo.InvariantCulture) : 0;
+            int patch = patchEnd > minorEnd ? int.Parse(source.Substring(minorEnd + 1, patchEnd - minorEnd - 1), NumberStyles.None, CultureInfo.InvariantCulture) : 0;
+            string prerelease = null;
+            string metadata = null;
+
+            if (tail < source.Length && source[tail] == '-')
```
mlorbetske: Add a comment that this is the check for the prerelease info section

src/Microsoft.TemplateEngine.Utils/SemanticVersion.cs
```diff
@@ -0,0 +1,438 @@
...
+                        version = null;
+                        return false;
+                    }
+                }
+
+                prerelease = source.Substring(tail + 1, end - tail - 1);
+                tail = end;
+            }
+
+            if (tail < source.Length && source[tail] == '+')
```
mlorbetske: Add a comment that this is a check for the build metadata section

test/Microsoft.TemplateEngine.Utils.UnitTests/SemanticVersionTests.cs
```diff
@@ -0,0 +1,207 @@
...
+        [InlineData("0.0.0", true)]
+        [InlineData("01.0.0", false)]
+        [InlineData("0.01.0", false)]
+        [InlineData("0.0.01", false)]
+        [InlineData("1.0.0-", false)]
+        [InlineData("1.0.0+", false)]
+        [InlineData("1.0.0-+", false)]
+        [InlineData("1.0.0-1.1", true, 1, 0, 0, "1.1")]
+        [InlineData("1.0.0-1!1", false)]
+        [InlineData("1.0.0-00", true, 1, 0, 0, "00")] //This is acutally valid, 00 just won't be treated as a numeric segment
```
mlorbetske: This one and the one below may change depending on whether the check to ensure any digit-only sections have leading zeros gets added to TryParse

test/Microsoft.TemplateEngine.Utils.UnitTests/SemanticVersionTests.cs
```diff
@@ -0,0 +1,234 @@
...
+        [InlineData("1.0.0+", false)]
+        [InlineData("1.0.0-+", false)]
+        [InlineData("1.0.0-1.1", true, 1, 0, 0, "1.1")]
+        [InlineData("1.0.0-1!1", false)]
+        [InlineData("1.0.0-00", false)]
+        [InlineData("1.0.0-a.00.b", false)]
+        [InlineData("1.0.0-0", true, 1, 0, 0, "0")]
+        [InlineData("1.0.0+1.1", true, 1, 0, 0, null, "1.1")]
+        [InlineData("1.0.0+1!1", false)]
+        [InlineData("1.0.0+00", true, 1, 0, 0, null, "00")] //This is acutally valid, 00 just won't be treated as a numeric segment
```
mlorbetske: This comment is wrong, this portion doesn't have the numeric guidelines & isn't intended for use in a meaningful comparison

test/Microsoft.TemplateEngine.Utils.UnitTests/SemanticVersionTests.cs
mlorbetske: This is done

src/Microsoft.TemplateEngine.Utils/SemanticVersion.cs
```diff
@@ -0,0 +1,674 @@
...
+        private int ComparePrereleaseInfo(SemanticVersion other)
+        {
+            if (PrereleaseInfo != null)
+            {
+                if (other.PrereleaseInfo != null)
+                {
+                    int prereleaseCompare = CompareToPrereleaseInfo(other.PrereleaseInfo);
+
+                    if (prereleaseCompare != 0)
+                    {
```
seancpeters: Nit: This check isn't strictly necessary, could just return the prereleaeCompare value here. If it's 0, we'll fall through to the final return of 0. But it might be best to keep the check, in case other things change which cause the fallthrough to be different.

src/Microsoft.TemplateEngine.Utils/SemanticVersion.cs
```diff
@@ -0,0 +1,674 @@
...
+        }
+
+        private static bool ValidateAllDigitSegments(string prerelease)
+        {
+            string[] parts = prerelease.Split('.');
+
+            for (int i = 0; i < parts.Length; ++i)
+            {
+                if (!IsNumericSegment(parts[i], out bool isInvalidBecauseOfLeadingZero) && isInvalidBecauseOfLeadingZero)
+                {
```
seancpeters: The name of this method makes me think it's checking tat all segments are numeric. But the implementation appears to return true unless there is a segment that starts with a lead zero but is otherwise numeric.

test/Microsoft.TemplateEngine.Utils.UnitTests/SemanticVersionTests.cs
```diff
@@ -0,0 +1,234 @@
...
+        [InlineData("1.0.0-1.1", true, 1, 0, 0, "1.1")]
+        [InlineData("1.0.0-1!1", false)]
+        [InlineData("1.0.0-00", false)]
+        [InlineData("1.0.0-a.00.b", false)]
+        [InlineData("1.0.0-0", true, 1, 0, 0, "0")]
+        [InlineData("1.0.0+1.1", true, 1, 0, 0, null, "1.1")]
+        [InlineData("1.0.0+1!1", false)]
+        [InlineData("1.0.0+00", true, 1, 0, 0, null, "00")]
+        [InlineData("1.0.0+0", true, 1, 0, 0, null, "0")]
+        public void SemanticVersionParse(string source, bool expectValid, int major = 0, int minor = 0, int patch = 0, string prerelease = null, string metadata = null)
```
seancpeters: Nit: perhaps add test cases that have > 3 numeric segments, like "1.2.3.4", "1.2.3.4-preview1", "1.2.3.4+build1", etc.

test/Microsoft.TemplateEngine.Utils.UnitTests/SemanticVersionTests.cs
```diff
@@ -0,0 +1,234 @@
...
+        [InlineData("1.0.0-alpha1+build42", "1.0.0-alpha1+build42", null, true)]
+
+        [InlineData("1.0.0-alpha1+build42", "1.0.0-beta1+build42", false, false)]
+        [InlineData("1.0.0-alpha1+build42", "1.0.0-beta1+build42", false, true)]
+        [InlineData("1.0.0-alpha1+build42", "1.0.0-beta1+build42", null, false)]
+
+        [InlineData("1.0.0-alpha1+build42", "1.0.0-alpha1+build43", true, true)]
+        [InlineData("1.0.0-alpha1+build42", "1.0.0-alpha1+build43", false, true)]
+        [InlineData("1.0.0-alpha1+build42", "1.0.0-alpha1+build43", null, true)] // This is expected to differ from the CompareTo test - as build metada should be ignored
+        public void SemanticVersionOperators(string ver1, string ver2, bool? greater, bool equal)
```
seancpeters: Nit: perhaps add some test cases where the version strings don't parse.

seancpeters:

PR #1274, opened 2017-10-07
Update to xUnit.net 2.3 RTM

bradwilson:
This updates the project templates, as well as the unit tests, to use xUnit.net 2.3 RTM.

Note that updating the unit test projects has caused several build warnings as a result of the fact that 2.3 now includes Roslyn code analyzers, and there are several unit tests that trigger code analysis warnings. There are a few things I could do about this:

- Nothing; someone on your team can fix the tests later
- Fix the tests myself
- For your unit test projects, stop using the `xunit` meta-package, and instead use `xunit.core` and `xunit.assert` (which will cause the analyzers not to get added to your project)
- For your unit test projects, revert back to 2.2 (but leave the templates at 2.3)


mlorbetske:
Thanks! Changes look good to me. @Faizan2304 any thoughts on what'd be the preferred way to proceed in regard to the warnings @bradwilson mentioned?

ManishJayaswal:
@genlu @jmarolf  - can you guys also review this please?

Issue #1280, opened 2017-10-10
Scanning a directory that contains a .dll results in the directory and all files being deleted


mrward:
When a directory is passed to the TemplateCache Scan method, and a subdirectory contains .dll files, then the directory and all files are deleted. When scanning a directory this should not happen since this may contain the original project and associated template files.

I created a simple console app that uses the IDE Bootstrapper to reproduce the problem:

https://github.com/mrward/TestDotNetTemplateInstaller

Building and running this console application results in the TestDotNetTemplateInstaller project directory being deleted. This was tested on the Mac with Mono 5.4.

Debugging the TemplateCache what happens is that when a .dll file is found then the directory is deleted in the [ScanForComponents method](https://github.com/dotnet/templating/blob/bfc767e526213c069efacf88f676ececb3db2070/src/Microsoft.TemplateEngine.Edge/Settings/TemplateCache.cs#L264).

The dotnet cli does not seem to have the same problem. Running `dotnet new -i /path/to/template` using a directory that has a template.json but also .dll files in a subdirectory works as expected and no project files are removed.

seancpeters:
@mrward - Which branch of dotnet/templating are you seeing this problem with?

mrward:
I built 2.1.0-preview1 locally whilst debugging this. The repro console project uses the latest NuGet packages (1.0.0-beta2-20171009-310) that I could find from the templating MyGet feed.

seancpeters:
Ok, I believe I'm repro'ing this issue, but want to make sure we're doing the same thing (more or less). 

First, I created a test template and installed it successfully using `dotnet new3 -I <path\to\template\>`. This did not delete the template directory.

Second, I went into the test template and ran `dotnet build` on it, resulting in a .dll under `bin\debug\netcoreapp2.0`. Then I installed the template again, and it deleted the directory.

Looking into it now...

mrward:
Testing `dotnet new -i /path/to/template` I was using .NET Core 2.0 which seems to be OK if there are .dlls in the bin directory. I did not try a later .NET Core SDK. I could only repro using the TemplateCache directly or using the bootstrapper.

seancpeters:
@mrward - I've merged a fix for this into rel/2.1.0_preview-1, please let me know if you have additional problems.

mlorbetske:
Closing this since the fix has been merged

PR #1281, opened 2017-10-10
fixes: template dir with dll was getting deleted if no content found

seancpeters:
Fixing the issue reported in #1280. If a directory being installed contains a .dll but isn't a component, the install directory was getting deleted.


seancpeters:

PR #1303, opened 2017-10-26
Allow the specification of DataType on non-parameter type symbols

mlorbetske:
Bind still always infers, this will need to be looked at


seancpeters:

natemcmaster:
Is "2.0.3" still going to be interpreted as a number if I don't specific 'datatype'?

mlorbetske:
Until additional changes are made to use invariant culture (or specified culture) for inference for configuration values, yes, they would be.

PR #1318, opened 2017-11-03
reworked built-in glob configs to configure from json snippets

seancpeters:
Regarding performance differences from before and after this change, I did some nominal testing and didn't see any real differences. I ran the mvc1.1 template 4 times against the code before and after my changes, with negligible timing differences:

New way:
 Content generation time 1158.7476
 Content generation time 976.8154
 Content generation time 1069.5988
 Content generation time 1022.9127
   
Old way
Content generation time 957.0962
Content generation time 1107.2132
Content generation time 975.9659
Content generation time 1142.1048

### DIFF
--- a/src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/Config/ConditionalConfig.cs
+++ b/src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/Config/ConditionalConfig.cs
@@ -17,6 +17,16 @@ namespace Microsoft.TemplateEngine.Orchestrator.RunnableProjects.Config
 
         public IEnumerable<IOperationProvider> ConfigureFromJObject(JObject rawConfiguration, IDirectory templateRoot)
         {
+            IEnumerable<IOperationProvider> operations = SetupFromJObject(rawConfiguration);
+
+            foreach (IOperationProvider op in operations)
+            {
+                yield return op;
+            }
+        }
+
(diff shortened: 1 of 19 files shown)


src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/Config/ConditionalConfig.cs
```diff
@@ -32,130 +42,40 @@ public IEnumerable<IOperationProvider> ConfigureFromJObject(JObject rawConfigura
             {
                 operations = ConditionalBlockCommentConfig.ConfigureFromJObject(rawConfiguration);
             }
-            else
+            else if (string.Equals(commentStyle, "builtin", StringComparison.OrdinalIgnoreCase))
```
seancpeters: Add a check to catch loops in definitions - a built-in could theoretically reference another built-in, or itself - which is an authoring error.

seancpeters:

mlorbetske:
We should take a look at what, if any impact there is to perf for this. Also, I assume this is the first step in moving to producing the core configuration from the JSON directly or is the intention to only build it in pieces?

seancpeters:
Regarding perf, see my initial PR comment - it's not definitive in any way. I can try to isolate the timings for the glob setup and compare.

Yes, this is just a first step toward more json-based configuration, but I'm still on the fence about the final form of core config setup. Using json files is more "pure", so that's probably where we'll land.

seancpeters:

seancpeters:

PR #1320, opened 2017-11-06
don't dump a stack when there's an error parsing settings.json

seancpeters:
Gives a better user experience when there is an error, for example #1319


seancpeters:

PR #1332, opened 2017-11-10
Multi template with single starts with param fix

seancpeters:
Fix for when an input parameter value is a "single starts with" choice match for multiple templates in the same group. Prevents template invocation because the match is ambiguous.

Also, a fix for what's displayed when the input matches a single template group, and each input parameter is valid for at least one template in the group, but the input parameters in combination are not valid for any templates in the group. For example, with the 1.x and 2.0 web templates installed:
`c:\Github\templating>dotnet new3 web -f netcoreapp1.1 --use-launch-settings -o W1`
image
For the 1.x web template: `-f netcoreapp1.1` is valid, but `--use-launch-settings` is not valid.
For the 2.0 web template: `-f netcoreapp1.1` is not valid (but the -f flag is), and `--use-launch-settings` is valid.


seancpeters:
@dotnet-bot test this please

seancpeters:
Closing, this has some old commits in it. Will re-create.

PR #1335, opened 2017-11-11
fix for scanning the install directory for components - part 1

seancpeters:
This as scanning based on _paths, as opposed to the user-provided install path.


mlorbetske:
As discussed, let's put the ability to install local assemblies behind a dev flag. Also, from the conversation last night about getting custom components loaded, let's try explicitly hydrating the guid to type lookup.

seancpeters:
The backwards-compatibility / update changes in ComponentManager.cs are less than ideal.

seancpeters:
This seems like a good logical unit of changes, as the scope of this work has increased. Scanner logic is separated from template cache (more install / scan refactoring to come in the future). A component can be installed, but scanning, registering of referenced packages is not in the scope of this PR.

PR #1358, opened 2017-11-29
fix for initial 'on' state of operations

seancpeters:


src/Microsoft.TemplateEngine.Core/Util/ProcessorState.cs
```diff
@@ -104,20 +104,31 @@ public ProcessorState(Stream source, Stream target, int bufferSize, int flushThr
...
+            }
+
+            if (explicitOnConfigurationRequired)
+            {
+                for (int i = 0; i < operationProviders.Count; ++i)
+                {
+                    IOperation op = operationProviders[i].GetOperation(encoding, this);
 
-                        if (explicitOnConfigurationRequired && op.IsInitialStateOn && !string.IsNullOrEmpty(op.Id))
+                    if (op != null)
```
mlorbetske: Looks like we can combine these checks

Issue #1355, opened 2017-11-23
Concurrency issue in templateengine


sbomer:
I'm running some integration tests that try to call `dotnet new`. When I execute them one at a time, they always succeed, but when I run them concurrently, I frequently get the following stack trace:

```
Error reading JObject from JsonReader. Path '', line 0, position 0.
    at Newtonsoft.Json.Linq.JObject.Load(JsonReader reader, JsonLoadSettings settings)
    at Newtonsoft.Json.Linq.JObject.Parse(String json, JsonLoadSettings settings)
    at Newtonsoft.Json.Linq.JObject.Parse(String json)
    at Microsoft.TemplateEngine.Edge.Settings.SettingsLoader.EnsureTemplatesLoaded()
    at Microsoft.TemplateEngine.Edge.Settings.SettingsLoader.WriteTemplateCache(IList`1 templates, String locale)
    at Microsoft.TemplateEngine.Edge.Settings.TemplateCache.WriteTemplateCacheForLocale(String locale, String existingCacheVersion)
    at Microsoft.TemplateEngine.Edge.Settings.TemplateCache.WriteTemplateCaches(String existingCacheVersion)
    at Microsoft.TemplateEngine.Edge.Settings.SettingsLoader.Save(TemplateCache cacheToSave)
    at Microsoft.TemplateEngine.Edge.Settings.SettingsLoader.Save()
    at Microsoft.TemplateEngine.Edge.Settings.ComponentManager.Register(Type type)
    at Microsoft.TemplateEngine.Edge.Settings.ComponentManager..ctor(ISettingsLoader loader, SettingsStore userSettings)
    at Microsoft.TemplateEngine.Edge.Settings.SettingsLoader.EnsureLoaded()
    at Microsoft.TemplateEngine.Edge.Settings.SettingsLoader.RebuildCacheFromSettingsIfNotCurrent(Boolean forceRebuild)
    at Microsoft.TemplateEngine.Cli.New3Command.<ExecuteAsync>d__59.MoveNext()
 --- End of stack trace from previous location where exception was thrown ---
    at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
    at System.Runtime.CompilerServices.ConfiguredTaskAwaitable`1.ConfiguredTaskAwaiter.GetResult()
[...]

PR #1360, opened 2017-11-30
added mutex to block an invocation if another is already running.

seancpeters:


seancpeters:

src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -79,7 +80,48 @@ public static int Run(string commandName, ITemplateEngineHost host, ITelemetryLo
             return Run(commandName, host, telemetryLogger, onFirstRun, args, null);
         }
 
+        static Mutex _entryMutex = new Mutex(false, "{5CB26FD1-32DB-4F4C-B3DC-49CFD61633D2}");
```
mlorbetske: Add explicit visibility here (private)

src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -79,7 +80,48 @@ public static int Run(string commandName, ITemplateEngineHost host, ITelemetryLo
...
         public static int Run(string commandName, ITemplateEngineHost host, ITelemetryLogger telemetryLogger, Action<IEngineEnvironmentSettings, IInstaller> onFirstRun, string[] args, string hivePath)
+        {
+            bool holdingMutex = false;
```
mlorbetske: Seems like rather than adding this flag it could be:

if (!_entryMutex.WaitOne())
{
    //exit with failure
}

try
{
    //existing stuff
}
finally
{
    _entryMutex.ReleaseMutex();
}

mlorbetske:
Fixes #1355

src/Microsoft.TemplateEngine.Cli/New3Command.cs
seancpeters: Yes, I'll make that change. I had added that flag when tracking down some abandoned mutex issues while testing and left if there for safety, but it's overkill.

PR #1369, opened 2017-12-06
Fix generated assembly version in version.props to handle large build…

phenning:
… numbers.

Assembly version will no longer increment as part of this change.


vijayrkn:
yay! vsts build passed with large build number!

PR #1375, opened 2017-12-21
fix for bug #1373

seancpeters:
When the only template match group doesn't have the default language, and no user language is input, then the template will still be chosen as long as the group only contains templates in one language.

### DIFF
--- a/src/Microsoft.TemplateEngine.Cli/TemplateListResolutionResult.cs
+++ b/src/Microsoft.TemplateEngine.Cli/TemplateListResolutionResult.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using Microsoft.TemplateEngine.Abstractions;
 using Microsoft.TemplateEngine.Edge.Template;
 
 namespace Microsoft.TemplateEngine.Cli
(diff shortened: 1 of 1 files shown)


src/Microsoft.TemplateEngine.Cli/TemplateListResolutionResult.cs
```diff
@@ -109,16 +110,46 @@ public bool TryGetAllInvokableTemplates(out IReadOnlyList<ITemplateMatchInfo> in
...
+                    {
+                        if (templateInfo.Info.Tags.TryGetValue("language", out ICacheTag languageTag))
+                        {
+                            languagesForMatches.Add(languageTag.DefaultValue);
+                        }
+                    }
+
+                    if (languagesForMatches.Count > 1)
+                    {
+                        // there are multiple languages in the possible templates. No way to decide which one to use.
```
mlorbetske: This isn't strictly true is it? Couldn't precedence disambiguate (if one language of the template were preferred by the author over the others)?

seancpeters: Up to now, we've only considered precedence among templates with the same group identity and language. I could add a check here so that if:
- all candidate templates are the same group identity
- user didn't specify a language
Then don't do the language check. These cases would fall through to the remaining existing checks, the last one being: take the highest precedence template if they're all the same group identity.

The check would change the new if at line 116 to be:
`if (_hasUserInputLanguage || _TemplateListResolver..AreAllTemplatesSameGroupIdentity(_coreMatchedTemplates))`

PR #1380, opened 2017-12-23
Make use of some newer language features

mlorbetske:


src/Microsoft.TemplateEngine.Core/Matching/TrieEvaluationDriver.cs
```diff
@@ -31,11 +30,9 @@ public TerminalLocation<T> Evaluate(byte[] buffer, int bufferLength, bool isFina
                         {
                             break;
                         }
-                        else
-                        {
-                            _evaluator.FinalizeMatchesInProgress(ref _sequenceNumber, out terminal);
-                            break;
-                        }
+
```
seancpeters: Seems simpler as:
```
if (isFinalBuffer)
{
  _evaluator.FinalizeMatchesInProgress(ref _sequenceNumber, out terminal);
}
break;
```

src/Microsoft.TemplateEngine.Core/Operations/Conditional.cs
```diff
@@ -318,32 +306,28 @@ public int HandleMatch(IProcessorState processor, int bufferLength, ref int curr
...
+
+                Debug.Assert(true, "Unknown token index: " + token);
+                return 0;   // TODO: revisit. Not sure what's best here.
             }
 
             // moves the buffer to the next token at the same level.
             // Returns false if no end token can be found at the same level.
             //      this is probably indicative of a template authoring problem, or possibly a buffer problem.
-            private bool SkipToMatchingEndif(IProcessorState processor, ref int bufferLength, ref int currentBufferPosition, ref int token)
+            private void SkipToMatchingEndif(IProcessorState processor, ref int bufferLength, ref int currentBufferPosition, ref int token)
```
seancpeters: The return value is not used where this is called from, so I can see the case for changing it from bool to void. But the return value indicates a potential token-balancing problem which could (should?) be checked for. Without it, there's no way to check for the problem.

src/Microsoft.TemplateEngine.Core/Operations/PhasedOperation.cs
```diff
@@ -37,6 +35,12 @@ public IOperation GetOperation(Encoding encoding, IProcessorState processorState
                 while (currentSource?.MoveNext() ?? false)
                 {
                     Phase c = currentSource.Current;
+
```
seancpeters: There's nothing analogous to the `c == null` check in the existing code. It seems like it can't happen unless there is a bug in the enumerator - MoveNext() should return false in this case.

src/Microsoft.TemplateEngine.Core/Operations/SetFlag.cs
```diff
@@ -105,9 +102,9 @@ public int HandleMatch(IProcessorState processor, int bufferLength, ref int curr
                 {
                     processor.Config.Flags[_owner.Name] = turnOn;
                 }
-                else if (_owner.Name == SetFlag.OperationName && turnOn)
+                else if (_owner.Name == OperationName && turnOn)
```
seancpeters: use string.equals ?

PR #1393, opened 2018-01-12
Baseline comparer

seancpeters:
Initial work for automated template output comparisons, to verify that code changes don't affect template creation.


tools/BaselineComparer/BaselineComparer/Baseline.cs
```diff
@@ -0,0 +1,87 @@
...
+                baseline = FromJObject(baselineJObject);
+            }
+            catch (Exception ex)
+            {
+                Console.WriteLine("Unable to deserialize the existing baseline data.");
+                Console.WriteLine($"Error: {ex.Message}");
+                throw;
+            }
+
+            return baseline;
```
mlorbetske: Feels like the block leading to here should be:

```
try
{
    JObject baselineObject = JObject.Parse(baselineContent);
    Baseline baseline = FromJObject(baselineJObject);
    return baseline;
}
catch (Exception ex)
{
    Console.WriteLine("Unable to deserialize the existing baseline data.");
    Console.WriteLine($"Error: {ex.Message}");
    throw;
}
```

tools/BaselineComparer/BaselineComparer/BaselineComparer.cs
```diff
@@ -0,0 +1,96 @@
...
+{
+    public class BaselineComparer
+    {
+        public BaselineComparer(string baselineFile, string newDataBaseDir)
+        {
+            _baselineFile = baselineFile;
+            _newDataBaseDir = newDataBaseDir;
+        }
+
+        private string _baselineFile;
```
mlorbetske: Looks like at least one of these fields can be made readonly

seancpeters:
With commit ab61345ad2977c8b432cf92b115596c5db47b413 this code should remain relatively stable for the rest of it's development, so I'd say it's reasonable to review at this point.

PR #1403, opened 2018-01-23
Make the full file name be matched when enumerating files for the in-memory file system

mlorbetske:
In some experimentation last night, I ran across a case where if an in-memory file system has a file that's subjected to scan and also contains a file name that has the original name as a substring, a stack overflow can occur.

### DIFF
--- a/src/Microsoft.TemplateEngine.Utils/InMemoryFileSystem.cs
+++ b/src/Microsoft.TemplateEngine.Utils/InMemoryFileSystem.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Text;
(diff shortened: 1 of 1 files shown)


src/Microsoft.TemplateEngine.Utils/InMemoryFileSystem.cs
```diff
@@ -499,7 +499,7 @@ public IEnumerable<string> EnumerateFileSystemEntries(string directoryName, stri
                 }
             }
 
-            Regex rx = new Regex(Regex.Escape(pattern).Replace("\\*", ".*").Replace("\\?", "."));
+            Regex rx = new Regex("^" + Regex.Escape(pattern).Replace("\\*", ".*").Replace("\\?", ".") + "$");
```
seancpeters: Change makes sense. But, do we want to do the replaces on the pattern and then escape the replaced value, as opposed to escaping the pattern and then doing the replacements?

PR #1405, opened 2018-01-24
Scan fix for template packs with dll's

### LINKED ISSUE #1397: Templates working on Windows - but not on Mac
Could you please have a look at this template nuget:

https://www.nuget.org/packages/IdentityServer4.Templates/

source code is here:
https://github.com/IdentityServer/IdentityServer4.Templates

The templates work as expected on Windows, on Mac the command just returns (without the usual "temlate crated" message - and directory is empty, e.g.

`dotnet new is4inmem` 

This is SDK version 2.1.4 on both platforms. This used to work a couple of days ago...

thanks.

seancpeters:
Fixes #1397 

High-level sketch of the lifecycle of files / archives to install:

Scanner.Scan() orchestration never deletes install candidates from their original location, as proposed to Scan(). (Installer deals with remote packages in the scratch dir, and doesn't remove local packages).

Scan()
* SetupMountPointScanInfoForDirectories()
-- GetOrCreateMountPointScanInfoForDirectory()
--- If the directory is an existing mount point, return it.
--- Otherwise: Setup a mount point for the source location. If the location is not under the ScratchDir, mark the source as ShouldStayInOriginalLocation. 

- ScanSourcesForComponents()
-- ScanForComponents()
--- For source not marked as ShouldStayInOriginalLocation, copy into Content/ then do the component scan from there. Otherwise leave it where it is.
--- After component scan, if nothing was found, and dealing with a copy in Content/ then delete the Content/ copy. For things not copied to Content/ do nothing.

- ScanSourcesForTemplatesAndLangPacks()
-- No copying or moving of sources. The original location as proposed to Scan() is dealt with.

- Cleanup at the end of Scan() for each of the proposed scan locations:
-- If it's a new MP and nothing was found:
--- Remove the mount point.
--- If GetOrCreateMountPointScanInfoForDirectory() made a copy into Packages/ then delete that copy.

### DIFF
--- a/src/Microsoft.TemplateEngine.Edge/Settings/Scanner.cs
+++ b/src/Microsoft.TemplateEngine.Edge/Settings/Scanner.cs
@@ -280,9 +280,10 @@ namespace Microsoft.TemplateEngine.Edge.Settings
                     // Note: no mount pount was created for this copy, so no need to release it.
                     _environmentSettings.Host.FileSystem.DirectoryDelete(diskPath, true);
                 }
-                catch
+                catch (Exception ex)
                 {
                     _environmentSettings.Host.LogDiagnosticMessage($"During ScanForComponents() cleanup, couldn't delete source copied into the content dir: {diskPath}", "Install");
+                    _environmentSettings.Host.LogDiagnosticMessage($"\tError: {ex.Message}", "Install");
                 }
             }
 
--- a/src/Microsoft.TemplateEngine.Edge/Settings/Scanner.cs
+++ b/src/Microsoft.TemplateEngine.Edge/Settings/Scanner.cs
@@ -34,6 +34,7 @@ namespace Microsoft.TemplateEngine.Edge.Settings
             ScanSourcesForComponents(sourceList, allowDevInstall);
             ScanResult scanResult = ScanSourcesForTemplatesAndLangPacks(sourceList);
 
+            // cleanup
             foreach (MountPointScanSource source in sourceList)
             {
                 if (source.AnythingFound)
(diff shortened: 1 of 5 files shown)


src/Microsoft.TemplateEngine.Edge/Settings/Scanner.cs
```diff
@@ -243,31 +260,40 @@ private bool ScanForComponents(IMountPoint mountPoint, string scanLocation, bool
...
             }
 
-            if (!anythingFound && !isInOriginalInstallLocation)
+            if (!anythingFound && isCopiedIntoContentDirectory)
             {
                 try
                 {
+                    // The source was copied to content and then scanned for components.
+                    // Nothing was found, and this is a copy that now has no use, so delete it.
+                    // Note: no mount pount was created for this copy, so no need to release it.
```
mlorbetske: nit: typo - should be "point"

src/Microsoft.TemplateEngine.Edge/Settings/Scanner.cs
```diff
@@ -277,21 +303,11 @@ private bool TryCopyContentForNonFileSystemBasedMountPoints(IMountPoint mountPoi
             }
             catch (IOException)
             {
+                _environmentSettings.Host.LogDiagnosticMessage($"Error copying scanLocation: {scanLocation} into the content dir: {path}", "Install");
                 diskPath = null;
                 return false;
             }
 
-            try
```
mlorbetske: Are these getting cleaned up somewhere else that we're able to get rid of the delete here?

src/Microsoft.TemplateEngine.Edge/Settings/Scanner.cs
seancpeters: Heh, this has become one of my standard typos lately.

src/Microsoft.TemplateEngine.Edge/Settings/Scanner.cs
seancpeters: Yes, near the end of Scan(), at line 55. 

TryCopyIntoContentForNonFileSystemBasedMountPoints() gets called from ScanForComponents() if any .dll's exist in the scan source. When the call occurs, there is no guarantee that any of those .dll's contain components. When there are no components, we still need to scan source in the packages directory for doing a template / langpack scan. 

Both scanning for components, and scanning for templates / langpacks will set source.AnythingFound = true when something is found. That flag is used at the end of Scan() to decide whether to do this delete.

seancpeters: Looking further, I found a problem - if a sources has components, it never gets deleted from Packages/ ands end up both in Packages/ and Content/, but should only be in Content/. Will fix.

src/Microsoft.TemplateEngine.Edge/TemplateUpdates/NupkgInstallUnitDescriptorFactory.cs
```diff
@@ -48,7 +48,7 @@ public bool TryCreateFromMountPoint(IMountPoint mountPoint, out IReadOnlyList<II
             descriptorList = allDescriptors;
 
             if (mountPoint.Info.Place != null
-                && Directory.Exists(mountPoint.Info.Place)  // nupkgs are expanded to files on install, so Directory.Exists() is more appropriate
+                && File.Exists(mountPoint.Info.Place)
```
mlorbetske: I think this exists check should run through IPhysicalFileSystem, shouldn't it?

seancpeters: Yep, fixed.

PR #1412, opened 2018-02-05
Add support for RegexMatch generated symbol to set a bool if a symbol matches a regex

mlorbetske:


src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/Macros/RegexMatchMacro.cs
```diff
@@ -0,0 +1,93 @@
...
+
+            if (!vars.TryGetValue(config.SourceVariable, out object working))
+            {
+                value = parameters.TryGetRuntimeValue(environmentSettings, config.SourceVariable, out object resolvedValue, true)
+                    ? resolvedValue.ToString()
+                    : string.Empty;
+            }
+            else
+            {
+                value = working?.ToString() ?? "";
```
seancpeters: Nit: string.Empty instead of ""

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/Macros/RegexMatchMacro.cs
```diff
@@ -0,0 +1,93 @@
...
+            }
+
+            bool result = Regex.IsMatch(value, config.Pattern);
```
seancpeters: Do we want to try-catch for malformed regex'es? (There are other places outside this review where we should consider doing the same).

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/Macros/Config/RegexMatchMacroConfig.cs
```diff
@@ -0,0 +1,29 @@
...
+
+        public string Type { get; }
+
+        public string SourceVariable { get; }
+
+        public string Pattern { get; }
+
+        public RegexMatchMacroConfig(string variableName, string dataType, string sourceVariable, string pattern)
+        {
+            DataType = dataType ?? "bool";
```
seancpeters: Is anything other than bool valid here? If not, might want to check & throw.

PR #1418, opened 2018-02-14
Apply 0003-Unify-Versions.patch

mlorbetske:
Part of: dotnet/source-build#177

Patch applied: [0003-Unify-Versions.patch](https://raw.githubusercontent.com/dotnet/source-build/dev/release/2.0/patches/templating/0003-Unify-Versions.patch)


mlorbetske:
@seancpeters the changes looked identical as it had the changes from the previous two patches as well. The only update as part of this is the CLI parser version bump

PR #1425, opened 2018-02-22
Support for defining defaults for when switches are given without values

seancpeters:
The new interface IAllowDefaultIfOptionWithoutValue has been implemented by relevant classes to support the new default type. All the code dealing with this is setup to also deal with concretes that don't implement the interface, which just causes things to work like they did prior to this change.


src/Microsoft.TemplateEngine.Cli/CommandParsing/CommandParserSupport.cs
```diff
@@ -95,23 +95,17 @@ public static HashSet<string> ArgsForBuiltInCommands
                     aliasesForParam.Add(shortVersion);
                 }
 
-                if (string.Equals(parameter.DataType, "choice", StringComparison.OrdinalIgnoreCase))
+                if (parameter is IAllowDefaultIfOptionWithoutValue parameterWithNoValueDefault
+                    && !string.IsNullOrEmpty(parameterWithNoValueDefault.DefaultIfOptionWithoutValue))
```
mlorbetske: Could empty string be a valid default?

seancpeters: For the normal default we don't differentiate between null & empty string, which effectively makes an empty string (regular) default not possible. I think we should treat possible values for this new default the same as the regular default. My thought is that we shouldn't allow empty string defaults, but it would be possible. We should discuss the possibilities.

src/Microsoft.TemplateEngine.Cli/HelpAndUsage/TemplateDetailsDisplay.cs
```diff
@@ -134,34 +135,67 @@ private static void ShowParameterHelp(IReadOnlyDictionary<string, string> inputP
...
+                        // Below, there's a check if the user entered a value. If so, set it.
+                        string resolvedValue = resolvedValueObject?.ToString() ?? string.Empty;
 
-                        if (!string.IsNullOrEmpty(resolvedValue)
-                            && !string.IsNullOrEmpty(param.DefaultValue)
-                            && !string.Equals(param.DefaultValue, resolvedValue))
+                        if (!string.IsNullOrEmpty(resolvedValue))
                         {
-                            configuredValue = resolvedValue;
+                            if (string.Equals(param.DataType, "bool", StringComparison.OrdinalIgnoreCase))
```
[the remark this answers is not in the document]
seancpeters: Definitely will make the change, seems more readable.

seancpeters:
Merging this into the dotnet:rel/2.1.0-preview2-template_info_reader_1.0.0.2 branch so I can base other work off it. We can tweak these changes if needed before merging dotnet:rel/2.1.0-preview2-template_info_reader_1.0.0.2 back into a mainline branch.

PR #1428, opened 2018-02-22
refactored the template info readers, to facilitate easier future changes

seancpeters:


src/Microsoft.TemplateEngine.Edge/Settings/TemplateInfoReaders/TemplateInfoReaderVersion1_0_0_0.cs
```diff
@@ -1,20 +1,33 @@
...
+        }
+
+        public TemplateInfoReaderVersion1_0_0_0(JObject jObject)
+        {
+            _jObject = jObject;
+        }
+
+        protected readonly JObject _jObject;
+
+        public virtual TemplateInfo FromJObject()
```
mlorbetske: Name of this method is a bit weird

seancpeters: Good point, I'll change it to Read()

PR #1431, opened 2018-02-27
Support for multiple short names in a template

seancpeters:
Also support for different short names among the templates in a group.

Short name can be either a string or an array of strings. When in array context, the first entry in the array will be the “preferred” short name for the template.

**Displaying a list of templates using dotnet new**: when there are multiple templates installed with the same group identity, those templates are coalesced into one entry in the list, using info from the template with the highest precedence when there is a need to choose among the templates (this is existing behavior). The displayed short name will be preferred short name from the highest precedence template in the group.

**Attempting to invoke a template using dotnet new**: Based on the user inputs, a search is performed on the installed templates to determine the best matched template. The first argument on the command line represents the template “name”, which is used for substring matching on the template name and short name. When there are multiple templates installed with the same group identity, all the short names for all the templates in the group will be considered as short names for every template in the group. This search behavior will be applicable when a template gets invoked because it’s the single unambiguous match. It’s also used when the match is ambiguous, causing a template list or help on a template group to be displayed; and also when the –list or –help flags are explicitly provided by the user.

### DIFF
--- a/src/Microsoft.TemplateEngine.Cli/New3Command.cs
+++ b/src/Microsoft.TemplateEngine.Cli/New3Command.cs
@@ -733,7 +733,21 @@ namespace Microsoft.TemplateEngine.Cli
             get
             {
                 IReadOnlyCollection<ITemplateMatchInfo> allTemplates = TemplateListResolver.PerformAllTemplatesQuery(_settingsLoader.UserTemplateCache.TemplateInfo, _hostDataLoader);
-                HashSet<string> allShortNames = new HashSet<string>(allTemplates.Select(x => x.Info.ShortName));
+
+                HashSet<string> allShortNames = new HashSet<string>();
+
+                foreach (ITemplateMatchInfo templateMatchInfo in allTemplates)
+                {
+                    if (templateMatchInfo.Info is IShortNameList templateWithShortNameList)
+                    {
+                        allShortNames.UnionWith(templateWithShortNameList.ShortNameList);
+                    }
+                    else
+                    {
+                        allShortNames.Add(templateMatchInfo.Info.ShortName);
+                    }
+                }
+
                 return allShortNames;
             }
         }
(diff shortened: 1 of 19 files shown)


src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -733,7 +733,21 @@ private HashSet<string> AllTemplateShortNames
             get
             {
                 IReadOnlyCollection<ITemplateMatchInfo> allTemplates = TemplateListResolver.PerformAllTemplatesQuery(_settingsLoader.UserTemplateCache.TemplateInfo, _hostDataLoader);
-                HashSet<string> allShortNames = new HashSet<string>(allTemplates.Select(x => x.Info.ShortName));
+
+                HashSet<string> allShortNames = new HashSet<string>();
```
mlorbetske: StringComparer

src/Microsoft.TemplateEngine.Edge/Settings/TemplateInfoReaders/TemplateInfoReaderVersion1_0_0_2.cs
```diff
@@ -1,6 +1,7 @@
 using Microsoft.TemplateEngine.Abstractions;
 using Microsoft.TemplateEngine.Utils;
 using Newtonsoft.Json.Linq;
+using System.Collections.Generic;
```
mlorbetske: Nit: sort

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/RunnableProjectTemplate.cs
```diff
@@ -68,7 +70,29 @@ public IDirectory TemplateSourceRoot
...
+        public string ShortName
+        {
+            get
+            {
+                if (ShortNameList.Count > 0)
+                {
+                    return ShortNameList[0];
+                }
+
+                return String.Empty;
```
mlorbetske: Nit: access via the intrinsic (string.Empty)

PR #1439, opened 2018-02-28
Add a sample line to NuGetConfig

eerhardt:
When using `dotnet new NuGetConfig`, I always have to look up the syntax for how to add a new URL.

Adding an example line to the template solves this.  Now I can just replace the name and URL with the feed I want to use.

@seancpeters @mlorbetske

### DIFF
--- a/template_feed/Microsoft.DotNet.Common.ItemTemplates/content/Nuget/nuget.config
+++ b/template_feed/Microsoft.DotNet.Common.ItemTemplates/content/Nuget/nuget.config
@@ -3,5 +3,6 @@
  <packageSources>
     <!--To inherit the global NuGet package sources remove the <clear/> line below -->
     <clear />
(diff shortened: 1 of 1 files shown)


eerhardt:
I can comment it out if you want.  The critical piece is that there is an example I can change without having to look up the syntax.

mlorbetske:
LGTM as is, with the clear being in there, the inclusion of NuGet as a feed seems to be a perfectly reasonable thing to have in by default

seancpeters:
I'm fine with leaving it in, it does seem like a reasonable default.

eerhardt:
OK, perfect.  Can someone merge this for me?  I don't have permission.

mlorbetske:
Yep, as soon as the build turns green, we'll merge. Thanks!

seancpeters:

PR #1455, opened 2018-03-06
Merge "template info reader 1.0.0.2" branch into Rel/2.1.0 preview2

seancpeters:


seancpeters:
This is the code for:
- multiple short names
- "no value default" for flags, different than regular defaults

PR #1478, opened 2018-03-17
Add conditional RootNamespace property to class library, console and test projects for 2.1

mlorbetske:
Implements the recommendation for internal bug 530795. In cases that the project name isn't a valid namespace, the namespaces generated at project creation time, but when items are added to the project, the value of RootNamespace (when not present in the project file, this is derived from the name of the project file) is not sanitized by the targets and results in non-compiling code being added to the project. This change adds a RootNamespace property to the project file only in the event that the file name is not valid as a namespace value.

For now, scoping this change to 2.1 as we'll need to update the engine in a 1.x/2.0 runtime carrying CLIs to have the supporting macro defined - if it's not there, the property with the condition on it is included in the project file. We'll also need to include a version of the engine that supports this in 15.7 cc @phenning 

@vijayrkn any concerns with this?

We'll probably want to do this for the 2.1+ web templates as well if this is merged. /cc @Eilon 

**UPDATE:** An engine update is no longer required with the usage of the internally generated symbol names - the current approach is less than ideal but will work with older CLIs/versions of VS


Eilon:
@mlorbetske thanks, I logged https://github.com/aspnet/templating/issues/371 to track for web projects.

Issue #1491, opened 2018-03-28
[Proposal] Update help to not aggregate all parameters in all templates across groups


mlorbetske:
**Problem**
Imagine a set of templates T1, T2 and T3 with the same `groupIdentity` and `shortName` "T"

|   | T1 | T2 | T3
|---|---|----|----
| Precedence | 1 | 2| 3
| --framework | netcoreapp1.0 | netcoreapp2.0 | netcoreapp3.0
| --a | yes | no | no
| --b | no | yes | no
| --c | no | no | yes

- a, b and c being optional parameters - yes and no indicating whether it's defined on that template
- framework's default assumed to be the value listed in the table

Today, calling up help on "T" will result in something like

```
--a                  Information about a
--b                  Information about b
--c                  Information about c
--framework          Information about framework
                         netcoreapp1.0
                         netcoreapp2.0
                         netcoreapp3.0
```

This is a bit confusing - all the options for all templates have been merged together & it's not clear to which template(s) `--a`, `--b` or `--c` apply or that they relate to the value of `--framework`

At invocation time:
- If no parameters are specified, you'll get `T3` since it has the highest precedence
- If you specify `--framework`, you'll get the template with the matching framework
- If you specify `--a`, `--b` or `--c`, you'll get the template for which they're defined
- If you specify a combination of `--a`, `--b` and `--c`, you'll get an error as no template defines a combination of them
- If you specify a combination of `--framework` and `--a`, `--b` or `--c`, where the additional option isn't defined on the template with the value of `--framework`, you'll also get an error

**Desired Outcome**
Display help for the most relevant specific subset of templates in the group by default, but allow opting-in to the current behavior to see the all the available data.

Ex. If the CLI in use carries the 2.0 runtime, help would ideally look like this (with better messages, of course):

```
--b                  Information about b
--framework          Information about framework
                         netcoreapp1.0 (more specific help available)
                         netcoreapp2.0
                         netcoreapp3.0 (more specific help available)

* Some information has been filtered out
  To see all help, run dotnet new T --help --all
  To see more specific help, try passing the parameter and value along with the --help switch
```

Running `dotnet new T` then should produce content corresponding to the one that information was shown about in help (in this case T2).

**Proposed Solution**
*For simplicity, it's assumed that we're talking about the single, unambiguous template group case. No multi-group scenarios should be impacted by this change*

[...]

PR #1499, opened 2018-04-04
[WIP] Support for using opinions from the host on which template is best within a group

mlorbetske:
This PR will eventually address #1491, but for now is just a place to discuss different implementation ideas.

So far, this is just a proof of concept for how the filtering could work & a little bit of the plumbing work for passing the `--show-all` flag through to help.

There are a couple different approaches that could be taken at this point:
* Expanding the data in `ITemplateMatchInfo` to capture whether a particular template is the one preferred by some intrinsic properties of the host
* Formalizing template groups (having the core query, or an adaptation on it, actually return things according to group identity, ordering the templates within it, having the ability to designate one or more as defining the help schema, etc.)

The current structure of the logic in `EnterTemplateManipulationFlow` essentially goes like this:
* If we're listing or asking for help
  * If it's a list, show a list
  * If it's help
    * If it's multiple groups, show a list
    * If it's a single group, show detailed help, aggregating parameters
* If we're not asking for help or a list
  * If it's a single group
    * Try to select a single template
      * If a single template was able to be selected, instantiate it
      * If a single template was not able to be selected, log to the authoring stream & show a list
  * If it's multiple groups
    * Show a list

There are two other usages of `CoordinateHelpAndUsageDisplay` outside of `EnterTemplateManipulationFlow` that relate to maintenance flows - this kind of limits how much extra stuff feels natural to pass in (as a bunch of extra context could be obtained up front in the manipulation flows that isn't relevant in the maintenance flow).

I wonder if we should rework things as follows:
* Perform the core query with the user's values (name matches, parameter matches, etc.)
* Establish the set of groups of templates returned
* Each group should have a notion of the "invokable match" as determined by a combination of the current logic & host intrinsics (which includes language)
* If there are multiple groups, or list is requested, show a list
* If there's a single group
  * If help is specified or there is no invokable match on the group, show help
    * If `--show-all` is specified or there is no invokable match on the group, do the current logic
    * If it's not specified, use the invokable match on the group to set the "help schema" & show help
  * Otherwise, invoke the invokable match

With this approach, help/list coordination always just takes in the following pieces of information & wouldn't need more context in particular cases
* The set of template groups
* Whether this is an explicit list operation

### DIFF
--- a/src/Microsoft.TemplateEngine.Cli/IntrinsicsFilter.cs
+++ b/src/Microsoft.TemplateEngine.Cli/IntrinsicsFilter.cs
@@ -1,0 +1,77 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.TemplateEngine.Abstractions;
+using Microsoft.TemplateEngine.Edge.Template;
+
+namespace Microsoft.TemplateEngine.Cli
+{
+    internal static class IntrinsicsFilter
+    {
+        public static bool TryFilterSingularGroupWithIntrinsics(IEngineEnvironmentSettings environmentSettings, IReadOnlyList<ITemplateMatchInfo> unambiguousTemplateGroup, out IReadOnlyList<string> specificIdentities, out IReadOnlyList<string> matchedIntrinsics)
+        {
+            //If we can't get the intrinsics keys collection, there's nothing to be done
+            if (!environmentSettings.Host.TryGetHostParamDefault("meta:intrinsics-keys", out string intrinsicsKeys))
+            {
+                matchedIntrinsics = null;
+                specificIdentities = null;
+                return false;
+            }
+
+            //Build up the intrinsics map
+            string[] keys = intrinsicsKeys.Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries);
+
+            Dictionary<string, string> intrinsics = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+
+            foreach (string key in keys)
+            {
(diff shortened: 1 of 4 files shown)


seancpeters:
I think that intrinsic matching should occur at the same time as other template matching occurs, specifically via 
`TemplateListResolver.GetTemplateResolutionResult()` which calls `TemplateListResolver.PerformCoreTemplateQuery()`

This code already collects all of the match info for each template into an ITemplateMatchInfo instance. The match info can then be checked as needed anywhere in the program without having to carry along or recheck the intrinsics. The property `ITemplateMatchInfo.DispositionOfDefaults` is already used for the default language match results, and could be used for other intrinsic match results.

src/Microsoft.TemplateEngine.Cli/IntrinsicsFilter.cs
```diff
@@ -0,0 +1,77 @@
...
+                specificIdentities = null;
+                return false;
+            }
+
+            HashSet<string> matchedTemplatesByIntrinsic = new HashSet<string>(StringComparer.Ordinal);
+            HashSet<ITemplateMatchInfo> matchedMatches = new HashSet<ITemplateMatchInfo>();
+
+            //Choice type parameters are represented as tags in the cache, see if any of the valid choices
+            //  match any of the intrinsics
+            foreach (ITemplateMatchInfo matchInfo in unambiguousTemplateGroup)
```
seancpeters: This loop will assign a template into matchedMatches if it matches on any intrinsic. But if there are multiple intrinsics, and a template only matches some of them, should it be a match?

I'm thinking it should be a match, because intrinsics are not required match criteria, but more of a fallback (like default language currently is).

PR #1521, opened 2018-04-26
Add source-build exclusion to resetting output paths.

crummel:
Source-build is using a newer CLI that has a problem with changing `BaseIntermediateOutputPath` at this point in the build.  For now, do not reset this property in source-build.


phenning:
@seancpeters Can you take a look at this and merge it?

crummel:
Thanks Sean and Phil!

PR #1534, opened 2018-05-11
Merge stabilize into master

seancpeters:
This is after release\2.1 was merged into stabilize.
After that merge, baseline comparisons between release\2.1 & stabilize were good, so proceeding with master.


seancpeters:
Baseline comparison between the merged master & release/2.1 is good.

PR #1559, opened 2018-06-12
Never trim quotes when inferring a string datatype

### LINKED ISSUE #1542: Symbol parameter default is concatenated when containing JSON escape characters
When a symbol has a `defaultValue` that contains JavaScript escape characters (e.g., `\"` for quotes), the resulting replacement will be missing a single character from the start _and_ from the end of the value in `defaultValue`

It doesn't seem to matter how many escaped characters are there, it's always the same two characters lost. For example, for _any_ of these `defaultValue` items, the resulting substitution will be truncated the same.

* `"defaultValue": "This is a sentence that needs \"quotation\" marks in it!",` (two escaped quotes)
* `"defaultValue": "This is a sentence that needs \"quotation marks in it!",` (one escaped quote)
* `"defaultValue": "This is a sentence that needs \"quotation\" \"marks in it!",` (three escaped quotes)

Regardless of the number of escaped quotes, the resulting output will be missing the first and last character. Here are the outputs of the above items, respectively.

* `his is a sentence that needs "quotation" marks in it` (missing starting `T` and trailing `!`)
* `his is a sentence that needs "quotation marks in it`
* `his is a sentence that needs "quotation" "marks in it`

I noticed similar behavior with `\n` in the default value, but I haven't tried all the other JSON escaped characters.

Attached below is the full template for reproducing the issue. Here are the basics.

### template.json excerpt

```json
…
"symbols": {
    "escapedThing": {
        "type": "parameter",
        "defaultValue": "This is a sentence that needs \"quotation\" marks in it!",
        "replaces": "escapedThingToken"
    }
}
```

### Template file

```
## Test file

Here is a test file with a symbol to replace.

escapedThingToken

That line should be replaced.
```

## Reproduction template

[template-repro.zip](https://github.com/dotnet/templating/files/2020393/template-repro.zip)

## System Info

`dotnet --version` = `2.1.200`
macOS v10.13.4

mlorbetske:
Fixes #1542

### DIFF
--- a/src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/RunnableProjectGenerator.cs
+++ b/src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/RunnableProjectGenerator.cs
@@ -660,11 +660,9 @@ namespace Microsoft.TemplateEngine.Orchestrator.RunnableProjects
                 {
                     return null;
                 }
-
-                return literal;
             }
 
-            return literal.Substring(1, literal.Length - 2);
+            return literal;
         }
 
         public ICreationEffects GetCreationEffects(IEngineEnvironmentSettings environmentSettings, ITemplate templateData, IParameterSet parameters, IComponentManager componentManager, string targetDirectory)


src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/RunnableProjectGenerator.cs
```diff
@@ -660,11 +660,9 @@ internal static object InferTypeAndConvertLiteral(string literal)
                 {
```
seancpeters: (not directly related): This check can be removed, it duplicates the check at line 638.

PR #1562, opened 2018-06-12
Display the package version in the output of -u when version information is available

### LINKED ISSUE #1548: Show version number of installed packages
With 

```
dotnet new -u
```

I can see all my installed templates but not their version. How can I see their version number?
Could you add this to -u aswell or a new parameter?

mlorbetske:
Fixes #1548

### DIFF
--- a/src/Microsoft.TemplateEngine.Cli/New3Command.cs
+++ b/src/Microsoft.TemplateEngine.Cli/New3Command.cs
@@ -11,6 +11,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.TemplateEngine.Abstractions;
 using Microsoft.TemplateEngine.Abstractions.Mount;
+using Microsoft.TemplateEngine.Abstractions.TemplateUpdates;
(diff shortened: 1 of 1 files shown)


src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -393,6 +394,14 @@ private CreationResultStatus EnterMaintenanceFlow()
                     foreach (string value in _settingsLoader.InstallUnitDescriptorCache.InstalledItems.Values)
                     {
                         Console.WriteLine($" {value}");
+
+                        if (_settingsLoader.InstallUnitDescriptorCache.Descriptors.TryGetValue(value, out IInstallUnitDescriptor descriptor))
```
seancpeters: Nit: For more compact output, do we want the version to display on the same line as the pack name?

mlorbetske: That was my first thought, but having that line be reserved for the uninstall string and an indented section below for any details about that install descriptor being below it made it more clear that it was details about that item than part of the command

PR #1567, opened 2018-06-13
Adds the UTF-8 BOM to all .fs template files

### LINKED ISSUE #1565: F# Classlibrary Template Default Encoding Isn't Unicode
When creating a new F# classlibrary (other projects/templates may have this issue as well but I didn't test it), the first library file `Library1.fs` that is created does not by default have Unicode encoding. For me with Japanese regional settings, it is encoded in Shift-JIS. This is in contrast to the default library file of a C# library, which is by default encoded with UTF-8 with signature.

Any ideas as to why this is?

More information: https://github.com/Microsoft/visualfsharp/issues/5159

mlorbetske:
Fixes #1565

@KevinRansom do you agree that adding the UTF-8 BOM (as suggested by the filer of the issue) is the correct solution here?

### DIFF
--- a/template_feed/Microsoft.DotNet.Common.ProjectTemplates.1.x/content/ClassLibrary-FSharp/Library.fs
+++ b/template_feed/Microsoft.DotNet.Common.ProjectTemplates.1.x/content/ClassLibrary-FSharp/Library.fs
@@ -1,4 +1,4 @@
-namespace Company.ClassLibrary1
+﻿namespace Company.ClassLibrary1
 
 module Say =
     let hello name =
--- a/template_feed/Microsoft.DotNet.Common.ProjectTemplates.2.0/content/ClassLibrary-FSharp/Library.fs
+++ b/template_feed/Microsoft.DotNet.Common.ProjectTemplates.2.0/content/ClassLibrary-FSharp/Library.fs
@@ -1,4 +1,4 @@
-namespace Company.ClassLibrary1
+﻿namespace Company.ClassLibrary1
 
 module Say =
     let hello name =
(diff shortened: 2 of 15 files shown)


mlorbetske:

mlorbetske:
@seancpeters the F# console templates aren't included as program.fs in each of them already has a BOM

abelbraaksma:
Just out of curiosity, why would the default not be utf8 if the bom is absent? It fixes this issue, but shouldn't visual studio default to utf8 in absence of other encoding information? (I understand that a file could be in almost any encoding, including ascii, but most encoding detection algorithms nowadays favour utf8 if other encoding info is absent, and some formats like html and xml require such approach).

The other issue might be, what if you add an ascii text file without bom and you add higher characters to it? Imo they should be encoded like utf8 too unless the file is known to be in another encoding already.

mlorbetske:
@abelbraaksma I believe that should be the behavior when the Auto-detect UTF-8 encoding without signature option is on (and it appears to be on by default). In terms of what we can influence from this project though, if we need to ensure the file is detected to be UTF-8, pretty much all we can do is add the BOM - provided that doesn't cause issues with the tools (which it doesn't seem to as Program.fs in the console templates have the BOM already). We've had some push back from the community before about having BOMs in content, but this is largely in the context of files that are likely to be concatenated where a BOM suddenly appearing in the middle of the merged file would break things, but this doesn't seem likely for these files.

abelbraaksma:
Thanks. I wasn't aware that it could be switched on or off. I know that some (Microsoft) tools don't work well with BOM and others don't work well without it. But that shouldn't stop us from doing the right thing here, though it appears that adding the BOM is the way to go ;)

PR #1569, opened 2018-06-13
Adds the ability to do a dry run of template creation with the --dry-run switch

### LINKED ISSUE #794: --dry-run option: a run through without making any changes [discussion-only]
I'd like to propose that option to make a command no-op is added to available options/switches. As the list of options grows there is a greater chance to make wrong output. I acknowledge existence of `--force` option and default no-override behavior, but `--dry-run` has usually different behavior.
This option is usually set to `false` by default has various aliases:
`--dry-run`
`-d`
`--dryRun`
Have you folks discussed such option? I know it usually has no real or little value in context of IDEs like VS, but I think it is useful for cli usage.

mlorbetske:
Adds a --dry-run switch which is currently only honored for template creation. This should be extended to support --install and --uninstall in some fashion

Fixes #794 

Support for seeing the templates that would be uninstalled as part of uninstall will come separately, support for seeing what would be installed will come with the refactor of installer


mlorbetske:
@dotnet-bot test this please

mlorbetske:
Showing the templates that would be removed as part of uninstall now happens when `-u` is specified without an argument (#1572)

PR #1572, opened 2018-06-13
Shows the list of templates included in an install descriptor when -u is passed without argument

### LINKED ISSUE #1296: improve dotnet new --uninstall
Currently dotnet new --uninstall requires knowledge about how the template was installed. For example if you install from a path you also need to uninstall providing that path. 
It would be much easier if we could uninstall by providing the **shortname**

- Its much simpler
- does not require knowledge about how the template was installed
- does not require the original config (might be moved or deleted)
- harmonizes with other commands (i.e. dotnet new <simplename>

ref #961

mlorbetske:
Fixes #1296

Example output:
```
D:\Projects>dotnet new3 -u
Template Instantiation Commands for .NET Core CLI

Currently installed items:
  D:\Projects\templating\template_feed
    Templates:
      global.json file (globaljson)
      NuGet Config (nugetconfig)
      Solution File (sln)
      Web Config (webconfig)
      Class library (classlib) C#
      Class library (classlib) F#
      Class library (classlib) VB
      Console Application (console) C#
      Console Application (console) F#
      Console Application (console) VB
      Unit Test Project (mstest) C#
      Unit Test Project (mstest) F#
      Unit Test Project (mstest) VB
      xUnit Test Project (xunit) C#
      xUnit Test Project (xunit) F#
      xUnit Test Project (xunit) VB
  Microsoft.DotNet.Web.Spa.ProjectTemplates
    Version: 2.1.0
    Templates:
      ASP.NET Core with Angular (angular) C#
      ASP.NET Core with React.js (react) C#
      ASP.NET Core with React.js and Redux (reactredux) C#
```

### DIFF
--- a/src/Microsoft.TemplateEngine.Cli/New3Command.cs
+++ b/src/Microsoft.TemplateEngine.Cli/New3Command.cs
@@ -391,15 +391,39 @@ namespace Microsoft.TemplateEngine.Cli
                     Console.WriteLine();
                     Console.WriteLine(LocalizableStrings.InstalledItems);
 
-                    foreach (string value in _settingsLoader.InstallUnitDescriptorCache.InstalledItems.Values)
+                    foreach (KeyValuePair<Guid, string> entry in _settingsLoader.InstallUnitDescriptorCache.InstalledItems)
                     {
-                        Console.WriteLine($" {value}");
+                        Console.WriteLine($"  {entry.Value}");
 
-                        if (_settingsLoader.InstallUnitDescriptorCache.Descriptors.TryGetValue(value, out IInstallUnitDescriptor descriptor))
+                        if (_settingsLoader.InstallUnitDescriptorCache.Descriptors.TryGetValue(entry.Value, out IInstallUnitDescriptor descriptor))
                         {
                             if (descriptor.Details != null && descriptor.Details.TryGetValue("Version", out string versionValue))
                             {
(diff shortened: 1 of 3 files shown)


src/Microsoft.TemplateEngine.Cli/New3Command.cs
```diff
@@ -391,15 +391,39 @@ private CreationResultStatus EnterMaintenanceFlow()
...
                             if (descriptor.Details != null && descriptor.Details.TryGetValue("Version", out string versionValue))
                             {
-                                Console.WriteLine($"    {LocalizableStrings.Version}", versionValue);
+                                Console.WriteLine($"    {LocalizableStrings.Version} {versionValue}");
+                            }
+                        }
+
+                        HashSet<string> displayStrings = new HashSet<string>(StringComparer.Ordinal);
+
+                        foreach (TemplateInfo info in _settingsLoader.UserTemplateCache.TemplateInfo.Where(x => x.ConfigMountPointId == entry.Key))
```
seancpeters: Nit: A mapping of the ConfigMountPointId's to their installed templates could be created before the outer loop, instead of iterating over all the templates for each of the install descriptors.

mlorbetske: A good point. This could save some time in cases where the template list is large.

mlorbetske:
@dotnet-bot test this please

PR #1576, opened 2018-06-14
Remove dotnet-xunit from 2.1 project templates

natemcmaster:
xunit is deprecating support for this package.

https://twitter.com/xunit/status/1001321969189437440
https://xunit.github.io/releases/2.4-beta2

cc @mlorbetske @vijayrkn do you think we should patch the 2.1.3xx SDK with this change? Also, should we drop this from 1.x and 2.0 templates too?

PR #1585, opened 2018-06-15
Remove dotnet-xunit from 2.0 and 1.x project templates

natemcmaster:
Follow-up to https://github.com/dotnet/templating/pull/1576.

xunit is deprecating support for this package. Per @bradwilson's comment, let's remove this from all project templates.

cc @mlorbetske


mlorbetske:
@natemcmaster please target the default branch (stabilize) going forward. I missed that it targeted master - I'll merge it to the correct branch now.

PR #1590, opened 2018-06-19
Return an error if no templates are found when list is specified

### LINKED ISSUE #1589: Please return an errorlevel when template not found
_From @aplocher on June 19, 2018 8:8_

## Steps to reproduce
`dotnet new some_template_that_doesnt_exist -l`

## Expected  behavior
Checking the `%ERRORLEVEL%` after searching for a template that isn't installed should result in something greater than 0 (error)

## Actual behavior
`%ERRORLEVEL%` is 0 (success) when searching for a non-existent template

## Environment data
`dotnet --info` output:

```
.NET Command Line Tools (2.1.201)

Product Information:
 Version:            2.1.201
 Commit SHA-1 hash:  7932dc6179

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.17692
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\2.1.201\

Microsoft .NET Core Shared Framework Host                                >

  Version  : 2.0.7
  Build    : 2d61d0b043915bc948ebf98836fefe9ba942be11

```

**Example Scenario:** I would like to have a basic .bat or .ps1 script that would be used by my dev team when they need to generate some code based on our templates.  It could check for existence of the template and if not found, it would install them and then generate an interactive prompt with the necessary parameters before calling "dotnet new".

Naturally, doing a search for the template (using `dotnet new <name> -l`) was the first thing I tried to do, but without a meaningful errorlevel, it's not easy to determine if it's installed.

_Copied from original issue: dotnet/cli#9513_

mlorbetske:
Move logic for determining whether there's a problem with the resolution into the resolution result, return an error code if no templates are matched instead of just printing a message

Fixes #1589

### DIFF
--- a/src/Microsoft.TemplateEngine.Cli/TemplateListResolutionResult.cs
+++ b/src/Microsoft.TemplateEngine.Cli/TemplateListResolutionResult.cs
@@ -256,5 +256,55 @@ namespace Microsoft.TemplateEngine.Cli
                 return _usingContextMatches;
             }
         }
+
+        public bool IsTemplateAmbiguous { get; private set; }
+
+        public bool IsNoTemplatesMatchedState { get; private set; }
+
+        public List<IReadOnlyList<ITemplateMatchInfo>> ContextProblemMatchGroups { get; private set; }
+
+        public List<IReadOnlyList<ITemplateMatchInfo>> RemainingPartialMatchGroups { get; private set; }
+
+        public void ComputeContextBasedAndOtherPartialMatches()
+        {
+            Dictionary<string, List<ITemplateMatchInfo>> contextProblemMatches = new Dictionary<string, List<ITemplateMatchInfo>>();
+            Dictionary<string, List<ITemplateMatchInfo>> remainingPartialMatches = new Dictionary<string, List<ITemplateMatchInfo>>();
+
+            // this filtering / grouping ignores language differences.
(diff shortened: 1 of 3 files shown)


src/Microsoft.TemplateEngine.Cli/TemplateListResolutionResult.cs
```diff
@@ -256,5 +256,55 @@ public bool UsingContextMatches
                 return _usingContextMatches;
             }
         }
+
+        public bool IsTemplateAmbiguous { get; private set; }
+
```
seancpeters: I'd consider changing IsTemplateAmbiguous and IsNoTemplatedMatchedState getters to an Ensure() type pattern, so that ComputeContextBasedAndOtherPartialMatches() doesn't need to be explicitly called to use these properties.

seancpeters: Same idea would need to apply to ContextProblemMatchGroups and RemainingPartialMatchGroups

mlorbetske: I made the method private and am now calling it as the last line of the contructor. I also fixed a null reference that was coming up when a template had a null groupIdentity in that logic

PR #1603, opened 2018-07-10
Make build the build process much faster

mlorbetske:
TODO:
- ~~Add back downloading the CLI~~
- The actual (not PR) CI agent will need to be updated to point at the new locations
- ~~Update the script on the share to account for Loc.Settings.targets moving in to the build folder~~

This also makes it possible to do `dotnet new3` at the repo root again


mlorbetske:
I think the next real step here is to merge to stabilize to see if we're getting good builds

mlorbetske:
Looks like JSON.NET versions may be getting set incorrectly, there’s at least one instance where 8.0.3 shows up in Linux builds

build/CoreBuild.proj
```diff
@@ -0,0 +1,139 @@
...
+    <PropertyGroup>
+        <RepoRoot>$(MSBuildThisFileDirectory)..\</RepoRoot>
+        <ArtifactsDir>$(RepoRoot)artifacts\</ArtifactsDir>
+        <DevPath>$(RepoRoot)dev\</DevPath>
+        <_PackageOutputPath>$(ArtifactsDir)packages\</_PackageOutputPath>
+        <TestResultsPath>$(ArtifactsDir)test_results\</TestResultsPath>
+        <TemplatesOutputPath>$(ArtifactsDir)templates\</TemplatesOutputPath>
+        <CopyBuildOutputToOutputDirectory>false</CopyBuildOutputToOutputDirectory>
+        <IncludeBuildOutput>False</IncludeBuildOutput>
+        <IncludeBuildOutput>False</IncludeBuildOutput>
```
seancpeters: Duplicate line

mlorbetske:
Running a test build on the real CI

PR #1625, opened 2018-07-22
Introduce new interfaces to preserve source to target mappings and consume them in post-actions

### LINKED ISSUE #1622: Restore packages postAction uses original primary output instead of renamed one
Given a template.json similar to this:
``` json
{
    "$schema": "http://json.schemastore.org/template",
    "author": "David Gardiner",
    "classifications": [
        "Common",
        "Console"
    ],
    "identity": "TemplateBug.CSharp",
    "name": "Template Bug Repro",
    "shortName": "template-bug-repro",
    "preferNameDirectory": true,
    "tags": {
        "language": "C#",
        "type": "project"
    },
    "primaryOutputs": [
        {
            "path": "Client.sln"
        }
    ],
    "symbols": {
        "clientPrefix": {
            "type": "parameter",
            "isRequired": true,
            "description": "Client prefix",
            "replaces": "ClientPrefixTemplateSymbol",
            "fileRename": "Client"
        }
    },
        "postActions": [
        {
            "description": "Restore NuGet packages required by this project.",
            "manualInstructions": [
                {
                    "text": "Run 'dotnet restore'"
                }
            ],
            "actionId": "210D431B-A78B-4D2F-B762-4ED3E3EA9025",
            "continueOnError": true
        }
    ]
}
```
and an existing file "Client.sln", when this template is used, it correctly renames Client.sln to the value supplied to the clientPrefix symbol, but the dotnet restore is run with "Client.sln"

```
C:\tmp\sample> dotnet new template-bug-repro --clientPrefix=Blah
The template "Template Bug Repro" was created successfully.

Processing post-creation actions...
Running 'dotnet restore' on C:\tmp\sample\Client.sln...
MSBUILD : error MSB1009: Project file does not exist.
Switch: C:\tmp\sample\Client.sln

Restore failed.
Post action failed.
Description: Restore NuGet packages required by this project.
Manual instructions: Run 'dotnet restore'
```

### LINKED ISSUE #1623: AddReferencePostActionProcessor for specific project
In a template which has multiple projects, is it possible to use AddReferencePostActionProcessor to add a reference to a package to just one specific project?

mlorbetske:
This should make referring to outputs in post-actions much easier after the rest have been adapted

Fixes #1622
Fixes #1623

On merging this, we'll need to update the documentation for the restore post-action to include the "files" parameter, the add reference post-action to include the "targetFiles" parameter and the add projects to solution post-action to include the "projectFiles" parameter, indicating that these will options are available in the 2.2+ CLI only.

This also adds an optional base class for post-actions that allows for looking up output files by their template sources (or globs)

### DIFF
--- a/src/Microsoft.TemplateEngine.Cli/PostActionProcessors/DotnetRestorePostActionProcessor.cs
+++ b/src/Microsoft.TemplateEngine.Cli/PostActionProcessors/DotnetRestorePostActionProcessor.cs
@@ -17,7 +17,7 @@ namespace Microsoft.TemplateEngine.Cli.PostActionProcessors
         {
         }
 
-        public bool Process(IEngineEnvironmentSettings environment, IPostAction actionConfig, ICreationEffects2 creationEffects, string outputBasePath)
+        public bool Process(IEngineEnvironmentSettings environment, IPostAction actionConfig, ICreationEffects2 creationEffects, ICreationResult templateCreationResult, string outputBasePath)
         {
             bool allSucceeded = true;
             IEnumerable<string> targetFiles = null;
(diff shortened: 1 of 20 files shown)


mlorbetske:
Before merging this, verify that all old behaviors are preserved & add comments about where the specific new properties are looked for to opt-in to using the creation effects instead

src/Microsoft.TemplateEngine.Cli/PostActionProcessors/DotnetRestorePostActionProcessor.cs
```diff
@@ -14,6 +17,80 @@ public DotnetRestorePostActionProcessor()
...
+            }
+            else
+            {
+                targetFiles = creationEffects.FileChanges.Select(x => x.TargetRelativePath);
+            }
+
+            if (targetFiles is null)
+            {
+                //TODO: add a message indicating that the files to restore couldn't be determined
+                return false;
```
seancpeters: Nit: TODO should be dealt with

PR #1664, opened 2018-09-21
Check on disk template timestamps

### LINKED ISSUE #1662: Parameter alias problem in templates installed from file system
I'm developing a new template locally with a parameter symbol I want to specify on the command line. I've found that if you add a parameter symbol to the `template.json` file on disk *after* installing it with `dotnet new -i <PATH>` the parameter appears in the output list, but has no aliases.

Steps to reproduce:

- Create this file:
`./parameter-issue/.template.config/template.json`
````
{
    "$schema": "http://json.schemastore.org/template",
    "author": "Mark Pim",
    "name": "ParameterAliases",
    "shortName": "paramaliases",
    "sourceName": "ParameterAliases"
  }
````
- `dotnet new -i /path/to/parameter-issue`
- Edit the template file:
````
{
    "$schema": "http://json.schemastore.org/template",
    "author": "Mark Pim",
    "name": "ParameterAliases",
    "shortName": "paramaliases",
    "sourceName": "ParameterAliases",
    "symbols": {
        "foo": {
            "type": "parameter",
            "datatype": "string"
        }
      }  
  }
````
- `dotnet new paramaliases -h`
````
Usage: new [options]
...
ParameterAliases
Author: Mark Pim
Options:

          string - Optional
````
- *(Notice no `--foo`)*
- Now `dotnet new -u /path/to/parameter-issue` `dotnet new -i /path/to/parameter-issue`
- `dotnet new paramaliases -h`
````
Usage: new [options]
...
ParameterAliases
Author: Mark Pim
Options:
  -f|--foo
            string - Optional
````

This was surprising behaviour as I'd been previously making changes to the template and source files and rerunning `dotnet new` without uninstalling and reinstalling the template and it had worked, and I'd expect the tool to either not pick up new parameters in the template, or pick them up completely (with aliases). Is this expected behaviour?

## Environment
OS: Windows 10
`dotnet --version`: `2.1.300`

### PR DESCRIPTION
**Pull request by jMarkP**
Fixes #1662 

I think this is feature complete, but I still need to test it locally. Raising this now to get some early feedback on the approach (as I'm out of time for more work on it today) in case I'm wildly out.

Summary:
- Add a `ConfigTimestampUtc` field to the `TemplateInfo`
- Write to this property in `RunnableProjectGenerator.cs`
- Add a method to get the last write time of a file path
- Refactor `SettingsLoader.RebuildCacheFromSettingsIfNotCurrent()` to first determine a list of mount points to (re)scan. This will be all mount points if the version is out of date or the user is forcing a reload, otherwise it will be any `FileSystemMountPointFactory` mount points whose templates are newer on disk than `ConfigTimestampUtc`

## Questions
- Is UTC appropriate here for the timestamp? I usually prefer it but not sure if you have any specific guidelines
- I see `Scanner.ScanMountPointForTemplatesAndLangpacks()` has logic to copy files in some situations. Could this result in the (copied) template file always getting a more recent timestamp and so always being rescanned?
- Other than debugging this locally and running `setup.sh` to run through the unit tests, are there any other things I should check before merging?

### DIFF
--- a/src/Microsoft.TemplateEngine.Edge/Settings/SettingsLoader.cs
+++ b/src/Microsoft.TemplateEngine.Edge/Settings/SettingsLoader.cs
@@ -242,12 +242,6 @@ namespace Microsoft.TemplateEngine.Edge.Settings
             ReloadTemplates();
         }
 
-        private void ReloadTemplates()
-        {
-            _templatesLoaded = false;
-            EnsureTemplatesLoaded();
-        }
-
         private IEnumerable<MountPointInfo> FindMountPointsToScan(bool forceRebuild)
         {
             bool forceScanAll = false;
@@ -302,6 +296,12 @@ namespace Microsoft.TemplateEngine.Edge.Settings
             }
         }
 
+        private void ReloadTemplates()
+        {
+            _templatesLoaded = false;
+            EnsureTemplatesLoaded();
+        }
+        
         public bool IsVersionCurrent
         {
             get
--- a/src/Microsoft.TemplateEngine.Edge/Settings/SettingsLoader.cs
+++ b/src/Microsoft.TemplateEngine.Edge/Settings/SettingsLoader.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.TemplateEngine.Abstractions;
 using Microsoft.TemplateEngine.Abstractions.Mount;
+using Microsoft.TemplateEngine.Abstractions.PhysicalFileSystem;
 using Microsoft.TemplateEngine.Edge.Mount.FileSystem;
 using Microsoft.TemplateEngine.Utils;
 using Newtonsoft.Json.Linq;
@@ -286,9 +287,15 @@ namespace Microsoft.TemplateEngine.Edge.Settings
                 }
 
                 string pathToTemplateFile = Path.Combine(mountPoint.Place, template.ConfigPlace.TrimStart('/'));
-                var timestampOnDisk = _environmentSettings.Host.FileSystem.GetLastWriteTimeUtc(pathToTemplateFile);
+
+                DateTime? timestampOnDisk = null;
+                if (_environmentSettings.Host.FileSystem is IFileLastWriteTimeSource timeSource)
+                {
+                    timestampOnDisk = timeSource.GetLastWriteTimeUtc(pathToTemplateFile);
+                }
+
                 if (!template.ConfigTimestampUtc.HasValue
-                    || template.ConfigTimestampUtc.Value < timestampOnDisk)
+                    || (timestampOnDisk.HasValue && template.ConfigTimestampUtc.Value < timestampOnDisk))
                 {
                     // Template on disk is more recent
                     yield return mountPoint;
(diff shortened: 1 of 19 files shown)

### CODE REVIEW

**Comment by mlorbetske** [comment]
@seancpeters can you take a look please?

File: src/Microsoft.TemplateEngine.Edge/Settings/SettingsLoader.cs [hunk 1]
```diff
@@ -222,56 +223,85 @@ public void RebuildCacheFromSettingsIfNotCurrent(bool forceRebuild)
...
-                localeMountPointIds.UnionWith(templatesForLocale.Select(x => x.LocaleConfigMountPointId));
+                allTemplates.UnionWith(_userTemplateCache.GetTemplatesForLocale(locale, _userSettings.Version));
             }
 
-            // Scan the unique local mount points
-            foreach (MountPointInfo mountPoint in MountPoints)
+            foreach(var template in allTemplates)
             {
-                if (localeMountPointIds.Contains(mountPoint.MountPointId) && scannedMountPoints.Add(mountPoint.MountPointId))
+                if (!_mountPoints.TryGetValue(template.ConfigMountPointId, out var mountPoint))
```
seancpeters: Nit: we don't use var in this repo, please use the explicit type.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/IRunnableProjectConfig.cs
```diff
@@ -54,5 +55,7 @@ public interface IRunnableProjectConfig
         IReadOnlyDictionary<string, IBaselineInfo> BaselineInfo { get; }
 
         bool HasScriptRunningPostActions { get; set; }
+
+        DateTime? ConfigTimestampUtc { get; set; }
```
seancpeters: I don't think this is necessary or desired. The timestamp is a template cache only concept, it won't exist in the actual template.json.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/SimpleConfigModel.cs
```diff
@@ -81,6 +81,8 @@ public SimpleConfigModel()
 
         public bool HasScriptRunningPostActions { get; set; }
 
+        public DateTime? ConfigTimestampUtc { get; set; }
```
seancpeters: Since this isn't needed in IRunnableProjectConfig, it won't be needed here. 
This class deals with reading & processing the actual tempalte.json, as opposed to the reduced data in the cache.

jMarkP:
Thanks for the detailed feedback!

I've added unit tests, some fixes and updates based on your comments. I'd still like to do some more tests locally but I think this is code complete now.

jMarkP:
I think this is now complete. I've tested it locally and it appears to work well. @seancpeters please feel free to merge if you're happy, otherwise I'm happy to make any changes you think are needed.

mlorbetske:
Thanks for doing this @jMarkP :) Hopefully we'll be able to take one more look at this in the next day or two. The only things we might want to change eventually were a few minor style things (inlining out parameters, addings braces to ifs). I'm good with how it is but I'll hold off on clicking the button until @seancpeters confirms we're good to go.

seancpeters:
@jMarkP - Thanks much for this feature! Sorry for the delay in getting it merged.

jMarkP:
No worries :) Believe me, I've waited far longer for PR merges in other projects!

PR #1711, opened 2018-11-29
Merge stabilize into master

mlorbetske:


mlorbetske:
@seancpeters @phenning after we get this build, we'll need to merge it in to VS.

cc @vijayrkn

PR #1777, opened 2019-03-06
Merge master back in to stabilize

mlorbetske:


azure-pipelines.yml
```diff
@@ -0,0 +1,136 @@
+trigger:
+  batch: true
+  branches:
+    include:
+    - master
```
joeloff: Will there be a separate definition to build this branch? Arcade and DARC will only flow dependencies for the branches that were registered

vijayrkn: We dont need a separate definition. We can add the branch to this & dependencies will continue to flow to master.

PR #1784, opened 2019-03-09
CLI Host specific data loader conversion

seancpeters:
And the start of an extensions and helper library for the new Json handling.

Allow case-insensitive key matches on the json data.

### DIFF
--- a/src/Microsoft.TemplateEngine.Cli/HostSpecificDataLoader.cs
+++ b/src/Microsoft.TemplateEngine.Cli/HostSpecificDataLoader.cs
@@ -25,10 +25,10 @@ namespace Microsoft.TemplateEngine.Cli
             try
             {
                 HostSpecificTemplateData hostData = new HostSpecificTemplateData();
-
+                
                 if (_settingsLoader.TryGetFileFromIdAndPath(templateInfo.HostConfigMountPointId, templateInfo.HostConfigPlace, out IFile file, out mountPoint))
                 {
-                    if (_jsonDomFactory.TryLoadFromFile(file, out IJsonToken root) && root is IJsonObject rootObject)
+                    if (_jsonDomFactory.TryLoadFromIFile(file, out IJsonToken root) && root is IJsonObject rootObject)
                     {
                         (string, Action<IJsonToken>)[] extractors = new (string, Action<IJsonToken>)[]
                         {
--- a/src/Microsoft.TemplateEngine.Cli/HostSpecificDataLoader.cs
+++ b/src/Microsoft.TemplateEngine.Cli/HostSpecificDataLoader.cs
@@ -1,19 +1,22 @@
-using System.IO;
+using System;
+using System.Collections.Generic;
 using Microsoft.TemplateEngine.Abstractions;
+using Microsoft.TemplateEngine.Abstractions.Json;
 using Microsoft.TemplateEngine.Abstractions.Mount;
-using Newtonsoft.Json;
-using Newtonsoft.Json.Linq;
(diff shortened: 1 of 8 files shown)


src/Microsoft.TemplateEngine.Cli/HostSpecificDataLoader.cs
```diff
@@ -1,38 +1,47 @@
...
             {
+                HostSpecificTemplateData hostData = new HostSpecificTemplateData();
+
                 if (_settingsLoader.TryGetFileFromIdAndPath(templateInfo.HostConfigMountPointId, templateInfo.HostConfigPlace, out IFile file, out mountPoint))
                 {
-                    JObject jsonData;
-                    using (Stream stream = file.OpenRead())
-                    using (TextReader textReader = new StreamReader(stream, true))
-                    using (JsonReader jsonReader = new JsonTextReader(textReader))
+                    if (_jsonDomFactory.TryLoadFromFile(file, out IJsonToken root) && root is IJsonObject rootObject)
```
mlorbetske: Given where we want to go with this, maybe sticking with extensions that take stream or IPhysicalFileSystem and zip path and path to entry or at least physical file system + path

src/Microsoft.TemplateEngine.Utils/JsonDocumentObjectModelFactoryExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+
+namespace Microsoft.TemplateEngine.Utils
+{
+    public static class JsonDocumentObjectModelFactoryExtensions
+    {
+        public static bool TryLoadFromFile(this IJsonDocumentObjectModelFactory jsonDomFactory, string pathToFile, out IJsonToken jsonToken)
+        {
+            string text;
+
+            using (Stream s = File.OpenRead(pathToFile))
```
mlorbetske: Must be from IPhysicalFileSystem for tests rather than the File class

src/Microsoft.TemplateEngine.Utils/JsonHelpers.cs
```diff
@@ -0,0 +1,45 @@
...
+                            target.Add((TType)element.Value);
+                        }
+                    }
+                }
+            );
+        }
+
+        public static (string, Action<IJsonToken>)[] CreateStringKeyDictionaryExtractor<TValue>(IJsonObject token, IDictionary<string, TValue> target)
+        {
+            List<(string, Action<IJsonToken>)> extractorList = new List<(string, Action<IJsonToken>)>();
```
mlorbetske: This could be made to be an array if we pre-collapse the set of property names & iterate over the collapsed set. Alternatively, modifying the abstraction over object could let us get away with only having to enumerate the underlying store once

src/Microsoft.TemplateEngine.Cli/HostSpecificDataLoader.cs
seancpeters: The extension method name is bad - in this method, the variable "file" is actually an IFile, so I think we're ok as-is. I'll rename it to TryLoadFromIFile() 

I'll also get rid of the file-based version of the extension method.

src/Microsoft.TemplateEngine.Utils/JsonHelpers.cs
seancpeters: Changed to constructing the extractors directly as an array.

seancpeters:
This was targeted at a predecessor of the current state of the new json parser.

PR #1794, opened 2019-03-15
Add JSON deserialization helper

mlorbetske:
Fix an issue where the Newtonsoft JSON adapter incorrectly returns a JValue instead of the underlying value when accessed through the abstraction

Add additional functionality to IJsonObject

Simplify the JsonDomFactory selection story

Plumb the ability to deal with the new JSON updates into the tests

### DIFF
--- a/src/Microsoft.TemplateEngine.Abstractions/Json/IJsonObject.cs
+++ b/src/Microsoft.TemplateEngine.Abstractions/Json/IJsonObject.cs
@@ -11,6 +11,10 @@ namespace Microsoft.TemplateEngine.Abstractions.Json
 
         IJsonObject SetValue(string propertyName, IJsonToken value);
 
-        IReadOnlyCollection<string> ExtractValues(params (string propertyName, Action<IJsonToken> valueExtractor)[] mappings);
+        ISet<string> ExtractValues(IReadOnlyDictionary<string, Action<IJsonToken>> mappings);
+
+        ISet<string> ExtractValues<T>(T context, IReadOnlyDictionary<string, Action<IJsonToken, T>> mappings);
+
+        IEnumerable<KeyValuePair<string, IJsonToken>> Properties();
     }
 }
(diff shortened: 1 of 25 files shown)


src/Microsoft.TemplateEngine.Abstractions/Json/IJsonObject.cs
```diff
@@ -11,6 +11,10 @@ public interface IJsonObject : IJsonToken
 
         IJsonObject SetValue(string propertyName, IJsonToken value);
 
-        IReadOnlyCollection<string> ExtractValues(params (string propertyName, Action<IJsonToken> valueExtractor)[] mappings);
+        ISet<string> ExtractValues(IReadOnlyDictionary<string, Action<IJsonToken>> mappings);
+
+        ISet<string> ExtractValues<T>(T context, IReadOnlyDictionary<string, Action<IJsonToken, T>> mappings);
+
+        IEnumerable<KeyValuePair<string, IJsonToken>> Properties();
```
seancpeters: Is there a reason to not have this return an IReadOnlyDictionary<string, IJsonToken> ?
It seems that the keys are necessarily unique.

src/dotnet-new3/NewtonsoftJsonDocumentObjectModel.cs
```diff
@@ -160,11 +187,24 @@ public IJsonArray RemoveAt(int index)
...
-            public object Value => Token.Value<object>();
+            public object Value
+            {
+                get
+                {
+                    object val = ((JValue)Token).Value;
+                    switch (Token.Type)
+                    {
+                        case JTokenType.Integer:
+                            return (double)(long)val;
```
seancpeters: Why cast twice?

mlorbetske: JSON.NET stores integer types as `long` so we need to unbox as the appropriate type before storing as `double`

PR #1841, opened 2019-04-08
[WIP] Include tool for creating model classes from simple definitions of expected JSON document structure

mlorbetske:


src/JsonModelGenerator/Program.cs
```diff
@@ -0,0 +1,223 @@
...
+                    if (compactifiedType.EndsWith("{}", StringComparison.Ordinal) || compactifiedType.EndsWith("[]", StringComparison.Ordinal))
+                    {
+                        usings.Add("using System.Collections.Generic;");
+                        (string property, string builderLine) = GeneratePropertyForArrayOrDictionary(className, compactifiedType, propertyName, usings);
+                        properties.Add(property);
+                        builderLines.Add(builderLine);
+                    }
+                    else
+                    {
+                        if (Regex.IsMatch(compactifiedType, @"(Guid|DateTime)(\?)?(?=[$\[\{])"))
```
seancpeters: Should this regex be anchored? I'm not sure of the full syntax of the file.

Nit: in the character class, the `[` and `{` don't need to be escaped, that part could be `[$[{]` - seems a lot more readable.

Also, might want to consider storing this regex in a static, as a pre-compiled regex. Since the program is relatively small (has few regex's), it probably stays stored in the implicit cache, so may not be relevant to pre-compile.

src/JsonModelGenerator/Program.cs
```diff
@@ -0,0 +1,223 @@
...
+        private static readonly IReadOnlyDictionary<string, string> BuiltInTypes = new Dictionary<string, string>(StringComparer.Ordinal)
+        {
+            { "int", "Int" },
+            { "string", "String" },
+            { "Guid", "Guid" },
+            { "DateTime?", "NullableDateTime" },
+            { "bool", "Bool" }
+        };
+
+        public static string GenerateCode(string sourceFilePath, string @namespace)
```
seancpeters: This method is quite long. Perhaps break it into multiple methods:
- one to read the source file into a model
- one to generate the output

mlorbetske: Extracted two methods from this to try to help make it more readable

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/Microsoft.TemplateEngine.Orchestrator.RunnableProjects.csproj
```diff
@@ -37,4 +37,20 @@
             <LastGenOutput>LocalizableStrings.Designer.cs</LastGenOutput>
         </EmbeddedResource>
     </ItemGroup>
+
+    <ItemGroup>
+      <_JsonModels Remove="Models\RunnableProjectConfig.jm" />
```
mlorbetske: Fix this

PR #1880, opened 2019-05-02
Update F# console templates to define a function

cartermp:
This is inspired by a slough of user studies we ran on getting started with F#. One of the top things people struggled with was how functions need to be declared before they are used.

Although not perfect, this has been validated to at least ease the situation for newcomers. They either:

* Immediately think to add or modify a function above `main`
* Have a function to act as a reference for when they get an error message if they define a function below `main`

### DIFF
--- a/template_feed/Microsoft.DotNet.Common.ProjectTemplates.2.1/content/ConsoleApplication-FSharp/Program.fs
+++ b/template_feed/Microsoft.DotNet.Common.ProjectTemplates.2.1/content/ConsoleApplication-FSharp/Program.fs
@@ -1,4 +1,4 @@
-// Learn more about F# at http://fsharp.org
+// Learn more about F# at http://docs.microsoft.com/dotnet/fsharp
 
 open System
 
--- a/template_feed/Microsoft.DotNet.Common.ProjectTemplates.2.1/content/ConsoleApplication-FSharp/Program.fs
+++ b/template_feed/Microsoft.DotNet.Common.ProjectTemplates.2.1/content/ConsoleApplication-FSharp/Program.fs
@@ -2,7 +2,13 @@
 
 open System
 
+// Define a function to print a message
+// Functions are defined before they are used in F#
+let printMessage message =
+    printfn "%s" message
+
 [<EntryPoint>]
 let main argv =
-    printfn "Hello World from F#!"
+    // Call the previously-defined function
+    printMessage "Hello world from F#!"
     0 // return an integer exit code
--- a/template_feed/Microsoft.DotNet.Common.ProjectTemplates.2.1/content/ConsoleApplication-FSharp/Program.fs
+++ b/template_feed/Microsoft.DotNet.Common.ProjectTemplates.2.1/content/ConsoleApplication-FSharp/Program.fs
@@ -2,7 +2,7 @@
 
 open System
 
-// Define a function to print a message
+// Define a function to construct a message to print
 let from whom =
     sprintf "from %s" whom
 
--- a/template_feed/Microsoft.DotNet.Common.ProjectTemplates.2.1/content/ConsoleApplication-FSharp/Program.fs
+++ b/template_feed/Microsoft.DotNet.Common.ProjectTemplates.2.1/content/ConsoleApplication-FSharp/Program.fs
@@ -1,14 +1,13 @@
-﻿// Learn more about F# at http://fsharp.org
+// Learn more about F# at http://fsharp.org
 
 open System
 
 // Define a function to print a message
-// Functions are defined before they are used in F#
-let printMessage message =
-    printfn "%s" message
+let from who =
+    sprintf "from %s" message
 
 [<EntryPoint>]
 let main argv =
-    // Call the previously-defined function
-    printMessage "Hello world from F#!"
+    let message = from "F#" // Call the function
+    printMessage "Hello world %s" message
     0 // return an integer exit code
--- a/template_feed/Microsoft.DotNet.Common.ProjectTemplates.2.1/content/ConsoleApplication-FSharp/Program.fs
+++ b/template_feed/Microsoft.DotNet.Common.ProjectTemplates.2.1/content/ConsoleApplication-FSharp/Program.fs
@@ -3,11 +3,11 @@
 open System
 
 // Define a function to print a message
-let from who =
-    sprintf "from %s" message
+let from whom =
+    sprintf "from %s" whom
 
 [<EntryPoint>]
 let main argv =
     let message = from "F#" // Call the function
-    printMessage "Hello world %s" message
+    printfn "Hello world %s" message
     0 // return an integer exit code
--- a/template_feed/Microsoft.DotNet.Common.ProjectTemplates.2.2/content/ConsoleApplication-FSharp/Program.fs
+++ b/template_feed/Microsoft.DotNet.Common.ProjectTemplates.2.2/content/ConsoleApplication-FSharp/Program.fs
@@ -1,4 +1,4 @@
-// Learn more about F# at http://fsharp.org
+// Learn more about F# at http://docs.microsoft.com/dotnet/fsharp
 
 open System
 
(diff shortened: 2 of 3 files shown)


cartermp:
@vijayrkn @livarcocc Assuming these validate just fine can we get them in ASAP?

cartermp:
restarting due to azure outage

template_feed/Microsoft.DotNet.Common.ProjectTemplates.3.0/content/ConsoleApplication-FSharp/Program.fs
```diff
@@ -2,7 +2,13 @@
...
+// Define a function to print a message
+// Functions are defined before they are used in F#
+let printMessage message =
+    printfn "%s" message
+
 [<EntryPoint>]
 let main argv =
-    printfn "Hello World from F#!"
+    // Call the previously-defined function
+    printMessage "Hello world from F#!"
```
enricosada: I think it's really a good idea @cartermp 

just a feedback, maybe 

```fsharp
let hello from =
     sprintf "Hello world from %s!" from

let main argv =
    let message =  hello "F#"	    // Call the previously-defined function
    printfn "%s" message
```

can do the same (show usage of functions BEFORE main), but:
- show how to create a binding
- show `sprintf` usage to create strings

maybe is just me, but `printMessage` and `printfn` seems confusing

template_feed/Microsoft.DotNet.Common.ProjectTemplates.2.2/content/ConsoleApplication-FSharp/Program.fs
```diff
@@ -2,7 +2,13 @@
 
 open System
 
+// Define a function to print a message
```
enricosada: While i think is a good idea, to show declaration of functions BEFORE the main, i i think is that is annoying because that's template is also used **A LOT** for creating new applications, not just from new users, but for everyone.

That mean, we';ll need to remove that code **everytime**.
It's already annoying the `0  // return an integer exit code` at the end. but with also that will become really fast a noise

I think can be possibile to (weighted by my personal preference :D):

- or make the compiler error REALLY REALLY more helpful, if people declare the function AFTER usage. Atm is just `The value or constructor 'add' is not defined.`  but adding `if 'add' is a function, it must be defined before the usage, so before `main`
- or just add a comment to declare the functions BEFORE the main function, instead of declare a variable.
- or add a template option to make the main empty like `dotnet new console --empty`
- or add a template option (defaulted to behaviour for newbie users), to remove that. so like having and env var `FSHARP_TEMPLATE_BEGINNER=0`

enricosada:
@cartermp added some comment.
That template is used a lot, not just from beginners, so maybe we should try to not add friction if possibile, but considering also new users.

As a note, all hello world usually are just main + `printfn "Hello World"` and that's simplicity ihmo is valued a lot, when people see a language, it's less intimidating, than a bigger template.

So maybe just a comment can be enough (low friction), if the compiler error message cannot be improved?

template_feed/Microsoft.DotNet.Common.ProjectTemplates.2.2/content/ConsoleApplication-FSharp/Program.fs
enricosada: @cartermp added https://github.com/Microsoft/visualfsharp/issues/6677 to track error message, if can be improved

template_feed/Microsoft.DotNet.Common.ProjectTemplates.3.0/content/ConsoleApplication-FSharp/Program.fs
cartermp: That would be great, I'll do something like that 👍

template_feed/Microsoft.DotNet.Common.ProjectTemplates.2.2/content/ConsoleApplication-FSharp/Program.fs
cartermp: The message improvement is good to have regardless, thanks for filing.

I'm much more keen on being "chatty" here in the templates than not. Yes, experienced F# devs would need to remove more stuff, but typically they're already removing nearly everything already. For example, the new library template also has a module and function that people typically remove.

I think the cost to pay is worth it so that people who are turned away by this don't just give up. We don't know exactly how many people try F# and fail, but after observing 9/9 people fail to grasp this concept with templates as-is, I felt something needed to be done.

template_feed/Microsoft.DotNet.Common.ProjectTemplates.2.1/content/ConsoleApplication-FSharp/Program.fs
```diff
@@ -1,14 +1,13 @@
...
+// Learn more about F# at http://fsharp.org
 
 open System
 
 // Define a function to print a message
-// Functions are defined before they are used in F#
-let printMessage message =
-    printfn "%s" message
+let from who =
+    sprintf "from %s" message
```
bartelink: s/message/who/

Also, get to a grammar checker - I strongly suspect it should actually be whom 🤓 

Also, the header comment can be improved on (not by me, but it doesn't really 'print' anything IMO)

template_feed/Microsoft.DotNet.Common.ProjectTemplates.2.1/content/ConsoleApplication-FSharp/Program.fs
```diff
@@ -1,8 +1,13 @@
-﻿// Learn more about F# at http://fsharp.org
+// Learn more about F# at http://fsharp.org
 
 open System
 
+// Define a function to print a message
```
bartelink: this does not do what the comment says

Also perhaps consider a switch that breaks the option to have such a function out into a template option ? (that's obviously a slippery slope into having a separate Tutorial template of course, and the fact the VS UI is pretty useless for templates doesnt help)

cartermp: Currently the template engine doesn't support this (or rather, it does just by having fully-fledge templates for each switch). I'd rather not go that route if possible.

bartelink: the engine supports it (i.e. the stuff where you define a conditional compilation symbol, use either #ifdefs or //#ifdefs and the engine does preprocessing)- the VS UI doesn't

but yes, that is likely a step too far - the text is better now, thanks!

vijayrkn:
@cartermp - If you still want to take this change, Can you please re-target the change to master?

cartermp:
@vijayrkn Yep! Rebased.

jzabroski:
@cartermp Should the print function have typed parameters?

cartermp:
@jzabroski Possibly. But I think it's fine to not have the type annotation for now

cartermp:
@vijayrkn I don't have merge permissions; could you or another with permissions merge this please?

vijayrkn:
done. Thanks for taking care of this 👍

cartermp:
Hey @vijayrkn these changes are in .NET Core 3.0, but don't appear to be in .NET Core 3.1 or the .NET 5 templates either. Was there something wrong that forced the templates for 3.1 and 5.0 to go back to the old ones?

vijayrkn:
These changes are in master and were made to 3.0. So this should have made it into 3.1 & 5.0. Not sure what happened. 

@phenning - any ideas? Also tagging @grinrag who is the current maintainer for this repo.

phenning:
The release/3.1 branch was created in September, and the templates added then but this change didn't get merged until November. We then manually added the 3.1 templates to master and based the 5.0 templates off of them.

It seems like the same change needs to be applied to the 3.1 templates in both the release/3.1 and master branches and for the 5.0 templates in master and other already branched but not shipped 5.0 branches.

vijayrkn:
Can we port this change to master for both 3.1 & 5.0 templates so that the next 5.0 preview gets this change?

phenning:
Sure, I can prepare the 3.1 and 5.0 changes as separate commits and PRs to master, then we can also see about porting to the 3.1 release branch for the next servicing release.

cartermp:
Thanks folks.

@vijayrkn Yes, that seems like a good approach. So long as the 3.1 templates eventually get this in (LTS 🙂 ) and the 5.0 get it in the preview cycle then I'm good!

PR #1887, opened 2019-05-07
Update web.config item template

anurse:
Fixes https://github.com/aspnet/AspNetCore/issues/4035

This updates the web.config item template to align with what `dotnet publish` produces when publishing

review @jkotalik @shirhatti

### DIFF
--- a/template_feed/Microsoft.DotNet.Common.ItemTemplates/content/WebConfig/web.config
+++ b/template_feed/Microsoft.DotNet.Common.ItemTemplates/content/WebConfig/web.config
@@ -8,7 +8,7 @@
     <system.webServer>
       <handlers>
         <remove name="aspNetCore"/>
-        <add name="aspNetCore" path="*" verb="*" modules="AspNetCoreModule" resourceType="Unspecified"/>
+        <add name="aspNetCore" path="*" verb="*" modules="AspNetCoreModuleV2" resourceType="Unspecified"/>
       </handlers>
       <aspNetCore processPath="%LAUNCHER_PATH%" arguments="%LAUNCHER_ARGS%" stdoutLogEnabled="false" stdoutLogFile=".\logs\stdout" />
     </system.webServer>
--- a/template_feed/Microsoft.DotNet.Common.ItemTemplates/content/WebConfig/web.config
+++ b/template_feed/Microsoft.DotNet.Common.ItemTemplates/content/WebConfig/web.config
@@ -4,13 +4,15 @@
   <!-- To customize the asp.net core module uncomment and edit the following section. 
   For more info see https://go.microsoft.com/fwlink/?linkid=838655 -->
   <!--
-  <system.webServer>
-    <handlers>
-      <remove name="aspNetCore"/>
-      <add name="aspNetCore" path="*" verb="*" modules="AspNetCoreModule" resourceType="Unspecified"/>
-    </handlers>
-    <aspNetCore processPath="%LAUNCHER_PATH%" arguments="%LAUNCHER_ARGS%" stdoutLogEnabled="false" stdoutLogFile=".\logs\stdout" />
-  </system.webServer>
+  <location path="." inheritInChildApplications="false">
+    <system.webServer>
+      <handlers>
+        <remove name="aspNetCore"/>
+        <add name="aspNetCore" path="*" verb="*" modules="AspNetCoreModule" resourceType="Unspecified"/>
+      </handlers>
+      <aspNetCore processPath="%LAUNCHER_PATH%" arguments="%LAUNCHER_ARGS%" stdoutLogEnabled="false" stdoutLogFile=".\logs\stdout" />
+    </system.webServer>
+  </location>
   -->
 
 </configuration>


template_feed/Microsoft.DotNet.Common.ItemTemplates/content/WebConfig/web.config
```diff
@@ -4,13 +4,15 @@
...
-    </handlers>
-    <aspNetCore processPath="%LAUNCHER_PATH%" arguments="%LAUNCHER_ARGS%" stdoutLogEnabled="false" stdoutLogFile=".\logs\stdout" />
-  </system.webServer>
+  <location path="." inheritInChildApplications="false">
+    <system.webServer>
+      <handlers>
+        <remove name="aspNetCore"/>
+        <add name="aspNetCore" path="*" verb="*" modules="AspNetCoreModule" resourceType="Unspecified"/>
+      </handlers>
+      <aspNetCore processPath="%LAUNCHER_PATH%" arguments="%LAUNCHER_ARGS%" stdoutLogEnabled="false" stdoutLogFile=".\logs\stdout" />
```
anurse: @shirhatti @jkotalik The publish output has `processPath=dotnet` and `arguments=.\appname.dll`. Should we do something similar here? The `processPath` is easy, but the `appname` is harder since the item template doesn't really know what the name of the app is.

shirhatti: Huh? Why would you want that. 

This item template in it's current incarnation works! The webSDK will substitute LAUNCHER_PATH and LAUNCHER_ARGS.

John0King:
Please consider putting 
```xml
    <!-- Enable PUT,DELETE and Other customized HTTP Verbs  -->
    <system.webServer>
      <modules>
        <remove name="WebDAVModule" />
      </modules>
    </system.webServer>
```
to the template too.

and recently, I noticed that `dotnet publish` will add the  necessary node to the `web.config`.
for example this will also work (and it will use the correct ANCM version eg `AspNetCoreModuleV2`) :
```xml
<configuration>
  
  <location path="." inheritInChildApplications="false">
    <system.webServer>
      <modules>
        <remove name="WebDAVModule" />
      </modules>
    </system.webServer>
  </location>
  
</configuration>
```

template_feed/Microsoft.DotNet.Common.ItemTemplates/content/WebConfig/web.config
anurse: Ok, that's what I wanted to clarify.

anurse:
Thoughts on the above suggestion @jkotalik @shirhatti ?

shirhatti:
@John0King Is the ask specifically to remove the `WebDAVModule`? I most definitely don't want to do that.

As you've discovered, the WebSDK will add in the ASP.NET Core elements if required. You can continue doing what what you're doing, but I don't think it's a common enough concern to warrant putting in the templates.

anurse:
Undrafting given @shirhatti 's comments. @shirhatti does this change look OK to you? All we've done is add the `<location>` element to constrain the section to this directory (which is what the websdk does).

shirhatti:
@vijayrkn Can you take a look?

template_feed/Microsoft.DotNet.Common.ItemTemplates/content/WebConfig/web.config
```diff
@@ -4,13 +4,15 @@
...
-      <remove name="aspNetCore"/>
-      <add name="aspNetCore" path="*" verb="*" modules="AspNetCoreModule" resourceType="Unspecified"/>
-    </handlers>
-    <aspNetCore processPath="%LAUNCHER_PATH%" arguments="%LAUNCHER_ARGS%" stdoutLogEnabled="false" stdoutLogFile=".\logs\stdout" />
-  </system.webServer>
+  <location path="." inheritInChildApplications="false">
+    <system.webServer>
+      <handlers>
+        <remove name="aspNetCore"/>
+        <add name="aspNetCore" path="*" verb="*" modules="AspNetCoreModule" resourceType="Unspecified"/>
```
jkotalik: Shouldn't this default to ANCMV2? Is this template used for 2.1, 2.2, etc?

John0King:
@shirhatti I'm not asking to remove the `WebDAVModule`, I'm asking to Enable `PUT`, `DELETE` Verbs to an WebAPI project in IIS, and to remove the `WebDAVModule` is the only way that I can achieve that .

shirhatti:
The default ANCM handler mapping is `path="*" verb="*"`. Unless you have a conflicting handler mapping that takes precedence, ANCM is already listening to `PUT` and `DELETE`. Unfortunately, in your case, WebDAV is the conflicting handler registered at the web server level taking precedence over your handler mapping in your site's web.config.

As I mentioned earlier, I don't think it's a common enough issue to warrant having in by default.

template_feed/Microsoft.DotNet.Common.ItemTemplates/content/WebConfig/web.config
anurse: I think it's totally fine to default to V2. What's the necessary change there?

jkotalik:
```suggestion
        <add name="aspNetCore" path="*" verb="*" modules="AspNetCoreModuleV2" resourceType="Unspecified"/>
```

Is the change needed. However, in 2.1, ANCMV2 isn't present in the hosting bundle (which may change).

anurse:
@shirhatti @jkotalik Can you give this another pass and let me know if it looks right to you?

anurse:
I mean, we want to transition even 2.1 apps to AspNetCoreModuleV2 so it seems reasonable to leave it at v2 :)

anurse:
I don't have permission to merge this. Can some contributor merge it? We (ASP.NET owners) are happy with it now :)

vijayrkn:
Adding @seancpeters who maintains this repo. @seancpeters - can you please take a look & merge if everything looks good?

vijayrkn:
@seancpeters - Does we need to cherry-pick this change to master?

seancpeters:
@vijayrkn - We'll either need to cherry-pick it, or merge stabilize into master.

PR #1966, opened 2019-06-26
Merging stabilize into master

seancpeters:


seancpeters:
@phenning - latest changes from stabilize-> master

PR #2000, opened 2019-07-10
Add feeds from Versions.props to NuGet.config

riarenas:
We need all restore sources to be located in NuGet.config since internal feeds can only be restored from there. More details including next steps [here](https://github.com/dotnet/arcade/blob/master/Documentation/RestoreSourcesUpdateStatus.md)


riarenas:
Wrong branch.

PR #2128, opened 2019-10-02
update class library templates for dotnet3.0 to use netcoreapp3.0

boyeborg:
The dotnet 3.0 templates for class libraries currently uses `netstandard2.0` as default target framework. This updates them to use `netcoreapp3.0` as default instead.


vijayrkn:
@boyeborg Thank you for the PR. We want to continue to target netstandard.20 for our class libraries by default. So we won't be able to take this PR at this time. When the class libraries are ready to target the next version of NetStandard,  we would be happy to take PRs (with the correct version) updated. 

Will be closing this PR. 

/cc @terrajobst

terrajobst:
@vijayrkn is right. This isn't a bug -- using `netstandard2.0` as the starting point is by-design :-)

PR #2143, opened 2019-11-07
Update branding to preview3

phenning:


vijayrkn:
@mmitche - Is it too late to update the branding? This is not critical if the build is already available.

vijayrkn:
+@wtgodbe

wtgodbe:
It's too late for preview3, we should just wait for the stabilization PR & include that in GA

vijayrkn:
Sounds good. We never got to update to Preview 2 any ways. So this is not a big issue.

jzabroski:
lol, "failing to follow semantic versioning is not a big deal"

vijayrkn:
@jzabroski - There were absolutely no changes made in this repo after preview1 & also templating is not a .NET Core runtime component. That is the reason I mentioned this is not impactful. 

Hope this clarifies.

jzabroski:
Ah, thank you.  Do you know if the upcoming dotnet repo merge into a mono-repo will allow a global MSBuild props file to contain this configuration, so that packages are released in a "global, build at head" fashion?  That seems like the appropriate solution.

Issue #1747, opened 2019-01-24
Support --interactive for templates behind private nuget feed


wli3:
1. For dotnet new -i. To use install a template behind private nuget feed the user cannot specify nuget feed in the command line . The user need to set the nuget feed in the global nuget config.

2. After adding a private nuget feed, if the feed need auth. The user will see the following error

```
/Users/williamli/Documents/cli/artifacts/tmp/Debug/dotnet/sdk/3.0.100-preview-009752/NuGet.targets(114,5): warning : The plugin credential provider could not acquire credentials. Authentication may require manual action. Consider re-running the command with --interactive for `dotnet`, /p:NuGetInteractive="true" for MSBuild or removing the -NonInteractive switch for `NuGet` [/Users/williamli/.templateengine/dotnetcli/v3.0.100-preview-009752/scratch/restore.csproj]
```

However, --interactive command is not available for dotnet new

To repro add my private feed to ~/.nuget/NuGet/NuGet.Config

```
<add key="testingnugetauth" value="https://williamleewul.pkgs.visualstudio.com/_packaging/testingnugetauth/nuget/v3/index.json" />
```

And run `dotnet new -i wul.DotNet.Common.ItemTemplates`

3. If the template uses a private package in another feed. The restore will fail due to no `--interactive` is passed to subsequence restore.

related to https://github.com/dotnet/templating/issues/1545 but now we have --interactive

mlorbetske:
@wli3 if no auth is needed, the `--nuget-source` allows you to specify an alternate feed to use

KathleenDollard:
@mlorbetske your answer clears up #1. 

For #2 and 3, we've done work in the CLI to improve the authentication experience for NuGet private feeds. We'd like to see this added to `dotnet new` as well. The interactive switch is needed both to access the feed with the template (#2) and the restore if it uses any packages that come from a different feed and that feed needs authentication #3

mlorbetske:
@KathleenDollard understood, I was only commenting with regard to the first item. We'll schedule the work to pass the `--interactive` switch in the internal calls to restore

maracuja-juice:
Is there any documentation on this new --interactive flag or can anyone clear up for me if this will help with #1545 ?

wli3:
Any update on this issue?

seancpeters:
No update at this point.

allysonjoses:
Any update on this issue?

donJoseLuis:
[...]

PR #2276, opened 2020-02-15
Add --interactive flag

kroymann:
Add --interactive flag which if specified is passed through to the internal dotnet restore calls. This addresses issue #1747


thecontrarycat:
It would be nice to see this PR merged.  As an Azure Artifacts user it's a real pain to try to get the `dotnet new` commands to work as it is too stupid to ignore sources that 401.

michasHL:
@vijayrkn would you mind taking a lot at this PR.
We're really waiting for this feature to get into master to finally be able to use template nuget packages from our private feed.

grinrag:
@KathleenDollard I would like to get your opinion on the UX for this new `--interactive` flag. Is is consistent across all the dotnet CLI tools? Does it do what user could expect from that parameter? 

The changes itself LGTM, and we can merge after you ack.

grinrag:
This PR is a part of the fix for the issue #179.

PR #2307, opened 2020-03-10
Cache Blob Store Downloads

brianrob:
Prior to this change, every invocation of `dotnet new` downloaded a JSON document describing the list of available templates to ensure that all clients have an up-to-date template list.  With this PR, each invocation of `dotnet new` will do the following:
 - Only download the document if it has not been downloaded within the last hour.
 - If downloading the document, use If-None-Match and ETag headers to avoid downloading the actual document unless it is different than the local copy.  This avoids a usually redundant ~1MB download.

### DIFF
--- a/src/Microsoft.TemplateSearch.Common/BlobStoreSourceFileProvider.cs
+++ b/src/Microsoft.TemplateSearch.Common/BlobStoreSourceFileProvider.cs
@@ -2,12 +2,14 @@ using System;
 using System.Net.Http;
 using System.Threading.Tasks;
 using Microsoft.TemplateEngine.Abstractions;
+using Microsoft.TemplateEngine.Abstractions.PhysicalFileSystem;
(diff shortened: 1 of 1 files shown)


brianrob:
cc: @KathleenDollard, @dsplaisted, @phenning, @donJoseLuis

brianrob:
Any thoughts on this?  Thanks.

phenning:
LGTM

brianrob:
Thanks @phenning.  Anyone else have any comments or concerns?  I don't have merge permissions, so I'll be depending on y'all to merge this.

brianrob:
Any other concerns on this one or can someone with permissions merge this?  Thanks.

dsplaisted:
Looks like I don't have permissions either.  @phenning, do you have the write permissions to merge this?

brianrob:
Thanks much @phenning!

PR #2310, opened 2020-03-12
Update .NET 5 templates with net5.0 tfm

ViktorHofer:
Contributes towards https://github.com/dotnet/sdk/issues/10756

cc @grinrag and Jose Aguilar (what's your GH handle? ;))

### DIFF
--- a/template_feed/Microsoft.DotNet.Common.ProjectTemplates.5.0/content/ClassLibrary-CSharp/.template.config/dotnetcli.host.json
+++ b/template_feed/Microsoft.DotNet.Common.ProjectTemplates.5.0/content/ClassLibrary-CSharp/.template.config/dotnetcli.host.json
@@ -19,6 +19,6 @@
     }
   },
   "usageExamples": [
-    "--framework netcoreapp3.1"
+    "--framework net5.0"
   ]
 }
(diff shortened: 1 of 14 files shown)


grinrag:
thank you for the contributions @ViktorHofer ! I will test the changes.
cc @donJoseLuis

ViktorHofer:
@grinrag can we merge this? this will require a new SDK anyway to make restore and the sdk parsing logic work.

vijayrkn:
@phenning - How would this change impact the templates displayed in VS?

phenning:
Pretty severely. we really needed time to proof VS for this change.

vijayrkn:
@ViktorHofer - This will break a lot of our VS tooling scenarios. Can we make this change only in preview3 branch and not in preview2 branch so that there is time for tooling to react to this change.

ViktorHofer:
How is the 5.0-preview2 branch connected to VS? That branch versions with dotnet. If you need to ship these templates to VS as well the I recommend creating a 16.x branch and not include my template change. That said I know that nuget ships the net5.0 compatibility feature to VS for  16.6.

phenning:
We are talking about release timing, not a hard dependency here.

ViktorHofer:
I see, thanks for clarifying.

> Can we make this change only in preview3 branch and not in preview2 branch so that there is time for tooling to react to this change.

Unfortunately not, as for 5.0-preview2 the goal is to land the net5.0 TFM change. Here's the tracking issue: https://github.com/dotnet/sdk/issues/10756.

> Pretty severely. we really needed time to proof VS for this change.

cc @leecow @SteveCarrollMSFT

PR #2318, opened 2020-03-14
Fix csharp classlib template to point to net5.0

ViktorHofer:
All other classlibs use net5.0 by default, fixing this oversight.

cc @grinrag


ViktorHofer:
Merging to unblock preview2.

PR #2341, opened 2020-03-25
Update the Last Write Time of nugetTemplateSearchInfo.json on NotModified

brianrob:
Follow on to #2307.

Download of nugetTemplateSearchInfo.json now uses ETags and If-None-Match to avoid downloading the entire payload if it matches the local copy.  The download code also avoids attempting to download more often than every hour.  The composition of these two changes resulted in a bug where after one hour we always attempt to download because the local file isn't modified when there are no changes.  Thus, the last write time won't change until the remote copy of the file is updated.

This change modifies the last write time of the local file when it is determined that the remote copy is the same as the local copy, so that no attempts are made to re-download the file for at least one hour.

### DIFF
--- a/src/Microsoft.TemplateEngine.Utils/InMemoryFileSystem.cs
+++ b/src/Microsoft.TemplateEngine.Utils/InMemoryFileSystem.cs
@@ -819,8 +819,10 @@ namespace Microsoft.TemplateEngine.Utils
             if (!IsPathInCone(file, out string processedPath))
             {
                 if (_basis is IFileLastWriteTimeSource lastWriteTimeSource)
+                {
                     return lastWriteTimeSource.GetLastWriteTimeUtc(file);
-                throw new NotImplementedException("Basis file system must implement IFileLastWriteTimeSource");
+                }
+                throw new NotImplementedException($"Basis file system must implement {nameof(IFileLastWriteTimeSource)}");
             }
 
             file = processedPath;
@@ -854,8 +856,10 @@ namespace Microsoft.TemplateEngine.Utils
             if (!IsPathInCone(file, out string processedPath))
             {
                 if (_basis is IFileLastWriteTimeSource lastWriteTimeSource)
+                {
                     lastWriteTimeSource.SetLastWriteTimeUtc(file, lastWriteTimeUtc);
-                throw new NotImplementedException("Basis file system must implement IFileLastWriteTimeSource");
+                }
+                throw new NotImplementedException($"Basis file system must implement {nameof(IFileLastWriteTimeSource)}");
             }
 
             file = processedPath;
--- a/src/Microsoft.TemplateEngine.Utils/InMemoryFileSystem.cs
+++ b/src/Microsoft.TemplateEngine.Utils/InMemoryFileSystem.cs
@@ -109,7 +109,7 @@ namespace Microsoft.TemplateEngine.Utils
             }
 
             public FileAttributes Attributes { get; set; }
-            public DateTime LastWriteTimeUtc { get; private set; }
+            public DateTime LastWriteTimeUtc { get; set; }
         }
 
         private class DisposingStream : Stream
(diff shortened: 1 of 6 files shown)


brianrob:
@phenning, PTAL.

phenning:
@donJoseLuis Can someone from your team also review and merge?

brianrob:
@donJoseLuis any concerns on this one?

donJoseLuis:
@grinrag please review & merge on 4/1.

src/Microsoft.TemplateEngine.Utils/InMemoryFileSystem.cs
```diff
@@ -848,5 +848,40 @@ public DateTime GetLastWriteTimeUtc(string file)
...
             return targetFile.LastWriteTimeUtc;
         }
+
+        public void SetLastWriteTimeUtc(string file, DateTime lastWriteTimeUtc)
+        {
+            if (!IsPathInCone(file, out string processedPath))
+            {
+                if (_basis is IFileLastWriteTimeSource lastWriteTimeSource)
+                    lastWriteTimeSource.SetLastWriteTimeUtc(file, lastWriteTimeUtc);
+                throw new NotImplementedException("Basis file system must implement IFileLastWriteTimeSource");
```
donJoseLuis: minor - if the `IFileLastWriteTimeSource` type's name changes, this message would be outdated.  Does it make sense to cache a corresponding Type name in the instance & use it as a variable here?

test/Microsoft.TemplateEngine.Mocks/MockFileSystem.cs
```diff
@@ -237,5 +237,15 @@ public DateTime GetLastWriteTimeUtc(string file)
 
             return _files[file].LastWriteTimeUtc;
         }
+
+        public void SetLastWriteTimeUtc(string file, DateTime lastWriteTimeUtc)
+        {
+            if (!FileExists(file))
```
donJoseLuis: minor: wondering why a custom file existence method is needed instead of `System.IO.File.Exists()`

test/Microsoft.TemplateEngine.Mocks/MockFileSystem.cs
```diff
@@ -237,5 +237,15 @@ public DateTime GetLastWriteTimeUtc(string file)
...
+            {
+                throw new Exception($"File {file} doesn't exist");
+            }
+
+            _files[file].LastWriteTimeUtc = lastWriteTimeUtc;
```
donJoseLuis: minor: is it possible `_files` would not contain file?  If yes, are we ok throwing here?  Probably yes, as we are throwing an exception above.

test/Microsoft.TemplateEngine.Mocks/MockFileSystem.cs
```diff
@@ -237,5 +237,15 @@ public DateTime GetLastWriteTimeUtc(string file)
 
             return _files[file].LastWriteTimeUtc;
         }
+
+        public void SetLastWriteTimeUtc(string file, DateTime lastWriteTimeUtc)
+        {
+            if (!FileExists(file))
+            {
+                throw new Exception($"File {file} doesn't exist");
```
donJoseLuis: minor: why not `FileNotFoundException` here? 
https://docs.microsoft.com/en-us/dotnet/api/system.io.filenotfoundexception?view=netcore-3.1

src/Microsoft.TemplateEngine.Utils/InMemoryFileSystem.cs
```diff
@@ -848,5 +848,40 @@ public DateTime GetLastWriteTimeUtc(string file)
 
             return targetFile.LastWriteTimeUtc;
         }
+
+        public void SetLastWriteTimeUtc(string file, DateTime lastWriteTimeUtc)
+        {
+            if (!IsPathInCone(file, out string processedPath))
+            {
+                if (_basis is IFileLastWriteTimeSource lastWriteTimeSource)
```
donJoseLuis: minor:  in other parts, one liner "if" statements ' bodies are enclosed in { }.  For homogeneity, should this if statement do the same?

donJoseLuis:
@brianrob greetings.  Approved with a handful of minor comments.

grinrag:
@brianrob could you, please, comment/fix that minor concerns from Jose? I am ok to merge this PR after this.

src/Microsoft.TemplateEngine.Utils/InMemoryFileSystem.cs
brianrob: My goal here was to maintain as much similarity to existing code as possible, so `SetLastWriteTimeUtc` was essentially copied from `GetLastWriteTimeUtc` which already had the type name as part of the string.  Rather than caching a type name, I went ahead and used `nameof` instead.

test/Microsoft.TemplateEngine.Mocks/MockFileSystem.cs
brianrob: There is a whole file system abstraction in the template engine to allow for other types beyond a physical file system.  In this case, this is a very simple in-memory file system called `MockFileSystem` so calling `System.IO.File.Exists` would not work as it would go directly to disk.

test/Microsoft.TemplateEngine.Mocks/MockFileSystem.cs
brianrob: Yes, this just preserves the behavior of all of the other APIs.

test/Microsoft.TemplateEngine.Mocks/MockFileSystem.cs
brianrob: It's reasonable to use `FileNotFoundException` here, but this just preserves the behavior of other APIs in this class.  Every API that tests for existence and throws if the file doesn't exist throws an `Exception`.

src/Microsoft.TemplateEngine.Utils/InMemoryFileSystem.cs
brianrob: Essentially the same as the others - this was copied from `GetLastWriteTimeUtc`.  I'll fix it up for both.

brianrob:
Thanks much @donJoseLuis and @grinrag.  With the latest commit, this should be ready for merge assuming the CI passes.

brianrob:
As an FYI, I don't have merge permissions, so if one of you can merge, I would appreciate it.  Thanks.

PR #2270, opened 2020-02-13
Allow data columns in grids to shrink to fit the screen buffer width

Eilon:
Updated layout logic for grid to allow a single column to be maked as `shrinkIfNeeded`. If the screen buffer is not wide enough to fit the data, the shrinkable column will be shrunk to fit the screen exactly. If the screen is so small so that there isn't even room for that, the rendering uses a minimum width, so it looks a bit ugly, but not as bad as it might otherwise look with huge wrapped text.

The only help grid that is marked as having a shrinkable column is the one that lists the installed templates.

Fixes these issues:
* dotnet new -l : Table view can be improved on. #361
* dotnet new usage results in line wrapping under default console size #1815
* The table in "dotnet new" is too big #2062

Note: This PR contains a few commits because of some code cleanup that doesn't change any functionality. The main part is the last commit that changes the grid layout functionality.

Here are examples of the new rendering:

Full width:

```
Templates                                    Short Name                Language          Tags
---------------------------------------      --------------------      ------------      --------------------
Experimental Mobile Blazor Bindings App      mobileblazorbindings      [C#]              Blazor/Xamarin.Forms
NUnit 3 Test Project                         nunit                     [C#], F#, VB      Test/NUnit
NUnit 3 Test Item                            nunit-test                [C#], F#, VB      Test/NUnit
```

A bit narrower (first item shortened):

```
Templates                                  Short Name                Language          Tags
-------------------------------------      --------------------      ------------      --------------------
Experimental Mobile Blazor Binding...      mobileblazorbindings      [C#]              Blazor/Xamarin.Forms
NUnit 3 Test Project                       nunit                     [C#], F#, VB      Test/NUnit
NUnit 3 Test Item                          nunit-test                [C#], F#, VB      Test/NUnit
```

Even narrower (all items shortened):

```
Templates             Short Name                Language          Tags
----------------      --------------------      ------------      --------------------
Experimental ...      mobileblazorbindings      [C#]              Blazor/Xamarin.Forms
NUnit 3 Test ...      nunit                     [C#], F#, VB      Test/NUnit
NUnit 3 Test ...      nunit-test                [C#], F#, VB      Test/NUnit
```

So narrow that even column header is shortened:

```
Temp...      Short Name                Language          Tags
-------      --------------------      ------------      --------------------
Expe...      mobileblazorbindings      [C#]              Blazor/Xamarin.Forms
NUni...      nunit                     [C#], F#, VB      Test/NUnit
[...]

PR #2364, opened 2020-04-15
Make truncating of template information opt-in rather than opt-out

### LINKED ISSUE #2360: `dotnet new --list` table truncated when run in subprocess
### Problem description

Some software that I am building calls `dotnet` as a subprocess (in C# land, a `Process`) without a console window. It parses the output and shows it to the user. However, with .NET 5, the template names are truncated.

To see this easily on macOS, install the .NET 5 preview SDK (preview 1+), then run `dotnet new --list | cat` in a terminal. You wouldn't normally run the output through `cat`, of course, but any other process/pipeline that gets the data is going to be truncated. You can also see an example more akin to what I am doing here: https://github.com/Deadpikle/DotnetProcessTesting (see the flag at the top of `Main`).

### Problem Cause

PR #2270 (fix of #2062) made tables look really nice on smaller console windows. It handles redirection if the output is redirected (https://github.com/dotnet/templating/blob/master/src/Microsoft.TemplateEngine.Utils/EngineEnvironmentSettings.cs#L109), but still acts as though the console is small (https://github.com/dotnet/templating/blob/master/src/Microsoft.TemplateEngine.Utils/EngineEnvironmentSettings.cs#L105).

As a console user, I really like data being easily read if I don't care about full template names! As a console user, I don't like having to re-run commands to see full template names if I do care about seeing the whole thing (resize window -> re-run command). As a developer wanting to use `dotnet` from another program, I'd rather see full template names than truncated ones.

### Possible solutions

1. Increase the [`DefaultBufferWidth`](https://github.com/dotnet/templating/blob/master/src/Microsoft.TemplateEngine.Utils/EngineEnvironmentSettings.cs#L105) to a much larger number so that redirected output gets all full template names. --- have tested this locally and it works to fix my specific use case. It is probably the easiest "quick fix" here.
2. Add a command line parameter that can be used to avoid any data truncation. (The benefit of this option is, as a user, I could always use this flag and simply make my console window bigger to see all data rather than having to re-run the command in a bigger window.)
3. Change [this line of code](https://github.com/dotnet/templating/blob/master/src/Microsoft.TemplateEngine.Cli/HelpAndUsage/HelpForTemplateResolution.cs#L201) from `.DefineColumn(t => t.Name, LocalizableStrings.Templates, shrinkIfNeeded: true)` to `.DefineColumn(t => t.Name, LocalizableStrings.Templates, shrinkIfNeeded: !Console.IsOutputRedirected)` ---- tried this locally and it didn't seem to work for some reason (data wasn't truncated when output not redirected)
4. Otherwise fix it so that redirected output is not truncated. `dotnet` doesn't know how the data will be used, so my opinion is that it's best to not truncate data so that whatever software is calling `dotnet` will have all data available.
5. ???

### Context

I previously posted about this here: https://github.com/dotnet/sdk/issues/10885 -- but the issue belonged here and not the .NET SDK repo. I apologize for posting in the wrong place.

Deadpikle:
## Description

#2360 details the problem more fully, but basically template information in a command like `dotnet new --list` is very likely to be truncated when run in a subprocess or output to a file or similar.

## Solution

Added a new `--allow-clip` flag to make this truncation opt-in rather than opt-out based on the recommendations set forth in [this issue comment for #2360](https://github.com/dotnet/templating/issues/2360#issuecomment-614099788).

If the `--allow-clip` flag is added to a `dotnet new` command, output will be truncated per the changes in #2270. If not added, no data will be clipped/truncated, no matter the size of the console.

## Notes

* The name and description is open for debate. I thought "--allow-clip" was better than something like "--truncate-data" because it uses shorter and easier English words and describes the change it will make a little bit better.
* ~~I also increased the default console buffer width to a rather arbitrary value of 1024 to make things more _likely_ to look OK if redirecting output and the `--allow-clip` parameter is set. I realize this isn't going to work for everyone since template names could be arbitrarily long.~~

## Related Issues

Closes #2360

### DIFF
--- a/src/Microsoft.TemplateEngine.Cli/CommandParsing/CommandParserSupport.cs
+++ b/src/Microsoft.TemplateEngine.Cli/CommandParsing/CommandParserSupport.cs
@@ -140,6 +140,7 @@ namespace Microsoft.TemplateEngine.Cli.CommandParsing
                                         // don't give this a default, otherwise 'new -lang' is valid and assigns the default. User should have to explicitly give the value.
                     Create.Option("--update-check", LocalizableStrings.UpdateCheckCommandHelp, Accept.NoArguments()),
                     Create.Option("--update-apply", LocalizableStrings.UpdateApplyCommandHelp, Accept.NoArguments()),
+                    Create.Option("--allow-clip", LocalizableStrings.AllowClipDescription, Accept.NoArguments()),
                 };
             }
         }
--- a/src/Microsoft.TemplateEngine.Cli/CommandParsing/INewCommandInput.cs
+++ b/src/Microsoft.TemplateEngine.Cli/CommandParsing/INewCommandInput.cs
@@ -42,6 +42,8 @@ namespace Microsoft.TemplateEngine.Cli.CommandParsing
 
         bool IsShowAllFlagSpecified { get; }
 
+        bool IsAllowClipFlagSpecified { get; }
+
         string TypeFilter { get; }
 
         string Language { get; }
--- a/src/Microsoft.TemplateEngine.Cli/CommandParsing/NewCommandInputCli.cs
+++ b/src/Microsoft.TemplateEngine.Cli/CommandParsing/NewCommandInputCli.cs
@@ -234,6 +234,8 @@ namespace Microsoft.TemplateEngine.Cli.CommandParsing
 
         public bool IsShowAllFlagSpecified => _parseResult.HasAppliedOption(new[] { _commandName, "all" });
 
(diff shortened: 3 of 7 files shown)


Deadpikle:
@grinrag I've opened #2365, which has just the change for increasing the buffer width to 160. I updated this PR to have just the one commit for adding the flag.

grinrag:
@Deadpikle sorry for the long delay. After a discussion with PM, we decided to keep it as is, because we need to have consistency between all the dotnet cli tools. I mean consistency in behavior, arguments and so on. Since increasing the value of DefaultBufferWidth to 160 helped you, I am closing this PR. Thanks.

Issue #2359, opened 2020-04-02
Proposal to Expand the TemplateEngine / CLI Interface


brianrob:
## Motivation
The primary motivation for this proposal is to eliminate unnecessary process creation overhead when communicating from the template engine back to the SDK.  Not only does this allow for reduction in additional "chaining" processes when performing actions in the SDK, but this also allows us to remove the managed overhead (JIT, R2R initialization, etc.) associated with these additional processes.

The most immediate use of this proposal would be to eliminate extra processes required to restore new templates during `DotNetRestorePostActionProcessor.Process`.

## Current State
As it stands today, there is communication between the template engine and the CLI in both directions.

**SDK --> Template Engine:** This works by allowing the SDK to pass arguments to the template engine as part of the template engine entrypoint - `New3Command.Run`.

```
public static int Run(string commandName, ITemplateEngineHost host, ITelemetryLogger telemetryLogger, Action<IEngineEnvironmentSettings, IInstaller> onFirstRun, string[] args, string hivePath)
```

**Template Engine --> SDK:** The template engine can communicate with the SDK in two ways today.  The first is by invoking delegate arguments that were passed to the template engine from the SDK when calling `New3Command.Run`.  A good example here is the `onFirstRun` delegate which implements what the SDK would like done when the template engine determines that it is being run for the first time.  The second mechanism is by allowing the template engine to invoke dotnet.exe directly which spawns a new SDK process and allows the template engine to ask the SDK to perform operations on behalf of the template engine.  An example of where this is used is in `DotNetRestorePostActionProcessor.Process`, which invokes dotnet.exe directly to call `dotnet restore` on the newly instantiated template.

## Proposal
I would like to propose that we expand on the mechanisms that the template engine has to communicate with the SDK by allowing for more delegates to be implemented by the SDK, such that the template engine no longer has to call dotnet.exe in order to ask the SDK to perform actions on its behalf.  For the most part, the extra process creation should be unnecessary, and materially increases the time required to complete the action.  This comes from both the cost of spawning the new process, as well as new managed process overhead, such as JIT and R2R initialization time.

## Strawman Code Pattern
Rather than passing individual delegates, I propose replacing the `onFirstRun` argument with a callbacks argument of type `DotNetNew3Callbacks` that lives in the template engine codebase (today the SDK depends on the template engine, not the other way around):

```
public sealed class DotNetNew3Callbacks
{
[...]

PR #2370, opened 2020-04-22
Create and Plumb New3Callbacks to Allow for Removal of Extra Processes in Dotnet CLI

brianrob:
Part of #2359.

Exposes `New3Callbacks` class that declares a set of callbacks to be implemented by the CLI.  The template engine will then invoke these callbacks when necessary, in the course of executing a call to `New3Command.Run`.

The first new addition to `New3Callbacks` allows the CLI to implement project restore without requiring the template engine to invoke a new process to call `dotnet new project/to/restore.csproj`.  Instead, the template engine will call the `RestoreProject` callback in-process.

I have tested this functionality locally and recorded these numbers:

Baseline: 8,080.724ms
Change: 6,901.032ms
Diff: -1,179.692 (-14.599%)

Note that the binaries are privates that came from the sdk repo and not from the installer repo, so these are not absolute numbers.  They are only useful as an indicator of relative change.

### DIFF
--- a/src/Microsoft.TemplateEngine.Cli/New3Callbacks.cs
+++ b/src/Microsoft.TemplateEngine.Cli/New3Callbacks.cs
@@ -1,0 +1,25 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System;
+using Microsoft.TemplateEngine.Abstractions;
+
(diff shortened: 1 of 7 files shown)


brianrob:
PTAL @donJoseLuis, @grinrag, @phenning 
cc: @KathleenDollard, @DamianEdwards, @davidfowl

phenning:
LGTM :shipit:

brianrob:
Thanks much all!

PR #2395, opened 2020-06-06
Remove unnecessary Imports in VB Console template

Youssef1313:
image


donJoseLuis:
Hello @Youssef1313 , thanks for your PR & idea.  In this case, including System is the experience we want for our customers.  Sure in your case, you compiled just the code out of the box, for most cases people add to this code immediately needing types from System.
Please continue participating & sending us PRs with your suggestions!

Issue #2426, opened 2020-07-30
Extend Hive (for .Net 5) cache with additional information about Optional Workload Templates


grinrag:
Template Engine keeps the information about all templates in the hive in the cache file. The location of cache file:
`~\.templateengine\dotnetcli\<version>\templatecache`

To track the information about templates from optional workloads, we need to extend the cache with some tag:
```json
{ "OptionalWorkload": true }
```

This tag should be added during the installation template into the Hive

ladipro:
@grinrag does this require more than just adding the new flag to the schema (to `ITemplateInfo`?) and plumbing it through the code? It looks like it should not be the case but the estimated size of **3.0** suggests that maybe more work is expected. Thanks!

PR #2435, opened 2020-08-05
Extend template cache for optional workloads

bozturkMSFT:
Delivers #2426 

Added new property IsPartOfAnOptionalWorkload to ITemplateInfo.


ladipro:
@bozturkMSFT are you missing an assignment to the new property in `TemplateCache.LocalizeTemplate()`? The entire template info is being copied there.

bozturkMSFT:
We are abandoning this PR since we have agreed to associate optional workload related data with packages instead of templates.

PR #2437, opened 2020-08-05
design proposal: IInstaller interface changes for Optional Workloads

bozturkMSFT:
When installing new templates, we need the ability to specify which template is part of an optional workload and which is not.

All overloads of `InstallPackages` method accept a list of installation requests of type `string` ( `IEnumerable<string> installationRequests`), but no way to pass which templates are optional.

Changes in this PR propose changing the signature of the methods so that it is possible to pass a new list of requests: `optionalWorkloadRequests`.

Some notes on the changes:
- Existing `installationRequests` parameter will continue to contain all the templates to be installed.
- New `optionalWorkloadRequests` parameter will be a subset of `installationRequests` and it will only contain the templates that are part of an optional workload.
- Interface has no documentation. We should add more to make things more clear. I only added some that is relevant to this proposal.

### DIFF
--- a/src/Microsoft.TemplateEngine.Cli/IInstaller.cs
+++ b/src/Microsoft.TemplateEngine.Cli/IInstaller.cs
@@ -5,6 +5,12 @@ namespace Microsoft.TemplateEngine.Cli
 {
     public interface IInstaller : IInstallerBase
     {
+        void InstallPackages(IEnumerable<string> installationRequests, IList<string> nuGetSources);
+
+        void InstallPackages(IEnumerable<string> installationRequests, IList<string> nuGetSources, bool debugAllowDevInstall);
+
+        void InstallPackages(IEnumerable<string> installationRequests, IList<string> nuGetSources, bool debugAllowDevInstall, bool interactive);
+
         /// <param name="optionalWorkloadRequests">Requests from <paramref name="installationRequests"/> which are part of an optional workload.</param>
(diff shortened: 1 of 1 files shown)


src/Microsoft.TemplateEngine.Cli/IInstaller.cs
```diff
@@ -10,5 +10,14 @@ public interface IInstaller : IInstallerBase
         void InstallPackages(IEnumerable<string> installationRequests, IList<string> nuGetSources, bool debugAllowDevInstall);
 
         void InstallPackages(IEnumerable<string> installationRequests, IList<string> nuGetSources, bool debugAllowDevInstall, bool interactive);
+
+        /// <param name="optionalWorkloadRequests">Requests from <paramref name="installationRequests"/> which are part of an optional workload.</param>
+        void InstallPackages(IEnumerable<string> installationRequests, IList<string> optionalWorkloadRequests, IList<string> nuGetSources);
```
ladipro, line 15: Wondering what you'd think about this:

```suggestion
        void InstallPackages(IEnumerable<InstallationRequest> installationRequests, IList<string> nuGetSources);
```

and

```C#
struct InstallationRequest
{
    public string FilePath; // Is it a path? It's not clear from the current interface what the string means.
    public bool IsOptionalWorkload;
}
```

Compared to the current proposal it is:
- More future proof - add more fields to `InstallationRequest` in the future, instead of adding overloads with even more parameters.
- More readable / self-explanatory - since it explicitly names the string.

grinrag:
closing this proposal as obsolete

PR #2446, opened 2020-08-12
Mark template packages as nonshipping

wtgodbe:
These are evidently transport packages, not shipping packages intended for nuget.org. Marking them as such


wtgodbe:
ping @phenning @grinrag @vijayrkn - does this change look good to you?

Issue #2436, opened 2020-08-05
Audit and fix all culture-dependent calls


AR-May:
Conditionals use the language of locale for parsing. In different locales, some types (for example double) have different formats. If an author of a template uses one locale and a consumer of the template uses another then the conditional may execute incorrectly.
```
#if (1.4 < 2.5)
// Code will not show in RUS, FRA locales
#endif
```

AR-May:
One of the unit tests catches this problem in some locales (for example RUS, FRA).  I disabled (PR #2450) this unit test in order to keep the project buildable in such locales. Once this issue is done, it would be necessary to turn it on back.

AR-May:
Issue #2073 also concerns the problem with doubles.

bozturkMSFT:
This was fixed with https://github.com/dotnet/templating/pull/2547. Closing.

PR #2450, opened 2020-08-18
Temporarily disable the failing unit test.

AR-May:
The changes disable the unit test which fails in RUS, FRA (and maybe in some other) locales.
This is to make the project buildable in all locales while we are fixing the problem.

The test should be turned on after fixing issue #2436 .

### DIFF
--- a/test/Microsoft.TemplateEngine.Core.UnitTests/ConditionalTests.CStyleEvaluator.cs
+++ b/test/Microsoft.TemplateEngine.Core.UnitTests/ConditionalTests.CStyleEvaluator.cs
@@ -1,4 +1,5 @@
 ﻿using System.Collections.Generic;
+using System.Globalization;
 using System.IO;
 using System.Text;
 using Microsoft.TemplateEngine.Core.Contracts;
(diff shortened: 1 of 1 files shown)


test/Microsoft.TemplateEngine.Core.UnitTests/ConditionalTests.CStyleEvaluator.cs
```diff
@@ -1395,7 +1395,10 @@ public void VerifyIfElseEndifConditionUsesFalse()
             Verify(Encoding.UTF8, output, changed, value, expected);
         }
 
-        [Fact(DisplayName = nameof(VerifyIfElseEndifConditionUsesDouble))]
+        [Fact(
+            DisplayName = nameof(VerifyIfElseEndifConditionUsesDouble),
+            Skip = "Temporarily disabled. The test does not pass in the RUS locale."
```
ladipro, line 1400: Please add a link to the issue tracking the disabled test.

AR-May: done

test/Microsoft.TemplateEngine.Core.UnitTests/ConditionalTests.CStyleEvaluator.cs
```diff
@@ -1,11 +1,24 @@
...
 using Xunit;
 
 namespace Microsoft.TemplateEngine.Core.UnitTests
 {
+
+    public sealed class SkipOnNonEnUsLocale : FactAttribute
+    {
+        public SkipOnNonEnUsLocale(string displayName, string message) {
+            DisplayName = displayName;
+            if (!CultureInfo.InstalledUICulture.Name.Equals("en-US"))
```
ladipro, line 15: This should be `CultureInfo.CurrentCulture`. Example: On English Windows I can set "Region" to Russia which will affect .NET formatting/parsing and expose the bug but `InstalledUICulture` will still be `en-US`.

AR-May: I see, fixed this. Thanks!

ladipro (approved):

grinrag (approved):

Issue #1641, opened 2018-08-13
closing a condition block on the last line of file fails when followed by a blank line


williamdenton:
Part of my dotnet project template has a `yml` file for which i have had to add my own `SpecialCustomOperations` configuration so comments can be detected

`template.json:`
```j son
  "SpecialCustomOperations": {
    "*.yml" : {
      "Operations": [
        {
          "type": "conditional",
          "configuration": {
            "style": "line",
            "token": "#"
          }
        }
      ]
    }
  }
```

I have found that when an `#endif` statement is on  the last line of the file followed by a blank line (as posix requires) the `dotnet new` operation fails to initialise the template completely. However if the file has no blank lines or more than one blank line the template is processed correctly.

`my-config-file.yml:`
```yml

thing:
  - thing1
  - thing2
##if( condition )
  -- thing3
##endif

```

The error i get is
```
Template "DemoProject" could not be created.
Error while processing file /my-config-file.yml
Non-negative number required.
Parameter name: srcOffset
```

williamdenton:
oops this appears to be a duplicate of https://github.com/dotnet/templating/issues/1593. Not sure how i missed that before posting this issue. I did try and search - honest!

mlorbetske:
Thanks @williamdenton! We were looking for a repro for that issue & it seems you’ve found it, I’ll close the other issue in preference to this one since this is more actionable :)

bartelink:
Can confirm this is sadly still an issue
```
dotnet --version
3.1.101
```

This is very painful as it falls foul of `.editorconfig` rules dictating [arguably!] redundant CR at EOF so it's easy to leave a template malformed

donJoseLuis:
@grinrag , let's look at a fixing this in the short-term

AR-May:
Fixed!

RehanSaeed:
Will this come out in the .NET 5 timeframe?

grinrag:
@RehanSaeed I will try my best to include this fix into .NET 5 release (5.0.100 mid of November). Also, feel free to join our Gitter channel, where you can ask questions like this (https://gitter.im/dotnet/templating).

PR #2467, opened 2020-08-24
Fix buffer behavior near EOF that results in an exception

AR-May:
Fixes issue #1641.

### DIFF
--- a/src/Microsoft.TemplateEngine.Core/Util/ProcessorState.cs
+++ b/src/Microsoft.TemplateEngine.Core/Util/ProcessorState.cs
@@ -255,7 +255,10 @@ namespace Microsoft.TemplateEngine.Core.Util
 
                 //Calculate the sequence number at the head of the buffer
                 int headSequenceNumber = CurrentSequenceNumber - CurrentBufferPosition;
-                int bufferPositionToAdvanceTo = _trie.OldestRequiredSequenceNumber - headSequenceNumber;
+
+                // Calculate the buffer position to advance to. It can not be negative.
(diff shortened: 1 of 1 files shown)


src/Microsoft.TemplateEngine.Core/Util/ProcessorState.cs
```diff
@@ -255,7 +255,10 @@ public bool Run()
 
                 //Calculate the sequence number at the head of the buffer
                 int headSequenceNumber = CurrentSequenceNumber - CurrentBufferPosition;
-                int bufferPositionToAdvanceTo = _trie.OldestRequiredSequenceNumber - headSequenceNumber;
+
+                // Calculate the buffer position to advance to. It can not be negative.
+                // Taking a maximum is a workaround for the case of async of the _trie.OldestRequiredSequenceNumber which may appear near EOF.
```
ladipro, line 260: super nit: The word async is a C# keyword referring to asynchronous processing so it looks a bit out of place in this context. Suggestion: "... a workaround for out-of-sync _trie.OldestRequiredSequenceNumber which may ...".

ladipro (approved):
The fix looks good! It's likely not addressing the root cause but understanding the parsing logic in its complex entirety may not be worth the while. If this fixes the bug and is low risk, then let's go for it. Thank you!

src/Microsoft.TemplateEngine.Core/Util/ProcessorState.cs
AR-May: Fixed, thanks for correcting the wording!

grinrag (approved):

PR #2510, opened 2020-09-10
Added CODEOWNERS file

bozturkMSFT:
This file will make sure that a newly created PR will always request reviews from the maintainers team.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,1 @@
+* @dotnet/templating-engine-maintainers


vlada-shubina (approved):

ladipro (approved):
The CODEOWNERS file in the sdk repo mentions that there's a default but it doesn't explicitly codify it. Just wanted to bring it up because it means that either the comment is wrong or there's another way to configure the default reviewer.

https://github.com/dotnet/sdk/blob/5f20e65a49cd010429b1d05eea087f157110744d/CODEOWNERS#L6

AR-May (approved):

bozturkMSFT:
> 
> 
> The CODEOWNERS file in the sdk repo mentions that there's a default but it doesn't explicitly codify it. Just wanted to bring it up because it means that either the comment is wrong or there's another way to configure the default reviewer.
> 
> https://github.com/dotnet/sdk/blob/5f20e65a49cd010429b1d05eea087f157110744d/CODEOWNERS#L6

According to this [PR](https://github.com/dotnet/sdk/pull/12829), there was a default code owner specified in the file before. They removed it because they were getting too many pings, but they didn't update the comment.

I have also found open PRs in that repo with no reviewers assigned. I believe this means that they _don't_ have a default code owner configured now, otherwise those PRs would have at least one reviewer assigned.

ladipro:
Thank you for checking!

Issue #2068, opened 2019-08-13
Restore postAction does not work when primaryOutputs renamed with sourceName


npuc85:
My template includes 3 projects:

- src/My.Project.csproj
- test/integration/My.Project.Test.Integration.csproj
- test/unit/My.Project.Test.Unit.csproj

I am utilizing the "sourceName" property of the template.json to rename these to whatever name is passed into `dotnet new`, and I want to run restore on all 3 of the projects to restore nuget packages. My template.json looks something like this:

```
{
  ...
  "shortName": "mytemplate",
  "sourceName": "My.Project",
  "preferNameDirectory": false,
  "primaryOutputs": [
    {
      "path": "src\\My.Project.csproj"
    },
    {
      "path": "test\\unit\\My.Project.Test.Unit.csproj"
    },
    {
      "path": "test\\integration\\My.Project.Test.Integration.csproj"
    }
  ],
  "postActions": [
    {
      "condition": "(!skipRestore)",
      "description": "Restore NuGet packages required by this project.",
      "manualInstructions": [
        {
          "text": "Run 'dotnet restore'"
        }
      ],
      "actionId": "210D431B-A78B-4D2F-B762-4ED3E3EA9025",
      "continueOnError": true
    }
  ],
  "symbols": {
    "skipRestore": {
      "type": "parameter",
      "datatype": "bool",
      "description": "If specified, skips the automatic restore of the project on creation.",
      "defaultValue": "false"
    }
  }
}
```

When I install the template and run `dotnet new mytemplate -n A.Project.Name` I get all 3 projects in the correct folders with the names I expect (.\src\A.Project.Name.csproj, .\test\integration\A.Project.Name.Test.Integration.csproj, .\src\A.Project.Name.Test.Unit.csproj), and it does attempt to run the restore post action for each project, but it fails because it's trying to restore the actual names that are in my primary outputs. It is trying to restore src\My.Project.csproj, which obviously does not exist because it was renamed to src\A.Project.Name.csproj

davidortinau:
I'm having the same issue. 

```
Processing post-creation actions...
Running 'dotnet restore' on CometApp3/NewApp.iOS\NewApp.iOS.csproj...
MSBUILD : error MSB1009: Project file does not exist.
Switch: CometApp3/NewApp.iOS/NewApp.iOS.csproj
```

@sayedihashimi is there a workaround for this? 

[...]

PR #2557, opened 2020-09-29
Normalize primary outputs

AR-May:
Delivers issue #2068.

Renaming of primary outputs does not work if backslashes are used. The changes normalize primary outputs paths from template.json.

### DIFF
--- a/src/Microsoft.TemplateEngine.Utils/FileSystemInfoExtensions.cs
+++ b/src/Microsoft.TemplateEngine.Utils/FileSystemInfoExtensions.cs
@@ -27,6 +27,11 @@ namespace Microsoft.TemplateEngine.Utils
             }
         }
 
+        public static string NormalizePath(this string path)
+        {
+            return path.Replace('\\', '/');
+        }
+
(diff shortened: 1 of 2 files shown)


ostorc (approved):

grinrag (approved):

src/Microsoft.TemplateEngine.Utils/FileSystemInfoExtensions.cs
```diff
@@ -1,4 +1,4 @@
-﻿using System.Collections.Generic;
+using System.Collections.Generic;
```
ladipro, line 1: nit: You are removing the UTF-8 BOM from the file. Is this an intentional change? Same for CreationPathModel.cs. Thanks!

image

AR-May: That's unintentional, I will fix this, thank you!

vlada-shubina (approved):

ladipro (approved):

Issue #2072, opened 2019-08-17
Post-action chmod fails on Mac


jasondentler:
On Mac, the chmod post-creation action fails. Just to be difficult, it also swallows the exit code, and outputs a generic error message.

This is not a duplicate of #1743 as the execute bit wasn't actually set.

### Details
I have a single post action, copied directly from the docs [here](https://github.com/dotnet/templating/wiki/Post-Action-Registry#change-file-permissions).

```
  "postActions": [{
    "condition": "(OS != \"Windows_NT\")",
    "description": "Make scripts executable",
    "manualInstructions": [ { "text": "Run 'chmod +x *.sh'" }  ],
    "actionId": "cb9a6cf3-4f5c-4860-b9d2-03a574959774",
    "args": {
        "+x": "*.sh"
    },
    "continueOnError": true
}]
```

```
› dotnet new my-app
The template "My Custom Template." was created successfully.

Processing post-creation actions...
open: invalid option -- c
Usage: open [-e] [-t] [-f] [-W] [-R] [-n] [-g] [-h] [-s <partial SDK name>][-b <bundle identifier>] [-a <application>] [filenames] [--args arguments]
Help: Open opens files from a shell.
      By default, opens each file using the default application for that file.
      If the file is in the form of a URL, the file will be opened as a URL.
Options:
      -a                Opens with the specified application.
      -b                Opens with the specified application bundle identifier.
      -e                Opens with TextEdit.
      -t                Opens with default text editor.
      -f                Reads input from standard input and opens with TextEdit.
      -F  --fresh       Launches the app fresh, that is, without restoring windows. Saved persistent state is lost, excluding Untitled documents.
      -R, --reveal      Selects in the Finder instead of opening.
      -W, --wait-apps   Blocks until the used applications are closed (even if they were already running).
          --args        All remaining arguments are passed in argv to the application's main() function instead of opened.
      -n, --new         Open a new instance of the application even if one is already running.
      -j, --hide        Launches the app hidden.
      -g, --background  Does not bring the application to the foreground.
      -h, --header      Searches header file locations for headers matching the given filenames, and opens them.
      -s                For -h, the SDK to use; if supplied, only SDKs whose names contain the argument value are searched.
                        Otherwise the highest versioned SDK in each platform is used.
Unable to apply permissions +x to "*.sh".
Post action failed.
Description: Make scripts executable
Manual instructions: Run 'chmod +x *.sh'
```

```
› ls -al *.sh
[...]

Issue #1547, opened 2018-06-01
chmod post-action broken by recent .Net Core changes


rmunn:
See https://github.com/dotnet/corefx/issues/24704#issuecomment-393750554 (and the whole issue discussion above that comment) for details, but basically, the meaning of UseShellExecute has changed. It no longer means "use `/bin/sh`"; now it means "use `xdg-open`". This results in errors like the following when the `chmod` post-action is used:

```plaintext
Processing post-creation actions...
xdg-open: unexpected argument '+x'
Try 'xdg-open --help' for more information.
Unable to apply permissions +x to "*.sh".
Post action failed.
Description: Make scripts executable
Manual instructions: Run 'chmod +x *.sh'
```

The only fix I can think of so far is to change the https://github.com/dotnet/templating/pull/1042 logic to run `/bin/sh` with arguments `$"-c \"chmod {entry.Key} {file}\""` (note that the quotation marks after the `-c` MUST be double-quotation marks so that globs like `*.sh` will be expanded by the shell; single-quotes mean "Do not expand globs" to the Unix shell, which would not be the desired behavior here),

Also see https://github.com/dotnet/templating/issues/1028#issuecomment-393738796 and comments further down for even more details.

PR #2579, opened 2020-10-14
Use UseShellExecute=false for chmod process

AR-May:
Delivers issue #2072. 
Related to PR #1555 and issue #1547.

The chmod post-action got broken because `UseShellExecute` option of `System.Diagnostics.Process.Start` changed its meaning from "run /bin/sh" to "run xdg-open" on Linux (and "run open" on Mac OS). 
PR #1555 partially fixed the problem, but it is needed also to switch the `UseShellExecute` from true to false and refuse to use this unstable option to completely fix the problem.

### DIFF
--- a/src/Microsoft.TemplateEngine.Cli/PostActionProcessors/ChmodPostActionProcessor.cs
+++ b/src/Microsoft.TemplateEngine.Cli/PostActionProcessors/ChmodPostActionProcessor.cs
@@ -43,7 +43,7 @@ namespace Microsoft.TemplateEngine.Cli.PostActionProcessors
                     {
                         RedirectStandardError = false,
                         RedirectStandardOutput = false,
-                        UseShellExecute = true,
+                        UseShellExecute = false,
                         CreateNoWindow = false,
                         WorkingDirectory = outputBasePath,
                         FileName = "/bin/sh",


donJoseLuis (approved):

src/Microsoft.TemplateEngine.Cli/PostActionProcessors/ChmodPostActionProcessor.cs
```diff
@@ -43,7 +43,7 @@ public bool Process(IEngineEnvironmentSettings environment, IPostAction actionCo
                     {
                         RedirectStandardError = false,
                         RedirectStandardOutput = false,
-                        UseShellExecute = true,
+                        UseShellExecute = false,
```
bozturkMSFT, line 46: Does this behave as expected in both windows and linux?

bozturkMSFT (approved):

src/Microsoft.TemplateEngine.Cli/PostActionProcessors/ChmodPostActionProcessor.cs
AR-May: For windows, the behavior did not change (the post-action is not intended for windows). For Linux and Mac, this fixes the problem.

grinrag (approved):

Issue #2754, opened 2020-08-03
Need to be able to specify dotnet UI language and support localized messages


KirillOsenkov:
See https://github.com/dotnet/msbuild/issues/1596#issuecomment-667665304

when I type `dotnet new -i Microsoft.PowerShell.Standard.Module.Template`, there's a first few messages that are printed in local language (in my case, Korean) saying "determining projects to restore". I've tried setting chcp 437, $env:DOTNET_CLI_UI_LANGUAGE = "en-us" and $env:VSLANG=1033 but none of them seem to work.

Reported by @Gnbrkm41

DavidKarlas:
Message about "Determining projects to restore..." is coming NuGet/MSBuild and will be fixed once original bug is fixed
But `dotnet new` indeed doesn't respect any of this, so what needs to happen is this:
https://github.com/dotnet/sdk/blob/52f079f4/src/Cli/dotnet/commands/dotnet-new/NewCommandShim.cs#L84 needs to be changed to `CultureInfo.DefaultThreadCurrentUICulture` because thats what uts set here https://github.com/dotnet/sdk/blob/52f079f4/src/Cli/dotnet/UILanguageOverride.cs#L27

DavidKarlas:
While fixing this, drop `--locale` option to align with all other `dotnet cli` tools... https://github.com/dotnet/templating/blob/c7c62e0b99d4e7574dfa1743e057791a00fd9010/src/Microsoft.TemplateEngine.Cli/New3Command.cs#L510

PR #2886, opened 2021-02-24
Fix: dotnet-new3 doesn't respect DOTNET_CLI_UI_LANGUAGE env variable

bozturkMSFT:
### Problem
Fixes https://github.com/dotnet/templating/issues/2754

### Solution
DOTNET_CLI_UI_LANGUAGE env variable is now used to decide on the UI language.
Removed --locale option.

### Checks:
- [x] Added unit tests

### DIFF
--- a/src/Microsoft.TemplateEngine.Abstractions/ITemplateEngineHost.cs
+++ b/src/Microsoft.TemplateEngine.Abstractions/ITemplateEngineHost.cs
@@ -18,8 +18,6 @@ namespace Microsoft.TemplateEngine.Abstractions
 
         void LogTiming(string label, TimeSpan duration, int depth);
 
-        string Locale { get; }
-        
         void LogMessage(string message);
 
         void OnCriticalError(string code, string message, string currentFile, long currentPosition);
@@ -36,8 +34,6 @@ namespace Microsoft.TemplateEngine.Abstractions
         void LogDiagnosticMessage(string message, string category, params string[] details);
 
         bool TryGetHostParamDefault(string paramName, out string value);
-
-        void UpdateLocale(string newLocale);
 
         void VirtualizeDirectory(string path);
 
--- a/src/dotnet-new3/Program.cs
+++ b/src/dotnet-new3/Program.cs
@@ -1,3 +1,8 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+#nullable enable
+
 using System;
 using System.Collections.Generic;
 using System.Globalization;
@@ -7,11 +12,11 @@ using System.Reflection;
 using System.Runtime.CompilerServices;
 using Microsoft.TemplateEngine.Abstractions;
 using Microsoft.TemplateEngine.Cli;
+using Microsoft.TemplateEngine.Cli.PostActionProcessors;
 using Microsoft.TemplateEngine.Edge;
+using Microsoft.TemplateEngine.Edge.TemplateUpdates;
 using Microsoft.TemplateEngine.Orchestrator.RunnableProjects;
 using Microsoft.TemplateEngine.Utils;
-using Microsoft.TemplateEngine.Edge.TemplateUpdates;
-using Microsoft.TemplateEngine.Cli.PostActionProcessors;
 using Microsoft.TemplateSearch.Common.TemplateUpdate;
 
 [assembly: InternalsVisibleTo("dotnet-new3.UnitTests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100f33a29044fa9d740c9b3213a93e57c84b472c84e0b8a0e1ae48e67a9f8f6de9d5f7f3d52ac23e48ac51801f1dc950abe901da34d2a9e3baadb141a17c77ef3c565dd5ee5054b91cf63bb3c6ab83f72ab3aafe93d0fc3c2348b764fafb0b1c0733de51459aeab46580384bf9d74c4e28164b7cde247f891ba07891c9d872ad2bb")]
@@ -23,8 +28,8 @@ namespace dotnet_new3
         private const string HostIdentifier = "dotnetcli-preview";
         private const string HostVersion = "v1.0.0";
         private const string CommandName = "new3";
-        private const string DOTNET_CLI_UI_LANGUAGE = nameof(DOTNET_CLI_UI_LANGUAGE);
-        private const string VSLANG = nameof(VSLANG);
+        private const string LanguageOverrideEnvironmentVar = "DOTNET_CLI_UI_LANGUAGE";
+        private const string VsLanguageOverrideEnvironmentVar = "VSLANG";
 
         public static int Main(string[] args)
         {
@@ -68,10 +73,14 @@ namespace dotnet_new3
 
             var builtIns = new AssemblyComponentCatalog(new[]
             {
-                typeof(RunnableProjectGenerator).GetTypeInfo().Assembly,            // for assembly: Microsoft.TemplateEngine.Orchestrator.RunnableProjects
-                typeof(NupkgInstallUnitDescriptorFactory).GetTypeInfo().Assembly,   // for assembly: Microsoft.TemplateEngine.Edge
-                typeof(DotnetRestorePostActionProcessor).GetTypeInfo().Assembly,    // for assembly: Microsoft.TemplateEngine.Cli
-                typeof(NupkgUpdater).GetTypeInfo().Assembly                         // for assembly: Microsoft.TemplateSearch.Common
+                // for assembly: Microsoft.TemplateEngine.Orchestrator.RunnableProjects
+                typeof(RunnableProjectGenerator).GetTypeInfo().Assembly,
+                // for assembly: Microsoft.TemplateEngine.Edge
+                typeof(NupkgInstallUnitDescriptorFactory).GetTypeInfo().Assembly,
+                // for assembly: Microsoft.TemplateEngine.Cli
+                typeof(DotnetRestorePostActionProcessor).GetTypeInfo().Assembly,
+                // for assembly: Microsoft.TemplateSearch.Common
+                typeof(NupkgUpdater).GetTypeInfo().Assembly
             });
 
             ConfigureLocale();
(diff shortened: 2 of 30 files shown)


DavidKarlas (approved):
I'm not thrilled about order of class members changes in New3Command.cs, since it will make discovery merging harder

src/dotnet-new3/Program.cs
```diff
@@ -14,7 +14,7 @@
 using Microsoft.TemplateEngine.Cli.PostActionProcessors;
 using Microsoft.TemplateSearch.Common.TemplateUpdate;
 
-[assembly:InternalsVisibleTo("dotnet-new3.UnitTests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100f33a29044fa9d740c9b3213a93e57c84b472c84e0b8a0e1ae48e67a9f8f6de9d5f7f3d52ac23e48ac51801f1dc950abe901da34d2a9e3baadb141a17c77ef3c565dd5ee5054b91cf63bb3c6ab83f72ab3aafe93d0fc3c2348b764fafb0b1c0733de51459aeab46580384bf9d74c4e28164b7cde247f891ba07891c9d872ad2bb")]
+[assembly: InternalsVisibleTo("dotnet-new3.UnitTests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100f33a29044fa9d740c9b3213a93e57c84b472c84e0b8a0e1ae48e67a9f8f6de9d5f7f3d52ac23e48ac51801f1dc950abe901da34d2a9e3baadb141a17c77ef3c565dd5ee5054b91cf63bb3c6ab83f72ab3aafe93d0fc3c2348b764fafb0b1c0733de51459aeab46580384bf9d74c4e28164b7cde247f891ba07891c9d872ad2bb")]
```
vlada-shubina, line 22: not your change - just wondering whether we need it

test/dotnet-new3.UnitTests/Sdk/SdkCommandSpec.cs
```diff
@@ -1,22 +1,30 @@
 ﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+#nullable enable
```
vlada-shubina, line 4: These classes were copied from dotnet/sdk - just wondering if we want to keep modifying them - it makes them diverge from original and if we need to update them from dotnet/sdk we probably have spend time to merge.

src/dotnet-new3/Program.cs
```diff
@@ -147,5 +150,44 @@ private static void FirstRun(IEngineEnvironmentSettings environmentSettings, IIn
                 paths.Copy(paths.Global.DefaultInstallTemplateSearchData, paths.User.NuGetScrapedTemplateSearchFile);
             }
         }
+
```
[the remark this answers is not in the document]
DavidKarlas: This is actually copied from SDK to get `new3` as close as possible to `new` https://github.com/dotnet/sdk/blob/52f079f4/src/Cli/dotnet/UILanguageOverride.cs#L38

src/dotnet-new3/Program.cs
```diff
@@ -147,5 +150,44 @@ private static void FirstRun(IEngineEnvironmentSettings environmentSettings, IIn
...
+        private static void ConfigureLocale()
+        {
+            CultureInfo selectedCulture = null;
+
+            // VSLANG=<lcid> is set by VS and we respect that as well so that we will respect the VS
+            // language preference if we're invoked by VS.
+            string vsLang = Environment.GetEnvironmentVariable(VSLANG);
```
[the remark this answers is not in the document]
DavidKarlas: Yes... https://github.com/dotnet/sdk/blob/52f079f4/src/Cli/dotnet/UILanguageOverride.cs#L38

src/dotnet-new3/Program.cs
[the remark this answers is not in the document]
bozturkMSFT: Like David said, I copied that part from CLI. So the behavior should be identical.

test/dotnet-new3.UnitTests/Sdk/SdkCommandSpec.cs
bozturkMSFT: Sounds like a valid concern to me. I will revert this.

src/dotnet-new3/Program.cs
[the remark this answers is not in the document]
vlada-shubina: I think we need to double check if it works as intended after this commit is inserted to SDK, or have you already checked?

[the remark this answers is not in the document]
DavidKarlas: Which commit? This change is in `dotnet-new3`, code I linked is already in their master for long time.

bozturkMSFT:
Reverted styling changes to New3Command as well as changes to files directly copied from CLI

src/dotnet-new3/Program.cs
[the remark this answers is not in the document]
vlada-shubina: But is it already working for dotnet new built in to CLI? I've tried to set `$env:DOTNET_CLI_UI_LANGUAGE = 'de-DE'` for 5.0.100, and run some commands - nothing gets translated; however I expect the restore messages gets translated (they have localization).

Probably that's due to https://github.com/dotnet/sdk/blob/a50a304d00533ff118e277112ab8cd31c60d17b6/src/Cli/dotnet/commands/dotnet-new/NewCommandShim.cs#L84 uses `CultureInfo.CurrentCulture.Name,`, dotnet new 3 uses `CultureInfo.CurrentUICulture.Name` cause the the code in https://github.com/dotnet/sdk/blob/52f079f4/src/Cli/dotnet/UILanguageOverride.cs#L38
 modifies UI language. 

I think we would need an additional PR to SDK anyway to get it worked, once this is merged and inserted we can test it to see whether it works from SDK CLI.

Also should we set env variable like SDK does, so the underlying processes are translated? 
https://github.com/dotnet/sdk/blob/52f079f4ea2c1165c12e7cda2aa8cf223c2a1cbc/src/Cli/dotnet/UILanguageOverride.cs#L21

src/dotnet-new3/Program.cs
```diff
@@ -66,13 +73,19 @@ private static DefaultTemplateEngineHost CreateHost(bool emitTimings)
...
+                // for assembly: Microsoft.TemplateEngine.Cli
+                typeof(DotnetRestorePostActionProcessor).GetTypeInfo().Assembly,
+                // for assembly: Microsoft.TemplateSearch.Common
+                typeof(NupkgUpdater).GetTypeInfo().Assembly
             });
 
-            DefaultTemplateEngineHost host = new DefaultTemplateEngineHost(HostIdentifier, HostVersion, CultureInfo.CurrentCulture.Name, preferences, builtIns, new[] { "dotnetcli" });
+            ConfigureLocale();
+
+            DefaultTemplateEngineHost host = new DefaultTemplateEngineHost(HostIdentifier, HostVersion, CultureInfo.CurrentUICulture.Name, preferences, builtIns, new[] { "dotnetcli" });
```
vlada-shubina, line 88: `CultureInfo.CurrentUICulture.Name` is set to `Locale` property of the cost (read-write property). Should we change this property to be read only and return `CultureInfo.CurrentUICulture.Name`. I'm a bit concerned that we have multiple sources for locale: `CultureInfo.CurrentUICulture.Name`, `CultureInfo.CurrentCulture.Name`  and `ITemplateEngineHost.Locale` and they might get out-of sync.

vlada-shubina: Or we can simply remove Locale property, not sure what's the benefit of it.

src/dotnet-new3/Program.cs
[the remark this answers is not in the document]
bozturkMSFT: We had already created a PR regarding `CurrentUICulture` here: https://github.com/dotnet/sdk/pull/16017 and it is now merged.

Though I'm not sure, `FlowOverrideToChildProcesses` looks like something we need here.

vlada-shubina (approved):
Not high priority: think of if we should keep ITemplateEngineHost.Locale / UpdateLocale as it is now as I see it as a possible source of mistakes for those who implements the host (remaining comment).

Example: 
- UI culture is de-DE
- host by mistake is passing en-US to host constructor 
- in this case UI will be de-DE localized, but not the template.json elements.

Also we need to check what VS is passing as locale - might be same issue as with SDK.

src/dotnet-new3/Program.cs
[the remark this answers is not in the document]
bozturkMSFT: Added `FlowOverrideToChildProcessees` (renamed to `ConfigureLocaleForChildProcesses`).

src/dotnet-new3/Program.cs
bozturkMSFT: Code seem to build without this. We will probably keep these and add even more when we are fixing the accessibility of types, but removing for the time being.

src/dotnet-new3/Program.cs
bozturkMSFT: Locale was always set to either `CultureInfo.CurrentCulture` or to "en-US" (mostly in tests). I don't see much point either. So I removed the locale from the host in the last commit. Please review with care as this maybe an impactful change.

DavidKarlas:
I wonder if we even need to pass CultureInfo around, or we can just rely on `CultureInfo.CurrentUICulture` of current thread...

bozturkMSFT:
One of the places where the locale is passed around is `TemplateCache.WriteTemplateCaches`. I think this is one of the examples of where passing it makes sense. Because we are not always using the `CultureInfo.CurrentUICulture` here and the culture depends on the languages supported by the localized templates.

https://github.com/dotnet/templating/blob/master/src/Microsoft.TemplateEngine.Edge/Settings/TemplateCache.cs#L181

src/Microsoft.TemplateEngine.Abstractions/ITemplateEngineHost.cs
```diff
@@ -18,8 +18,6 @@ public interface ITemplateEngineHost
 
         void LogTiming(string label, TimeSpan duration, int depth);
 
-        string Locale { get; }
```
vlada-shubina, line 21: just a thought: should we rather keep the property but rather force it to be CultureInfo.CurrentUICulture.Name and UpdateLocale to update UICulture when changed in our implementations of hosts?

random idea: some author host might want to implement Locale differently from CultureInfo.CurrentUICulture and this way they will still be able to do it. 
Also I'm not fan of Edge calling CultureInfo for file naming determining.

bozturkMSFT: Changing the `CurrentUICulture` inside `UpdateLocale` might have undesired effects. The caller will think that it is changing the locale of the host, but it is changing it for the whole thread. Someone may create 2 different hosts and give them different locales, which again won't work.

For your random idea: Localized resources in .resx will still use CurrentUICulture by default. If someone overrides the default locale, resource managers will need to be updated as well. For me, it sounds simple to have everything get locale from CurrentUICulture.

What part of the code are you talking about with "Edge calling CultureInfo"? I'm open to suggestions.

PR #2891, opened 2021-03-01
Update branding to 5.0.5

bozturkMSFT:


DavidKarlas (approved):
This feels a bit weird since we use runtime versioning but follow sdk, I think it should be fine.

donJoseLuis (approved):

Issue #2217, opened 2020-01-02
File encoding is not maintained when first line of file is a conditional statement


ajaybhargavb:
The file generated after doing `dotnet new` has a different encoding than what was in the original file for the template.

Repro:
- Install dotnet 3.1 sdk
- Unzip `C:\Program Files\dotnet\templates\3.1.0\microsoft.dotnet.web.projecttemplates.3.1.3.1.0.nupkg` and go to `Web.ProjectTemplates/content/BlazorServerWeb-Csharp/`
- Verify the encoding of `App.razor` (It should be UTF-8 with BOM)
- Now create a new app with `dotnet new blazor`
- Open the generated `App.razor` and check its encoding

Expected:
UTF-8 with BOM

Actual:
Windows Codepage 1252

cc @phenning

phenning:
This looks like a bug when the first line of the processed file contains a conditional construct such as #if.

I just verified that either adding a blank line to the top of the file, or moving the #if to the next line as follows results in the encoding being preserved.

@*
#if (NoAuth)
<Router AppAssembly="@typeof(Program).Assembly">

@ajaybhargavb You could use this as a workaround until the root cause of underlying bug is found and fixed.

ajaybhargavb:
@phenning, 
```
@*
#if (NoAuth)
<Router AppAssembly="@typeof(Program).Assembly">
```
doesn't work.

Adding a newline before the @* works but it ends up adding a blank line at the top of the generated file.

phenning:
You are right, I wasn't looking at the generated file correctly with that version of the workaround.

ajaybhargavb:
Is there a workaround that wouldn't add a newline at the top of the generated file?

mkArtakMSFT:
@phenning I'm concerned regarding taking a workaround for this in a template as that would be a per-file fix, and we can potentially run into this issue again in the future.
What about treating /forcing all the files to be UTF8? This should be very small change in the templating engine.

phenning:
That seems rather heavy handed, and is not the right fix either. 

Not all files are UTF-8. The right fix should be to fix the fact that the BOM gets stripped when the first line contains a conditional statement. The conditional itself gets stripped, so there is likely some sort of issue with that code when that line happens to be the first line in the file after the BOM. 

AFAIK, this is the only file in any of the .NET Core templates which have a conditional on the very first line of the file.

mkArtakMSFT:
[...]

PR #2939, opened 2021-03-16
Fix 2913: Using //-:cnd:noEmit on the first line of *.cs file corrupts file content when templating new project

DavidKarlas:
### Problem
Issue was that BOM was emitted only in some code path...
Which means that if "operation" was on first character, BOM was emitted after that operation, hence corrupted file...

### Solution
Now I moved BomEmitting outside loop to beginning of Run method... Even if output doesn't have any content, arguably, it still makes sense to emit BOM...

### Checks:
- [x] Added unit tests

### DIFF
--- a/src/Microsoft.TemplateEngine.Core/Util/ProcessorState.cs
+++ b/src/Microsoft.TemplateEngine.Core/Util/ProcessorState.cs
@@ -18,10 +18,8 @@ namespace Microsoft.TemplateEngine.Core.Util
         private readonly Stream _target;
         private readonly TrieEvaluator<OperationTerminal> _trie;
         private readonly int _flushThreshold;
+        private readonly int _bomSize;
         private Stream _source;
-        private byte[] _bom;
-        private bool _isBomNeeded;
-        private Encoding _encoding;
 
         public ProcessorState(Stream source, Stream target, int bufferSize, int flushThreshold, IEngineConfig config, IReadOnlyList<IOperationProvider> operationProviders)
         {
@@ -53,14 +51,12 @@ namespace Microsoft.TemplateEngine.Core.Util
             CurrentBuffer = new byte[bufferSize];
             CurrentBufferLength = source.Read(CurrentBuffer, 0, CurrentBuffer.Length);
 
-            Encoding encoding = EncodingUtil.Detect(CurrentBuffer, CurrentBufferLength, out _bom);
-            Encoding = encoding;
-            CurrentBufferPosition = _bom.Length;
-            CurrentSequenceNumber = _bom.Length;
-            if (_bom.Length > 0)
-            {
-                _isBomNeeded = true;
-            }
+            Encoding encoding = EncodingUtil.Detect(CurrentBuffer, CurrentBufferLength, out byte[] bom);
+            EncodingConfig = new EncodingConfig(Config, encoding);
+            _bomSize = bom.Length;
+            CurrentBufferPosition = _bomSize;
+            CurrentSequenceNumber = _bomSize;
+            target.Write(bom, 0, _bomSize);
 
             bool explicitOnConfigurationRequired = false;
             Dictionary<Encoding, Trie<OperationTerminal>> byEncoding = TrieLookup.GetOrAdd(operationProviders, x => new Dictionary<Encoding, Trie<OperationTerminal>>());
@@ -111,7 +107,7 @@ namespace Microsoft.TemplateEngine.Core.Util
                 Buffer.BlockCopy(CurrentBuffer, CurrentBufferPosition, tmp, 0, CurrentBufferLength - CurrentBufferPosition);
                 int nRead = _source.Read(tmp, CurrentBufferLength - CurrentBufferPosition, tmp.Length - CurrentBufferLength);
                 CurrentBuffer = tmp;
-                CurrentBufferLength += nRead - _bom.Length;
+                CurrentBufferLength += nRead - _bomSize;
                 CurrentBufferPosition = 0;
                 CurrentSequenceNumber = 0;
             }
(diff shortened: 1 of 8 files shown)


src/Microsoft.TemplateEngine.Core/Util/ProcessorState.cs
```diff
@@ -75,7 +75,8 @@ public ProcessorState(Stream source, Stream target, int bufferSize, int flushThr
             CurrentBufferLength = source.Read(CurrentBuffer, 0, CurrentBuffer.Length);
 
             Encoding encoding = EncodingUtil.Detect(CurrentBuffer, CurrentBufferLength, out _bom);
-            Encoding = encoding;
+            _encoding = encoding;
```
vlada-shubina, line 78: not sure of this change - why it was needed?

test/Microsoft.TemplateEngine.Core.UnitTests/SetFlagTests.cs
```diff
@@ -147,5 +148,41 @@ public void ValidateDontEmitFlagDoesntAddNewline()
             IProcessor processor = Processor.Create(engineConfig, operations);
             RunAndVerify(originalValue, expectedValue, processor, 9999);
         }
+
+        // Using //-:cnd:noEmit on the first line of *.cs file corrupts file content when templating new project
+        [Fact(DisplayName = nameof(Bug2913))]
+        public void Bug2913()
```
vlada-shubina, line 154: should we use better name? :) e.g. `ValidateConditionOnStartOfFile'`

src/Microsoft.TemplateEngine.Core/Util/ProcessorState.cs
DavidKarlas: It's to make compiler happy about #nullable enable so this 2 fields are set in constructor, wondering what would be better way to do this...

test/Microsoft.TemplateEngine.Core.UnitTests/SetFlagTests.cs
DavidKarlas: I will rename

src/Microsoft.TemplateEngine.Core/Util/ProcessorState.cs
DavidKarlas: Looks like there is much better solution, but only in C#9 :(
Where you can put attribute on method that is called from constructor and it will mark field as set...

image

Here is sharplab gist you can play with feature: https://sharplab.io/#gist:17b51a72b4f952b7ef06566c7e5208f1

What would you prefer to do now?

vlada-shubina: I see that setter is not used at all after this change. Should we remove it and make both properties read-only?

vlada-shubina (approved):

src/Microsoft.TemplateEngine.Core/Util/ProcessorState.cs
```diff
@@ -208,11 +208,6 @@ public bool Run()
 //Console.WriteLine("UnmatchedBlock");
 //string text = System.Text.Encoding.UTF8.GetString(CurrentBuffer, handoffBufferPosition - toWrite - matchLength, toWrite).Replace("\0", "\\0");
 //Console.WriteLine(text);
-                            if (_isBomNeeded)
```
bozturkMSFT, line 211: I suggest we get rid of the commented code in the file.

bozturkMSFT (approved):
LGTM, as long as we are not breaking https://github.com/dotnet/templating/issues/2217.

src/Microsoft.TemplateEngine.Core/Util/ProcessorState.cs
```diff
@@ -127,18 +123,9 @@ public ProcessorState(Stream source, Stream target, int bufferSize, int flushThr
...
-        {
-            get { return _encoding; }
-
-            set
-            {
-                _encoding = value;
-                EncodingConfig = new EncodingConfig(Config, _encoding);
-            }
-        }
+        public Encoding Encoding { get => EncodingConfig.Encoding; }
```
donJoseLuis, line 128: public Encoding Encoding => EncodingConfig.Encoding;

src/Microsoft.TemplateEngine.Core/Util/ProcessorState.cs
```diff
@@ -53,14 +51,12 @@ public ProcessorState(Stream source, Stream target, int bufferSize, int flushThr
...
 
-            Encoding encoding = EncodingUtil.Detect(CurrentBuffer, CurrentBufferLength, out _bom);
-            Encoding = encoding;
-            CurrentBufferPosition = _bom.Length;
-            CurrentSequenceNumber = _bom.Length;
-            if (_bom.Length > 0)
-            {
-                _isBomNeeded = true;
-            }
+            Encoding encoding = EncodingUtil.Detect(CurrentBuffer, CurrentBufferLength, out byte[] bom);
```
donJoseLuis, line 54: it's old code, given.  Should the .ctor key parameters be validated?

test/Microsoft.TemplateEngine.Core.UnitTests/SetFlagTests.cs
```diff
@@ -159,5 +159,43 @@ public void ValidateDontEmitFlagDoesntAddNewline()
...
+using System;
+#endif
+//+:cnd:noEmit";
+
+            string expectedValue = @"#if DEBUG
+using System;
+#endif
+";
+
+            VariableCollection vc = new VariableCollection();
```
donJoseLuis, line 179: `VariableCollection vc = new ();`

New changes should adopt this pattern.

src/Microsoft.TemplateEngine.Core/Util/ProcessorState.cs
```diff
@@ -53,14 +51,12 @@ public ProcessorState(Stream source, Stream target, int bufferSize, int flushThr
...
+            EncodingConfig = new EncodingConfig(Config, encoding);
```
donJoseLuis, line 55: I would consider a design pattern that would abstracts away from callers the dealing with `stream` interactions.  At a minimum, such overloads, as done in standard I\O .NET types.

donJoseLuis (approved):

bozturkMSFT (approved):

vlada-shubina (approved):

src/Microsoft.TemplateEngine.Core/Util/ProcessorState.cs
DavidKarlas: Done.

Issue #2895, opened 2021-03-02
Template Localizer tool


bozturkMSFT:
"TemplateLocalizer" tool will be a dotnet tool distributed via NuGet.org.
The behavior will be:
- read the given template.json file
- extract strings that should be localized
- create a templatestrings.json file under "localize" directory for each of the supported languages
  - If files already exist, existing localized strings should be preserved

The following tasks track the work to be completed
[ ] https://github.com/dotnet/templating/issues/3016 Extract localizable strings from template.json
[ ] https://github.com/dotnet/templating/issues/3017 Create/Update templatestrings.json file
[ ] https://github.com/dotnet/templating/issues/3022 Update localization deserializer to read templatestrings.json files
[ ] https://github.com/dotnet/templating/issues/3045 Add tests to template localizer projects
[ ] https://github.com/dotnet/templating/issues/3046 Update test templates to use new localization files

PR #2940, opened 2021-03-16
Template Localizer tool

bozturkMSFT:
### Problem
https://github.com/dotnet/templating/issues/2895

### Solution
This is a draft delivering the main project structure and command line parsing.

### Checks:
- [ ] Added unit tests


src/Microsoft.TemplateEngine.TemplateLocalizer/Commands/Export/ExportCommand.cs
```diff
@@ -0,0 +1,196 @@
...
+                runningExportTasks.Add(new Core.TemplateLocalizer(Logger).ExportLocalizationFiles(templateJsonFile.FullName, exportOptions, cancellationToken));
+
+                if (runningExportTasks.Count == ConcurrencyLevel)
+                {
+                    // We have reached the concurrency limit. Wait for one of the tasks to finish before issuing new ones.
+                    Task<ExportResult>? completedTask = null;
+
+                    try
+                    {
+                        completedTask = await Task.WhenAny(runningExportTasks).ConfigureAwait(false);
```
donJoseLuis, line 110: i'm probably missing something, at a glance this block looks like...
`exportResults.Add(await Task.WhenAny(runningExportTasks).ConfigureAwait(false));`

donJoseLuis, line 110: consider WaitAny & its' overload that accepts a cancellation token
https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.waitany?view=net-5.0#System_Threading_Tasks_Task_WaitAny_System_Threading_Tasks_Task___System_Threading_CancellationToken_

src/Microsoft.TemplateEngine.TemplateLocalizer/Commands/Export/ExportCommand.cs
```diff
@@ -0,0 +1,196 @@
...
+            });
+            exportCommand.Handler = this;
+
+            return exportCommand;
+        }
+
+        protected override async Task<int> Execute(ExportCommandArgs args, CancellationToken cancellationToken)
+        {
+            // Upgrade the cancellation token to allow also cancelling within this method.
+            using CancellationTokenSource cts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
```
donJoseLuis, line 58: is this necessary?  Can the token passed via method param be passed to the awaitables called in the method?

src/Microsoft.TemplateEngine.TemplateLocalizer/Commands/Export/ExportCommand.cs
```diff
@@ -0,0 +1,196 @@
...
+                    }
+
+                    if (completedTask != null)
+                    {
+                        runningExportTasks.Remove(completedTask);
+                    }
+                }
+            }
+
+            try
```
donJoseLuis, line 135: please modularize this method, it will make life easier moving forward

src/Microsoft.TemplateEngine.TemplateLocalizer/Commands/Export/ExportCommand.cs
```diff
@@ -0,0 +1,196 @@
...
+            {
+                Logger.LogError(LocalizableStrings.generic_log_commandExecutionFailed, CommandName);
+                return 1;
+            }
+
+            PrintResults(exportResults);
+            return cts.IsCancellationRequested ? 1 : 0;
+        }
+
+        private void PrintResults(IReadOnlyList<ExportResult> results)
```
donJoseLuis, line 172: super minor:  printing results seems like it does not belong to the export command abstraction.  if you agree, consider another pattern to carry out the printing...such as a decorator.

src/Microsoft.TemplateEngine.TemplateLocalizer/Commands/Export/ExportCommandArgs.cs
```diff
@@ -0,0 +1,43 @@
...
+            TemplateJsonProviders = templatePath;
+            Languages = language;
+            SearchSubdirectories = recursive;
+            DryRun = dryRun;
+        }
+
+        /// <summary>
+        /// Gets or sets the template.json file providers.
+        /// </summary>
+        public IEnumerable<TemplateJsonProvider> TemplateJsonProviders { get; set; }
```
donJoseLuis, line 24: seeing how these properties are initialized in teh .ctor, if you don't expect the values to change, consider making these readonly.

src/Microsoft.TemplateEngine.TemplateLocalizer/Commands/Export/ExportCommandArgs.cs
```diff
@@ -0,0 +1,43 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Collections.Generic;
+
+namespace Microsoft.TemplateEngine.TemplateLocalizer.Commands.Export
+{
+    /// <summary>
+    /// Model class representing the arguments of <see cref="ExportCommand"/>.
+    /// </summary>
+    internal sealed class ExportCommandArgs
```
donJoseLuis, line 11: feels very similar to `ExportOptions`.  Any chance for consolidating state?  Even a common ancestor?

src/Microsoft.TemplateEngine.TemplateLocalizer/Program.cs
```diff
@@ -0,0 +1,41 @@
...
+            var rootCommand = new RootCommand();
+            rootCommand.Name = "dotnet-template-localizer";
+
+            foreach(ExecutableCommand command in commands)
+            {
+                command.Logger = logger;
+                rootCommand.AddCommand(command.CreateCommand());
+            }
+
+            return await rootCommand.InvokeAsync(args);
```
donJoseLuis, line 32: consider configuring to false the awaitable.

src/Microsoft.TemplateEngine.TemplateLocalizer/TemplateJsonProvider.cs
```diff
@@ -0,0 +1,64 @@
...
+{
+    internal sealed class TemplateJsonProvider
+    {
+        /// <summary>
+        /// Creates an instance of <see cref="TemplateJsonProvider"/>.
+        /// </summary>
+        /// <param name="path">A path to a template.jsonfile or a path to a directory that contains a template.json file.</param>
+        public TemplateJsonProvider(string path)
+        {
+            Path = path;
```
donJoseLuis, line 17: what if path is null or inexistent?

donJoseLuis (changes requested):

src/Microsoft.TemplateEngine.TemplateLocalizer/TemplateJsonProvider.cs
bozturkMSFT: That is OK. Validating and making sense of the user provided value is done later in the instance methods of this type.

src/Microsoft.TemplateEngine.TemplateLocalizer/Commands/Export/ExportCommandArgs.cs
bozturkMSFT: One of these is used for parsing the command line input, the other one is part of the core API. Some members are common, but they are semantically different. I wouldn't want one of these to change when I modify the other either. So it makes more sense to me having them separate.

src/Microsoft.TemplateEngine.TemplateLocalizer/Commands/Export/ExportCommandArgs.cs
bozturkMSFT: Constructor is here to help command line parser map the members to user input. This type itself is basically a model and it should be OK to edit the members if needed (could be useful for testing).

src/Microsoft.TemplateEngine.TemplateLocalizer.Core/TemplateLocalizer.cs
```diff
@@ -0,0 +1,38 @@
...
+        public TemplateLocalizer() : this(NullLogger.Instance) { }
+
+        public TemplateLocalizer(ILogger logger)
+        {
+            _logger = logger;
+        }
+
+        public async Task<ExportResult> ExportLocalizationFiles(string templateJsonPath, ExportOptions options, CancellationToken cancellationToken)
+        {
+            ExportResult result = new ExportResult();
```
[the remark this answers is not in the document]
bozturkMSFT: This method is a stub and not yet implemented. Contents are here to test the command line parsing.

src/Microsoft.TemplateEngine.TemplateLocalizer.Core/TemplateLocalizer.cs
```diff
@@ -0,0 +1,38 @@
...
+        private readonly ILogger _logger;
+
+        public TemplateLocalizer() : this(NullLogger.Instance) { }
+
+        public TemplateLocalizer(ILogger logger)
+        {
+            _logger = logger;
+        }
+
+        public async Task<ExportResult> ExportLocalizationFiles(string templateJsonPath, ExportOptions options, CancellationToken cancellationToken)
```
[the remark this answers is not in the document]
bozturkMSFT: Sorry, I didn't mention this in the PR description. This method is a stub and not yet implemented. Contents are here to test the command line parsing.

src/Microsoft.TemplateEngine.TemplateLocalizer/Commands/ExecutableCommand.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// <returns>The created command.</returns>
+        public abstract Command CreateCommand();
+
+        /// <summary>
+        /// Executes the command with the given arguments specified in the invocation context.
+        /// </summary>
+        /// <returns>A task that tracks the asynchronous operation.
+        /// 0 result from the completed task means that the command execution was successful.
+        /// where any other value indicates a failure.</returns>
+        public abstract Task<int> InvokeAsync(InvocationContext context);
```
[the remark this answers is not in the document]
bozturkMSFT: This method signature originates from the command line parser library. Each command returns an int, representing the result of the process. In success, we return 0. But in failure, the returned value will potentially depend on the type of the error that occured.

src/Microsoft.TemplateEngine.TemplateLocalizer/Commands/ExecutableCommand.cs
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    /// <summary>
+    /// Represents an <see cref="ExecutableCommand"/> where the command line arguments
+    /// are mapped to a model class.
+    /// </summary>
+    /// <typeparam name="TModel">Type of the model class whose properties represent the arguments.</typeparam>
+    internal abstract class ModelBoundExecutableCommand<TModel> : ExecutableCommand where TModel : class
```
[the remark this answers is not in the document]
bozturkMSFT: Nope. Any class will do. Arguments passed by the user will be matched with the members of the type here, but there are no generic type constraints that can help with that.

src/Microsoft.TemplateEngine.TemplateLocalizer/Commands/Export/ExportCommand.cs
bozturkMSFT: `WaitAny` will block the current thread, unfortunately.

bozturkMSFT: We need a reference to the completed task to be able to remove it from the list `runningExportTasks`.

src/Microsoft.TemplateEngine.TemplateLocalizer/Commands/Export/ExportCommand.cs
bozturkMSFT: If one of the tasks below fails, we want to cancel all the other tasks. We can't do it with a `CancellationToken` as it doesn't expose a `Cancel` method. Therefore, I created another `CancellationTokenSource` linked to the given cancellation token, which I can cancel if needed.

src/Microsoft.TemplateEngine.TemplateLocalizer.Core/ExportOptions.cs
```diff
@@ -0,0 +1,48 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Collections.Generic;
+
+namespace Microsoft.TemplateEngine.TemplateLocalizer.Core
+{
+    public struct ExportOptions
```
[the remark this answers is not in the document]
bozturkMSFT: I was hoping to use `ImmutableList` instead of an `IEnumerable<string>`, but the type doesn't exist in .Net Standard 2.1. Open to suggestions here.

src/Microsoft.TemplateEngine.TemplateLocalizer/Commands/Export/ExportCommand.cs
ladipro: +1. Also, writing what's basically a scheduler with a given max concurrency level feels like a problem that must have been solved elsewhere. What would be the downside of doing this with a `Parallel.ForEach` for example?

ladipro (commented):
I've left a few comments inline.

src/Microsoft.TemplateEngine.TemplateLocalizer/Commands/Export/ExportCommand.cs
vlada-shubina: Also, what is the downside of just let TPL schedule this work here? In most of the cases we would have just one task. Is implementing the scheduler worth it? I feel like reaching out 16 is 0.01% case. 
I see the point in such implementation when it's likely to produce lots of tasks which may flood or you want to limit it more than max amount of threads. In case control is needed, I would consider using `Parallel` as well.

src/Microsoft.TemplateEngine.TemplateLocalizer/Commands/Export/ExportCommand.cs
bozturkMSFT: I was considering this case: there is a problem with one of my templates and I'm trying to fix it, I would want to get back to my work as soon as I know about the issue (ie. without waiting for all the templates to be exported). But I guess for that case, you can invoke the tool only for that specific template as well.

src/Microsoft.TemplateEngine.TemplateLocalizer/Commands/Export/ExportCommand.cs
```diff
@@ -0,0 +1,196 @@
...
+                {
+                    break;
+                }
+
+                string templateDirectory = Path.GetDirectoryName(templateJsonFile.FullName) ?? string.Empty;
+                exportOptions.TargetDirectory = Path.Combine(templateDirectory, "localize");
+
+                runningExportTasks.Add(new Core.TemplateLocalizer(Logger).ExportLocalizationFiles(templateJsonFile.FullName, exportOptions, cancellationToken));
+
+                if (runningExportTasks.Count == ConcurrencyLevel)
```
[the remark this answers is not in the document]
bozturkMSFT: Thanks for pointing out. That should only happen if we pass `null` to `Task.WhenAny` or one of the tasks we pass is null, neither of which should apply here. Which makes me question the necessity of the `if` below or the action to take if `completedTask==null`.

I plan to ensure there are no null entries in the task list and get rid of the `if` surrounding the `runningExportTasks.Remove`. Let me know if you suggest another action.

src/Microsoft.TemplateEngine.TemplateLocalizer/Commands/Export/ExportCommand.cs
bozturkMSFT: `Parallel.ForEach` is a blocking call and unfortunately doesn't support async delegates. This probably won't surface as a big issue since this is part of the console app and not the API, but this was the original idea I had behind staying away from `Parallel.` here.

The only replacements I could find that fully supports `async/await`:
- Using semaphore slim as [suggested here](https://devblogs.microsoft.com/pfxteam/implementing-a-simple-foreachasync/)
- Creating a scheduler as done in this example in docs page which I wish was built into the libraries: [LimitedConcurrencyLevelTaskScheduler](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskscheduler?view=net-5.0#examples)

The official support for [Parallel.ForEachAsync](https://github.com/dotnet/runtime/pull/46943) is here, but available starting .NET 6.0.

Any ideas @ladipro @vlada-shubina @donJoseLuis ?

src/Microsoft.TemplateEngine.TemplateLocalizer/Commands/Export/ExportCommand.cs
```diff
@@ -0,0 +1,196 @@
...
+                Description = LocalizableStrings.command_export_help_language_description,
+                Argument = new Argument()
+                {
+                    Arity = ArgumentArity.OneOrMore,
+                    ArgumentType = typeof(string),
+                },
+            });
+            exportCommand.AddOption(new Option("-d")
+            {
+                Name = "--DryRun",
```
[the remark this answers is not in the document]
bozturkMSFT: Casing is wrong here, thanks for pointing out. Let me know if I'm wrong here, but options don't have types, arguments of options do. This option doesn't have any arguments, so you are not allowed to specify a value anyway. You either say `--dry-run` or you don't.

src/Microsoft.TemplateEngine.TemplateLocalizer/Commands/Export/ExportCommand.cs
```diff
@@ -0,0 +1,196 @@
...
+            var exportCommand = new Command(CommandName, LocalizableStrings.command_export_help_description);
+            exportCommand.AddArgument(new Argument("template-path")
+            {
+                Arity = ArgumentArity.OneOrMore,
+                ArgumentType = typeof(string),
+                Description = LocalizableStrings.command_export_help_templatePath_description,
+            });
+            exportCommand.AddOption(new Option("-r")
+            {
+                Name = "--recursive",
```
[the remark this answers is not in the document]
bozturkMSFT: Maybe this is a `System.CommandLine` thing. This option is doesn't have arguments, so it doesn't have _any_ value. The option is either specified or not which translates to true or false.

src/Microsoft.TemplateEngine.TemplateLocalizer/LocalizableStrings.resx
```diff
@@ -0,0 +1,166 @@
...
+  </data>
+  <data name="command_export_help_language_description" xml:space="preserve">
+    <value>The list of languages to be supported by this template. The following language list will be used as default if this option is omitted: cs, de, en, es, fr, it, ja, ko, pl, pt-BR, ru, tr, zh-Hans, zh-Hant</value>
+  </data>
+  <data name="command_export_help_recursive_description" xml:space="preserve">
+    <value>When specified, subdirectories are also searched for template.json files.</value>
+    <comment>Do not localize: "template.json"</comment>
+  </data>
+  <data name="command_export_help_templatePath_description" xml:space="preserve">
+    <value>Specifies the path to the template.json file. If a directory is given, template.json file will be searched within the directory. If --recursive options is specified, all the template.json files under the given directory and its subdirectories will be taken as input.</value>
```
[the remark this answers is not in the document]
bozturkMSFT: That makes sense. I will add this.

src/Microsoft.TemplateEngine.TemplateLocalizer/TemplateJsonProvider.cs
```diff
@@ -0,0 +1,64 @@
...
+        public string Path { get; private set; }
+
+        /// <summary>
+        /// Using the <see cref="Path"/>, finds and returns all the template.json files.
+        /// </summary>
+        /// <param name="searchSubdirectories">Indicates weather the subdirectories should be searched
+        /// in the case that <see cref="Path"/> points to a directory. This parameter has no effect
+        /// if <see cref="Path"/> points to a file.</param>
+        /// <returns></returns>
+        public IEnumerable<FileInfo> GetTemplateJsonFiles(bool searchSubdirectories)
```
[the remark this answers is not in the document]
bozturkMSFT: That is a fair point. I will turn this into a static method. I guess I may have been carried away by the command line library's ability to deserialize passed values directly into custom types 🙂

src/Microsoft.TemplateEngine.TemplateLocalizer/Commands/Export/ExportCommand.cs
vlada-shubina: I would simply use `await Task.WhenAll(exportTasks).ConfigureAwait(false);` here, as mentioned before I don't see use case for slicing tasks here, and reaching out 16 concurrent tasks is very little chance.

src/Microsoft.TemplateEngine.TemplateLocalizer/Commands/Export/ExportCommand.cs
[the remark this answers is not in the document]
vlada-shubina: I think I need to learn more about `System.CommandLine`, in current parser options can have value or even multiple values.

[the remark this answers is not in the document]
bozturkMSFT: We can achieve that by adding arguments to options. See language option for instance. It allows specifying one or more languages like this `-l en fr-FR de`.

https://github.com/dotnet/templating/blob/2b39c74508279322df0796fa436ad36fa1efc148/src/Microsoft.TemplateEngine.TemplateLocalizer/Commands/Export/ExportCommand.cs#L35-L44

src/Microsoft.TemplateEngine.TemplateLocalizer.Core/ExportOptions.cs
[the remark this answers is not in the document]
donJoseLuis: One thing...
- Remove the _set_ access to the properties.  
- Set state in a .ctor. 
These instances would be immutable with the current data types.

Did you review: https://docs.microsoft.com/en-us/dotnet/api/system.collections.immutable.immutablelist-1?view=net-5.0

src/Microsoft.TemplateEngine.TemplateLocalizer/Commands/Export/ExportCommandArgs.cs
donJoseLuis: Common State
      |
      |____________Child_0
      |
      |____________Child_1

src/Microsoft.TemplateEngine.TemplateLocalizer/Commands/Export/ExportCommandArgs.cs
donJoseLuis: you're conscientiously voting down immutability (thread safety) for testability.  Specifically, testing of cases where an instance is created & after instance state is changed via set.

If the type will be relevant to concurrent scenarios, immutability may win in this case.

src/Microsoft.TemplateEngine.TemplateLocalizer/Commands/Export/ExportCommand.cs
bozturkMSFT: 16 is an arbitrary number I picked here, but the number of tasks going beyond 16 is not hard at all. We already have 6 templates in our repo under ProjectTemplates.6.0 folder. There will be 6 more with .NET 7. This is just for our very basic templates and I'm not considering ItemTemplates here. I imagine this may grow significantly for a repo where templates are more important, such as aspnetcore.

I don't believe we will re-invoke the tool for every template.json in our repo. This will be too much to manage from the build pipeline (you will have to pass the paths to all the template.json files in our repo in the yaml file). Same goes for other repos. I imagine that there will be a property where each repo will set in their own pipeline which will point to the root folder of all the templates. Then the tool will go over all the templates under this folder.

src/Microsoft.TemplateEngine.TemplateLocalizer/Commands/ExecutableCommand.cs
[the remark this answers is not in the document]
bozturkMSFT: We don't own the signature of this one.

src/Microsoft.TemplateEngine.TemplateLocalizer/Commands/Export/ExportCommandArgs.cs
bozturkMSFT: There is no thread safety concern at the moment. I have removed the setters anyway though. Let's start without them and bring them back if needed.

src/Microsoft.TemplateEngine.TemplateLocalizer/Commands/Export/ExportCommandArgs.cs
bozturkMSFT: I already had that picture in mind while writing my comment. Can you elaborate on how that answers my comment?

src/Microsoft.TemplateEngine.TemplateLocalizer/TemplateJsonProvider.cs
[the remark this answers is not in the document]
bozturkMSFT: I removed the type.

src/Microsoft.TemplateEngine.TemplateLocalizer/TemplateJsonProvider.cs
```diff
@@ -0,0 +1,64 @@
...
+        /// <param name="path">A path to a template.jsonfile or a path to a directory that contains a template.json file.</param>
+        public TemplateJsonProvider(string path)
+        {
+            Path = path;
+        }
+
+        /// <summary>
+        /// Gets the path used by this provider to search for template.json files.
+        /// </summary>
+        public string Path { get; private set; }
```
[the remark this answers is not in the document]
bozturkMSFT: I removed this type.

src/Microsoft.TemplateEngine.TemplateLocalizer/Program.cs
```diff
@@ -0,0 +1,41 @@
...
+        };
+
+        private static async Task<int> Main(string[] args)
+        {
+            ILogger logger = InitializeLogger();
+
+            var rootCommand = new RootCommand();
+            rootCommand.Name = "dotnet-template-localizer";
+
+            foreach(ExecutableCommand command in commands)
```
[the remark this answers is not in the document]
bozturkMSFT: Analyzers are working at "suggestion" level now, otherwise the project wouldn't build. This specific case was caught, but I missed the suggestion. Fixed now.

src/Microsoft.TemplateEngine.TemplateLocalizer/Commands/Export/ExportCommand.cs
bozturkMSFT: I replaced the logic here with a bit simpler batching and didn't use `Parallel.ForEach` since it lacks support for async.

src/Microsoft.TemplateEngine.TemplateLocalizer/Microsoft.TemplateEngine.TemplateLocalizer.csproj
```diff
@@ -0,0 +1,39 @@
...
+
+    <PackAsTool>true</PackAsTool>
+    <ToolCommandName>dotnet-template-localizer</ToolCommandName>
+    <LangVersion>9.0</LangVersion>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="6.0.0-preview.3.21164.3" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="6.0.0-preview.3.21164.3" />
+    <PackageReference Include="System.CommandLine" Version="2.0.0-beta1.21109.1" />
```
DavidKarlas, line 18: Same comment as in other .csproj file, lets use `Update` version in Directory.Build.props file on root of repo.

src/Microsoft.TemplateEngine.TemplateLocalizer.Core/ExportOptions.cs
```diff
@@ -0,0 +1,96 @@
...
+            "cs",
+            "de",
+            "en",
+            "es",
+            "fr",
+            "it",
+            "ja",
+            "ko",
+            "pl",
+            "pt-BR",
```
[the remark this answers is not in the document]
bozturkMSFT: I checked the latest installed sdk on my machine and looked what loc files I have.

src/Microsoft.TemplateEngine.TemplateLocalizer.Core/TemplateLocalizer.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            _logger = (ILogger?)loggerFactory?.CreateLogger<TemplateLocalizer>() ?? NullLogger.Instance;
+        }
+
+        public async Task<ExportResult> ExportLocalizationFilesAsync(string templateJsonPath, ExportOptions options, CancellationToken cancellationToken = default)
+        {
+            ExportResult result = new ExportResult();
+            result.TemplateJsonPath = templateJsonPath;
+            result.ErrorMessage = "Operation failed.";
+            return result;
```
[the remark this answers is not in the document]
bozturkMSFT: This method is just for testing. It will be fully replaced with some logic that actually makes sense 🙂

src/Microsoft.TemplateEngine.TemplateLocalizer/Commands/Export/ExportCommand.cs
```diff
@@ -0,0 +1,238 @@
...
+                }
+
+                int batchOffset = batchSize * batchIndex;
+                for (int taskIndex = 0; taskIndex < batchSize && batchOffset + taskIndex < templateJsonFiles.Count; taskIndex++)
+                {
+                    string templateJsonPath = templateJsonFiles[batchOffset + taskIndex];
+                    string templateDirectory = Path.GetDirectoryName(templateJsonPath) ?? string.Empty;
+                    string targetDirectory = Path.Combine(templateDirectory, "localize");
+
+                    ExportOptions exportOptions = new (args.DryRun, targetDirectory, args.Languages);
```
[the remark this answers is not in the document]
bozturkMSFT: I pity `new ()`. Just because he is new, he gets bullied. He just wants to be your friend and I don't think he deserves this 🙁

PR #3047, opened 2021-04-15
Use `--debug:custom-hive` for `.templateengine` and not for `.templateengine/dotnetcli/v6.100.0.0`

DavidKarlas:
### Problem
Until now `--debug:custom-hive` represented only host+version specific folder, which meant that `packages.json` aka. user installed packages were outside this folder...

### Solution

With this change we custom-hive covers everything under `.templateengine` folder... We can stop using `HOME` environment variable for unit test and use `--debug:custom-hive` instead...

### Checks:
- [ ] Added unit tests


DavidKarlas:
Notice that I added one more commit into this mix: "--debug:ephemeral-hive"

bozturkMSFT (approved):

vlada-shubina (approved):

PR #3051, opened 2021-04-16
Activator.CreateInstance can't create instance if constructor is not public

DavidKarlas:
### Problem
Will reported this:
```
dotnet new console
The template "Console Application" was created successfully.

 

Processing post-creation actions...
No parameterless constructor defined for type 'Microsoft.TemplateEngine.Cli.PostActionProcessors.DotnetRestorePostActionProcessor'.
```

### Solution
Problem was that constructor on that type is internal. I looked for other components as well.

### Checks:
- [ ] Added unit tests


DavidKarlas:
🤦

PR #3052, opened 2021-04-16
Remove cycle dependency with SDK repo

DavidKarlas:
### Problem
#2651

### Solution
This commit removes dependency on Microsoft.DotNet.TemplateLocator and also removes `OptionalWorkloadProviderFactory` which will be reintroduced on SDK repo
This is fixing #2651

### Checks:
- [ ] Added unit tests


vlada-shubina (approved):
There is also HiveException, related to OW

bozturkMSFT (approved):

PR #3072, opened 2021-04-21
Fixes #2741: FirstRun can take long time because it needs to install/scan multiple packages, this could be done in parallel

### LINKED ISSUE #2741: FirstRun can take long time because it needs to install/scan multiple packages, this could be done in parallel

DavidKarlas:
### Problem
#2741  First run of `dotnet new` is not as fast as it could be...

### Solution
This can save around 50% of time by doing it in parallel, primary because scanning is doing unzipping of .nupkg and parsing .json files

### Checks:
- [ ] Added unit tests


DavidKarlas:
I'm parking this PR for now, since we are going to probably change this area soon.

Issue #3046, opened 2021-04-15
[Localizer] Update test templates to use new localization files


bozturkMSFT:
We have templates in our repo that are used by tests. They need to be updated so that localization tests are running, they will have the necessary files to work with.

- Update TemplateWithLocalization template under TestTemplates project to contain templatestrings.json files.

Issue #2911, opened 2021-03-08
de-AT locale should read localizations from de-DE 


bozturkMSFT:

christiannagel:
I think it should also be considered to move *de-DE* localizations to *de*

PR #3089, opened 2021-04-23
templatestrings.json deserialization

### LINKED ISSUE #3022: Update localization deserializer to read templatestrings.json files
Template engine currently reads localization files with names formatted as "<language>.template.json". This is the old loc file format and the new loc files are named in  "<language>.templatestrings.json" format.

Update template engine loc file deserialization logic to:
- Read xx.templatestrings.json files instead of xx.template.json
- Correctly parse the new keys to determine what string it is localizing.
- Populate the data model with the localized string.

bozturkMSFT:
### Problem
Closes #3022, #3046 and #2911

### Solution
Done: Added logic to discover and templatestrings.json files under "localize" folder near the template.json.
Done: Added logic to deserialize templatestrings.json files.
Done: Post actions are now stored in a list instead of a dictionary.
Done: `Identity` was removed from Localization model.
Done: Localization keys no longer use `.` as a separator since it conflicts with some common choices such as "net5.0". Instead, we are now using '/'. I also have PM approval for this.

### Checks:
- [x] Added unit tests


src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/RunnableProjectGenerator.cs
```diff
@@ -118,58 +119,41 @@ public IParameterSet GetParametersForTemplate(IEngineEnvironmentSettings environ
...
+                            string locale = locFile.Name.Substring(0, locFile.Name.Length - LocalizationFileSuffix.Length);
+
+                            if (TryGetLangPackFromFile(locFile, out ILocalizationModel locModel))
+                            {
+                                ILocalizationLocator locator = new LocalizationLocator()
+                                {
+                                    Locale = locale,
+                                    MountPointUri = source.MountPointUri,
+                                    ConfigPlace = locFile.FullPath,
+                                    Identity = locModel.Identity,
```
DavidKarlas, line 147: Since code later uses this to match... At least for now, you want to use `template.Identity` here...

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/RunnableProjectGenerator.cs
```diff
@@ -118,58 +119,41 @@ public IParameterSet GetParametersForTemplate(IEngineEnvironmentSettings environ
...
-                            Identity = locModel.Identity,
-                            Author = locModel.Author,
-                            Name = locModel.Name,
-                            Description = locModel.Description,
-                            ParameterSymbols = locModel.ParameterSymbols
-                        };
-                        localizations.Add(locator);
+                            string locale = locFile.Name.Substring(0, locFile.Name.Length - LocalizationFileSuffix.Length);
+
+                            if (TryGetLangPackFromFile(locFile, out ILocalizationModel locModel))
```
DavidKarlas, line 140: From my understanding of this code... Looks like this is still calling old "deserialize" logic... Why not use new one?

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/LocalizationModelDeserializer.cs
```diff
@@ -0,0 +1,230 @@
...
+        /// </list>
+        /// </summary>
+        private static IReadOnlyList<IPostActionLocalizationModel?> LoadPostActionModels(IEnumerable<(IEnumerable<string> nameParts, string localizedString)> strings)
+        {
+            var results = new List<IPostActionLocalizationModel?>();
+
+            foreach (var postActionParts in strings.GroupBy(p => p.nameParts.FirstOrDefault()))
+            {
+                if (!GetIndexFromString(postActionParts.Key, PostActionIndexPrefix, PostActionIndexSuffix, out int postActionIndex) ||
+                    postActionIndex > 256)
```
DavidKarlas, line 141: What does 256 represent, I understand its crazy high number, but I don't see much benefit having this here...

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/LocalizationModelDeserializer.cs
```diff
@@ -0,0 +1,230 @@
...
+        /// </list>
+        /// </summary>
+        private static IReadOnlyList<string?> LoadManualInstructionModels(IEnumerable<(IEnumerable<string> nameParts, string localizedString)> strings)
+        {
+            var results = new List<string?>();
+
+            foreach (var instructionParts in strings.GroupBy(p => p.nameParts.FirstOrDefault()))
+            {
+                if (!GetIndexFromString(instructionParts.Key, ManualInstructionIndexPrefix, ManualInstructionIndexSuffix, out int instructionIndex) ||
+                    instructionIndex > 256)
```
DavidKarlas, line 184: Same, 256...

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/LocalizationModelDeserializer.cs
```diff
@@ -0,0 +1,230 @@
...
+        {
+            var locModel = new LocalizationModel();
+            var parameterLocalizations = new Dictionary<string, ParameterSymbolLocalizationModel>();
+
+            // Property names are in format: symbols.framework.choices.0.description
+            // Split them using '.' and store together with the localized string (property value).
+            IEnumerable<(IEnumerable<string> nameParts, string localizedString)> stringsWithNames = data.Properties()
+                .Where(p => p.Value.Type == JTokenType.String)
+                .Select(p => (p.Name.Split('.').AsEnumerable(), p.Value.ToString()))
+                .ToList();
```
DavidKarlas, line 32: I'm not crazy in love with all this Linq and .Split, I would prefer some more .StartsWith("symbols") and .IndexOf('.') stuff, but I guess its fine as long as it doesn't show up on perf...

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/LocalizationModelDeserializer.cs
```diff
@@ -0,0 +1,230 @@
...
+
+        public static ILocalizationModel Deserialize(JObject data)
+        {
+            var locModel = new LocalizationModel();
+            var parameterLocalizations = new Dictionary<string, ParameterSymbolLocalizationModel>();
+
+            // Property names are in format: symbols.framework.choices.0.description
+            // Split them using '.' and store together with the localized string (property value).
+            IEnumerable<(IEnumerable<string> nameParts, string localizedString)> stringsWithNames = data.Properties()
+                .Where(p => p.Value.Type == JTokenType.String)
```
DavidKarlas, line 30: I think I would prefer we throw some exception if this happens rather then silently swallow invalid json and let author struggle figuring it out...

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/LocalizationModelDeserializer.cs
DavidKarlas: I'm not talking just about this statement, but all others in this method and called methods like `LoadSymbolModels`, `LoadPostActionModels`...

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/LocalizationModelDeserializer.cs
bozturkMSFT: If someone accidentally types a high number in their localization file, I didn't want to allocate the memory for it. Can you think of any scenarios where this would hurt user experience?

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/LocalizationModelDeserializer.cs
bozturkMSFT: How about logging a warning? At least the user can continue to use the template until the author can fix it.

bozturkMSFT: I plan to add some more logs for other problems as well.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/RunnableProjectGenerator.cs
bozturkMSFT: This method call will have the new serialization logic. I just haven't gotten to it just yet. This PR is a draft after all :)

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/RunnableProjectGenerator.cs
bozturkMSFT: I searched the repo and that Identity is never used anywhere as far as I can see. I plan to remove it if that is really the case.

DavidKarlas: Here is used https://github.com/dotnet/templating/blob/main/src/Microsoft.TemplateEngine.Edge/Settings/TemplateCache.cs#L376 and later its used at https://github.com/dotnet/templating/blob/main/src/Microsoft.TemplateEngine.Edge/Settings/TemplateCache.cs#L204 to actually match it with "parent" template.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/LocalizationModelDeserializer.cs
DavidKarlas: I'm not following, where is memory allocated for it? I'm just not a fan of complicating code with some artificial limit and in theory limiting template post actions limit, I agree nobody should reach this limit but still...

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/LocalizationModelDeserializer.cs
DavidKarlas: I think that if we just throw exception.... template author should be able to notice this mistake before template is released.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/LocalizationModelDeserializer.cs
```diff
@@ -0,0 +1,230 @@
...
+using System.Linq;
+using Microsoft.TemplateEngine.Abstractions;
+using Microsoft.TemplateEngine.Orchestrator.RunnableProjects.Localization;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.TemplateEngine.Orchestrator.RunnableProjects
+{
+    internal static class LocalizationModelDeserializer
+    {
+        private const string PostActionIndexPrefix = "postActions[";
```
vlada-shubina, line 16: should those to be defined in localizer to be in sync? Or vice versa use those from localizer?

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/LocalizationModelDeserializer.cs
```diff
@@ -0,0 +1,230 @@
...
+
+        /// <summary>
+        /// Parses the index from string of format &lt;prefix&gt;index&lt;suffix&gt;.
+        /// </summary>
+        /// <param name="value">The string to be parsed.</param>
+        /// <param name="prefix">The prefix that the string should begin with. Index starts after the prefix.</param>
+        /// <param name="suffix">The suffix that the string should end with. Index ends with the suffix.</param>
+        /// <returns>True if parsing succeeded. False if the string was incorrectly formatted or the index
+        /// string cannot be converted to an integer.</returns>
+        private static bool GetIndexFromString(string value, string prefix, string suffix, out int index)
```
vlada-shubina, line 212: should it be extension instead?

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/LocalizationModelDeserializer.cs
```diff
@@ -0,0 +1,230 @@
...
+
+            // Property names are in format: symbols.framework.choices.0.description
+            // Split them using '.' and store together with the localized string (property value).
+            IEnumerable<(IEnumerable<string> nameParts, string localizedString)> stringsWithNames = data.Properties()
+                .Where(p => p.Value.Type == JTokenType.String)
+                .Select(p => (p.Name.Split('.').AsEnumerable(), p.Value.ToString()))
+                .ToList();
+
+            var symbols = LoadSymbolModels(stringsWithNames
+                .Where(s => s.nameParts.FirstOrDefault() == "symbols")
```
vlada-shubina, line 35: should constants be extracted and reused in localizer/here?

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/LocalizationModelDeserializer.cs
```diff
@@ -0,0 +1,230 @@
...
+        /// Generates parameter symbol localization models. The given name parts should begin with the parameter name
+        /// as shown below and should not contain the string "symbols".
+        /// <list type="table">
+        /// <item>framework.displayName</item>
+        /// <item>framework.description</item>
+        /// <item>framework.choices.net5_0.description</item>
+        /// <item>targetframeworkoverride.description</item>
+        /// </list>
+        /// </summary>
+        private static IReadOnlyDictionary<string, IParameterSymbolLocalizationModel> LoadSymbolModels(IEnumerable<(IEnumerable<string> nameParts, string localizedString)> strings)
```
vlada-shubina, line 63: I feel like name parts can be identified in this method instead, and starts with symbols and we do extraction in this method; same for similar methods. 
The input parameter should be probably json or parsed string collection (as string, localized string) and the method should filter relevant ones.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/LocalizationModelDeserializer.cs
vlada-shubina: addition to comment below:
probably we can do here:
`data.Properties().Where(p => p.Value.Type == JTokenType.String).Select(p => (p.Name, p.Value.ToString()))` and pass it to respective private methods, which evaluate which strings they should act upon.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/RunnableProjectGenerator.cs
bozturkMSFT: That is `ILocalizationLocator.Identity`, I will not change it. I meant `ILocalizationModel.Identity` isn't used.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/LocalizationModelDeserializer.cs
bozturkMSFT: Right below this if statement:
```csharp
while (results.Count <= postActionIndex)
{
    // We haven't processed localizations for these post actions. Just put null for now.
    // Localization file also may not specify translations for all the post actions.
    results.Add(null);
}
```

bozturkMSFT: This won't limit the number of post actions, but anything after 256 wouldn't be localized.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/LocalizationModelDeserializer.cs
bozturkMSFT: This is useful in very few cases (probably only in this file). I don't think it would be very useful as an extension method.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/LocalizationModelDeserializer.cs
bozturkMSFT: That was my first go to. However, I ended up doing the same thing in multiple places - creating duplicated code. Any property that goes deeper than the first level will need this.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/LocalizationModelDeserializer.cs
bozturkMSFT: These things are hidden behind `KeyCreator`s, so there is no const string like this in the localizer. I was initially planning to share some assembly between localizer and TE, but couldn't find a meaningful way to do it.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/PostActionModel.cs
```diff
@@ -75,16 +62,15 @@ internal static IReadOnlyList<IPostActionModel> ListFromJArray(JArray jObject, I
...
+                        instructionOptions.Add(new ManualInstructionModel(text, manualInstructions[i].ToString("condition")));
                     }
                 }
 
                 PostActionModel model = new PostActionModel()
                 {
                     Condition = action.ToString(nameof(model.Condition)),
                     Description = actionLocalizations?.Description ?? action.ToString(nameof(model.Description)),
-                    ActionId = actionId,
+                    ActionId = action.ToGuid(nameof(ActionId)),
```
vlada-shubina, line 73: is it a valid change?

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/RunnableProjectGenerator.cs
```diff
@@ -118,58 +118,41 @@ public IParameterSet GetParametersForTemplate(IEngineEnvironmentSettings environ
...
-                            MountPointUri = source.MountPointUri,
-                            ConfigPlace = file.FullPath,
-                            Identity = locModel.Identity,
-                            Author = locModel.Author,
-                            Name = locModel.Name,
-                            Description = locModel.Description,
-                            ParameterSymbols = locModel.ParameterSymbols
-                        };
-                        localizations.Add(locator);
+                            string locale = locFile.Name.Substring(0, locFile.Name.Length - LocalizationFileSuffix.Length);
```
vlada-shubina, line 137: should we check if it's  a valid locale here? We can have a file called `bla.templatestrings.json`

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/RunnableProjectGenerator.cs
```diff
@@ -118,58 +118,41 @@ public IParameterSet GetParametersForTemplate(IEngineEnvironmentSettings environ
...
+
+                            if (TryGetLangPackFromFile(locFile, out ILocalizationModel locModel))
+                            {
+                                ILocalizationLocator locator = new LocalizationLocator()
+                                {
+                                    Locale = locale,
+                                    MountPointUri = source.MountPointUri,
```
vlada-shubina, line 144: Do we need MountPointUri and ConfigPlace? I believe they are always same as primary json now.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/RunnableProjectGenerator.cs
```diff
@@ -118,58 +118,41 @@ public IParameterSet GetParametersForTemplate(IEngineEnvironmentSettings environ
...
-                            Name = locModel.Name,
-                            Description = locModel.Description,
-                            ParameterSymbols = locModel.ParameterSymbols
-                        };
-                        localizations.Add(locator);
+                            string locale = locFile.Name.Substring(0, locFile.Name.Length - LocalizationFileSuffix.Length);
+
+                            if (TryGetLangPackFromFile(locFile, out ILocalizationModel locModel))
+                            {
+                                ILocalizationLocator locator = new LocalizationLocator()
```
vlada-shubina, line 141: nit: i would prefer to have parameterized constructor for it.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/RunnableProjectGenerator.cs
```diff
@@ -492,7 +475,7 @@ internal bool AreAllTemplatePathsValid(IRunnableProjectConfig templateConfig, Ru
         internal JObject ReadJObjectFromIFile(IFile file)
         {
             using (Stream s = file.OpenRead())
-            using (TextReader tr = new StreamReader(s, true))
+            using (TextReader tr = new StreamReader(s, System.Text.Encoding.UTF8, true))
```
vlada-shubina, line 478: Will it be a breaking change for anyone used different encoding now?

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/SimpleConfigModel.cs
```diff
@@ -713,7 +712,7 @@ internal static SimpleConfigModel FromJObject(IEngineEnvironmentSettings environ
             Dictionary<string, IReadOnlyList<IOperationProvider>> localizations = new Dictionary<string, IReadOnlyList<IOperationProvider>>();
             if (localizationModel != null && localizationModel.FileLocalizations != null)
             {
-                foreach (FileLocalizationModel fileLocalization in localizationModel.FileLocalizations)
+                foreach (IFileLocalizationModel fileLocalization in localizationModel.FileLocalizations)
```
vlada-shubina, line 715: please remove this logic - deprecated.

src/Microsoft.TemplateEngine.TemplateLocalizer.Core/KeyCreators/ChildValueKeyCreator.cs
```diff
@@ -28,12 +28,12 @@ public string CreateKey(JsonElement element, string? elementName, string? parent
         {
             if (!element.TryGetProperty(MemberPropertyName, out JsonElement keyProperty) || keyProperty.ValueKind != JsonValueKind.String)
             {
-                string owningElementName = (parentElementName == null ? elementName : (parentElementName + "." + elementName)) ?? string.Empty;
+                string owningElementName = (parentElementName == null ? elementName : (parentElementName + TemplateStringExtractor._keySeparator + elementName)) ?? string.Empty;
```
vlada-shubina, line 31: should we add the logic to TE forbidding using this symbol in name? (might be different PR)

vlada-shubina (changes requested):
I saw couple of classes added by no tests for them - will we benefit from new unit tests for deseriaization logic? Error cases?

Will be good to add integration test for:
- installing valid template with localization (check for no error)
- installing invalid template with location (check for error in output) (with/without other templates being installed at same time) 
- running template with invalid localization (in theory possible, if you install template from folder and modify localization).

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/RunnableProjectGenerator.cs
bozturkMSFT: Apparently, stream reader [already uses UTF8 by default](https://docs.microsoft.com/en-us/dotnet/api/system.io.streamreader.-ctor?view=net-5.0#System_IO_StreamReader__ctor_System_String_), but it failed to read the characters correctly in some of the tests which I suspect are caused by the differences in `StreamReader` in .NET 4.7. If someone was using a different encoding, most likely their template didn't work on other machines anyway.

I think letting users choose their own encoding won't work for us, since these files will need to be read by others who are potentially on machines with different default encodings. We can try to determine the encoding as well. But since the default encoding of json files according to [json specification](https://www.ietf.org/rfc/rfc4627.txt) is already UTF8 and considering how common UTF-8 is, I think it is safe to go with UTF-8 here.

src/Microsoft.TemplateEngine.Abstractions/ILocalizationModel.cs
```diff
@@ -1,27 +1,45 @@
...
+        /// <summary>
+        /// Gets the localization models for the parameter symbols defined in this template.
+        /// </summary>
         IReadOnlyDictionary<string, IParameterSymbolLocalizationModel> ParameterSymbols { get; }
 
-        IReadOnlyDictionary<Guid, IPostActionLocalizationModel> PostActions { get; }
+        /// <summary>
+        /// Gets the localization models for the post actions defined in this template.
+        /// </summary>
+        IReadOnlyList<IPostActionLocalizationModel?> PostActions { get; }
```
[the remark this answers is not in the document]
bozturkMSFT: If user skipped localizing some of the post actions, this container could have nulls inside. But I will replace this with a Dictionary, so null entries will disappear.

src/Microsoft.TemplateEngine.Abstractions/IParameterSymbolLocalizationModel.cs
```diff
@@ -16,8 +16,14 @@ public interface IParameterSymbolLocalizationModel
...
 
+        /// <summary>
+        /// Gets the localized description of the symbol.
+        /// </summary>
         string? Description { get; }
 
+        /// <summary>
+        /// Gets the localization models for choices of this symbol.
+        /// </summary>
         IReadOnlyDictionary<string, ParameterChoiceLocalizationModel> Choices { get; }
```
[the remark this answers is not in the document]
bozturkMSFT: Having this null would save a bit on allocations, but I think it makes consumption of this property easier since you don't have to check for nulls. It is also easier for deserializer to just create the dictionary in advance, instead of creating it if needed here: 
https://github.com/dotnet/templating/blob/bb688b93882c20f34ae00493bf2be4711bc1e221/src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/LocalizationModelDeserializer.cs#L108-L124

So I'd say we keep it non-nullable. Let me know what you think.

src/Microsoft.TemplateEngine.Abstractions/IPostAction.cs
```diff
@@ -8,14 +8,29 @@ namespace Microsoft.TemplateEngine.Abstractions
...
+        /// Gets a value indicating wheather the template instantiation should continue
+        /// in case of an error with this post action.
+        /// </summary>
         bool ContinueOnError { get; }
 
         IReadOnlyDictionary<string, string> Args { get; }
 
+        /// <summary>
+        /// Gets the instructions that should be manually performed by the user
+        /// as part of this post action.
```
DavidKarlas, line 32: Nit: Add comment, that this is used when host doesn't have post action processor implemented.

src/Microsoft.TemplateEngine.Edge/Settings/TemplateCache.cs
```diff
@@ -75,28 +78,31 @@ public void WriteTemplateCaches(Dictionary<string, DateTime> mountPoints)
         {
             bool hasContentChanges = false;
 
-            HashSet<string> foundTemplates = new HashSet<string>();
-            List<TemplateInfo> mergedTemplateList = new List<TemplateInfo>();
+            var foundTemplates = new HashSet<string>();
+            var mergedTemplateList = new List<TemplateInfo>(TemplateInfo.Count + _templateMemoryCache.Count);
```
DavidKarlas, line 82: Thank you!

You can ignore this comment.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/LocalizationModelDeserializer.cs
DavidKarlas: I think Vlada is refering to:
```
LoadPostActionModels(stringsWithNames
                .Where(s => s.nameParts.FirstOrDefault().StartsWith(PostActionIndexPrefix))
                .Select(s => (s.nameParts, s.localizedString)));
```
should become:
`LoadPostActionModels(stringsWithNames);`
but same for Symbols...

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/LocalizationModelDeserializer.cs
bozturkMSFT: As we discussed internally, this is not swallowing an error. This is a valid use case. I changed the implementation to use dictionary.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/LocalizationModelDeserializer.cs
bozturkMSFT: Limit was removed.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/LocalizationModelDeserializer.cs
bozturkMSFT: The engine and the localizer doesn't share any dependencies. We could create a new assembly just for these shared strings, but I don't think that will be worth it.

src/Microsoft.TemplateEngine.TemplateLocalizer.Core/TemplateStringExtractor.cs
```diff
@@ -168,9 +169,9 @@ private void ProcessStringElement(JsonElement element, string elementName, strin
                 return;
             }
 
-            string finalKey = data.KeyPrefix == null ? key : (data.KeyPrefix + "." + key);
+            string finalKey = data.KeyPrefix == null ? key : (data.KeyPrefix + _keySeparator + key);
 
-            if (finalKey.StartsWith(".."))
+            if (finalKey.StartsWith(string.Empty + _keySeparator + _keySeparator))
```
DavidKarlas, line 174: nit: Can we find better syntax? :)

src/Microsoft.TemplateEngine.Edge/Settings/TemplateCache.cs
```diff
@@ -13,9 +14,11 @@ namespace Microsoft.TemplateEngine.Edge.Settings
...
     internal class TemplateCache
     {
-        // locale -> identity -> locator
-        private readonly IDictionary<string, IDictionary<string, ILocalizationLocator>> _localizationMemoryCache = new Dictionary<string, IDictionary<string, ILocalizationLocator>>();
-
+        /// <summary>
+        /// Contains localizations for templates and locales.
+        /// Letters in locale are always lowercase.
+        /// </summary>
+        private readonly Dictionary<(string templateId, string locale), ILocalizationLocator> _localizationMemoryCache = new ();
```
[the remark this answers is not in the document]
bozturkMSFT: Private fields start with lowercase, analyzers issue errors if I use uppercase.

Updated the dictionary to ignore casing.

src/Microsoft.TemplateEngine.Edge/Settings/TemplateCache.cs
```diff
@@ -366,13 +338,27 @@ private void AddTemplateToMemoryCache(ITemplate template)
 
         private void AddLocalizationToMemoryCache(ILocalizationLocator locator)
         {
-            if (!_localizationMemoryCache.TryGetValue(locator.Locale, out IDictionary<string, ILocalizationLocator> localeLocators))
+            string lowercaseLocale = locator.Locale.ToLowerInvariant();
```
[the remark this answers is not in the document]
bozturkMSFT: Extracted to a new method.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/Localization/LocalizationModel.cs
```diff
@@ -9,19 +11,33 @@ namespace Microsoft.TemplateEngine.Orchestrator.RunnableProjects.Localization
...
-        public string Name { get; set; }
+        public string? Name { get; set; }
 
-        public string Description { get; set; }
+        public string? Description { get; set; }
 
-        // Identifies the langpack association with an actual template. They'll both have the same identity.
-        // This is not localized info, more like a key
-        public string Identity { get; set; }
+        public string? Identity { get; set; }
```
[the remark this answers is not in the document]
bozturkMSFT: I forgot to remove this property. It's gone now.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/Localization/LocalizationModel.cs
```diff
@@ -9,19 +11,33 @@ namespace Microsoft.TemplateEngine.Orchestrator.RunnableProjects.Localization
...
 
         public IReadOnlyDictionary<string, IParameterSymbolLocalizationModel> ParameterSymbols { get; set; }
```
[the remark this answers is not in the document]
bozturkMSFT: Made the type immutable.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/ManualInstructionModel.cs
```diff
@@ -0,0 +1,37 @@
...
+
+#nullable enable
+
+namespace Microsoft.TemplateEngine.Orchestrator.RunnableProjects
+{
+    /// <summary>
+    /// Represents an instruction that should be manually performed by the user
+    /// as part of a post action.
+    /// </summary>
+    public sealed class ManualInstructionModel
```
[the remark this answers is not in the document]
bozturkMSFT: Changed to internal.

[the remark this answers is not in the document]
bozturkMSFT: Added. Also did a bit of refactoring in the base implementation.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/PostActionModel.cs
bozturkMSFT: Yes. I just inlined the variable that wasn't used anywhere else:
https://github.com/dotnet/templating/pull/3089/files/bb688b93882c20f34ae00493bf2be4711bc1e221#diff-1efd527e44892bff897ea620e1e8b65fb4909ffd884d370a86c48f1e14a72c81L40

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/RunnableProjectGenerator.cs
bozturkMSFT: I think we should do that, but I want that to be a part of change where `string Locale`s are replaced with `CultureInfo Locale`s everywhere in the solution. Until we do that, I think this one is harmless.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/RunnableProjectGenerator.cs
bozturkMSFT: I removed `MountPointUri` as it wasn't used anywhere.
`ConfigPlace` doesn't point to primary json though. And it is used in the code to load localizations.

src/Microsoft.TemplateEngine.TemplateLocalizer.Core/KeyCreators/ChildValueKeyCreator.cs
bozturkMSFT: Good idea. Created an issue here https://github.com/dotnet/templating/issues/3127.

src/Microsoft.TemplateEngine.TemplateLocalizer.Core/TemplateStringExtractor.cs
bozturkMSFT: Updated with a much better alternative. Let me know what you think! I'm also open to suggestions :)

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/Localization/LocalizationModel.cs
```diff
@@ -9,19 +11,33 @@ namespace Microsoft.TemplateEngine.Orchestrator.RunnableProjects.Localization
...
+        public LocalizationModel() : this(
+            new Dictionary<string, IParameterSymbolLocalizationModel>(),
+            new List<IPostActionLocalizationModel>(),
+            new List<IFileLocalizationModel>())
+        { }
+
+        public LocalizationModel(
+            IReadOnlyDictionary<string, IParameterSymbolLocalizationModel> parameterSymbols,
+            IReadOnlyList<IPostActionLocalizationModel?> postActions,
+            IReadOnlyList<IFileLocalizationModel> fileLocalizations)
```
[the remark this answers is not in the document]
bozturkMSFT: Removed.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/LocalizationModelDeserializer.cs
bozturkMSFT: We throw an exception now, with proper error string. It will be caught and printed.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/SimpleConfigModel.cs
bozturkMSFT: Removed.

src/Microsoft.TemplateEngine.Abstractions/IFileLocalizationModel.cs
```diff
@@ -5,11 +5,22 @@
```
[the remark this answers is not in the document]
bozturkMSFT: File is now removed.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/LocalizationModelDeserializer.cs
bozturkMSFT: Yes, I think I understand right. I'm just not sure how that is better. I don't see the encapsulation benefit we are getting from moving this code in between private methods within the same class. And this is also at the expense of code duplication.
Regardless, I updated the methods to move forward.

DavidKarlas (approved):
This looks great!

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/RunnableProjectGenerator.cs
```diff
@@ -119,58 +119,37 @@ public IParameterSet GetParametersForTemplate(IEngineEnvironmentSettings environ
...
-                            Identity = locModel.Identity,
-                            Author = locModel.Author,
-                            Name = locModel.Name,
-                            Description = locModel.Description,
-                            ParameterSymbols = locModel.ParameterSymbols
-                        };
-                        localizations.Add(locator);
+                            string locale = locFile.Name.Substring(0, locFile.Name.Length - LocalizationFileSuffix.Length);
+
+                            if (TryGetLangPackFromFile(locFile, out ILocalizationModel locModel))
```
vlada-shubina, line 140: `TryGetLangPackFromFile` uses `SimpleConfigModel.LocalizationFromJObject(srcObject)` which doesn't have validation? Is it intentional? 
I remember we discussed having validation on installation, which seems not to be covered now. Will it be covered in next PR?

vlada-shubina (changes requested):
2 points worth considering:
- no validation on installation time (we should consider it at least as future task)
- my previous comment about tests still apply: https://github.com/dotnet/templating/pull/3089#pullrequestreview-648982913. Now that validation is added, we should consider tests for it as well (i'm ok to do it in a separate PR too).

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/RunnableProjectGenerator.cs
```diff
@@ -119,58 +119,37 @@ public IParameterSet GetParametersForTemplate(IEngineEnvironmentSettings environ
...
+                                localizations.Add(new LocalizationLocator(
+                                    locale,
+                                    locFile.FullPath,
+                                    template.Identity,
+                                    locModel.Author,
+                                    locModel.Name,
+                                    locModel.Description,
+                                    locModel.ParameterSymbols));
+                            }
+                        }
```
vlada-shubina, line 151: for reference only

vlada-shubina (approved):
Approved, given https://github.com/dotnet/templating/issues/3132

src/Microsoft.TemplateEngine.Abstractions/IPostAction.cs
bozturkMSFT: Even if it is only implemented by the CLI today, we have this _property_ in the interface under Abstractions assembly. As I understand it, that is the purpose of `ManualInstructions`. The comment only explains the situation that is true regardless. That's why I think it makes sense to have it here.

src/Microsoft.TemplateEngine.Abstractions/IPostActionLocalizationModel.cs
```diff
@@ -10,10 +12,13 @@ public interface IPostActionLocalizationModel
     {
         Guid ActionId { get; }
 
-        string Description { get; }
+        string? Description { get; }
```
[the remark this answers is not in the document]
bozturkMSFT: Added comments to description. Removed the action id since it is not needed in the localization model.

src/Microsoft.TemplateEngine.Core.Contracts/IConditionedConfigurationElement.cs
```diff
@@ -1,16 +1,33 @@
...
     {
-        string Condition { get; }
+        /// <summary>
+        /// Gets or sets the condition string to be evaluated.
+        /// </summary>
+        string? Condition { get; }
 
+        /// <summary>
+        /// Gets the cached result of the evaluation.
+        /// Make sure <see cref="EvaluateCondition(IEngineEnvironmentSettings, IVariableCollection)"/> is called
```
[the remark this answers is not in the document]
bozturkMSFT: We can't automatically call it in the property, because it requires arguments that we don't have access to. But I agree with you on how this looks bad: we are just trying to access the result without knowing if the values was really evaluated or not and we don't have the option to evaluate it right there because we don't have access to the required objects to pass them as arguments.

It sounds like it would be better to have a new type like `EvaluatedConditionedConfiguration` or something that we can pass around once we do the evaluation.

src/Microsoft.TemplateEngine.Edge/Settings/LocalizationCacheKeyComparer.cs
```diff
@@ -0,0 +1,28 @@
...
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.TemplateEngine.Edge.Settings
+{
+    /// <summary>
+    /// Makes an ordinal, case insensitive comparison with the two strings of the given tuples.
+    /// </summary>
+    internal class LocalizationCacheKeyComparer : IEqualityComparer<(string? Key, string? Value)>
```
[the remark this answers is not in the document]
bozturkMSFT: I agree. I only do it this way because it is "the correct way". I would like to understand why this is considered a good practice and whether we can ignore it in cases such as this one.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/LocalizationModelDeserializer.cs
```diff
@@ -0,0 +1,210 @@
...
+                .Where(s => s.Key.StartsWith(PostActionIndexPrefix))
+                .Select(s => (s.Key.Split(KeySeparator).AsEnumerable(), s.Value))
+                .ToList();
+
+            foreach (var postActionParts in strings.GroupBy(p => p.NameParts.FirstOrDefault()))
+            {
+                if (!GetIndexFromString(postActionParts.Key, PostActionIndexPrefix, PostActionIndexSuffix, out int postActionIndex))
+                {
+                    // Invalid index.
+                    continue;
```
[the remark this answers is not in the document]
bozturkMSFT: I'm removing these sections in the next PR. Stay tuned!

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/PostActionModel.cs
```diff
@@ -1,48 +1,65 @@
...
-        public IReadOnlyList<KeyValuePair<string, string>> ManualInstructionInfo { get; private set; }
+        public IReadOnlyList<ManualInstructionModel> ManualInstructionInfo { get; private set; }
 
-        public string ConfigFile { get; private set; }
+        public string? ConfigFile { get; private set; }
 
-        internal static IReadOnlyList<IPostActionModel> ListFromJArray(JArray jObject, IReadOnlyDictionary<Guid, IPostActionLocalizationModel> localizations)
+        internal static IReadOnlyList<IPostActionModel> ListFromJArray(JArray jObject, IReadOnlyDictionary<int, IPostActionLocalizationModel>? localizations, ITemplateEngineHost host)
         {
+            // TODO Host is only here to allow logging. Once ILogger is available, remove host from required parameters.
```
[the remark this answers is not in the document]
bozturkMSFT: Yes, I meant replacing the host with `ILogger`.

I think most (if not all) of the usages of `ILogger` shouldn't know that `ILogger` comes from the host. I think it should rather be a separate parameter and it should be up to the caller to pass whatever `ILogger` they want.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/RunnableProjectGenerator.cs
bozturkMSFT: We don't have validation at this point, because this part just extracts the values from the file. The validation happens later when we are comparing the template with the localization file prior to writing the cache.

This is probably a bit late for such validation, so I will add an earlier validation in the next PR with https://github.com/dotnet/templating/issues/3133.

src/Microsoft.TemplateEngine.Abstractions/IPostAction.cs
DavidKarlas: Imo, its VS bug to not inform user to do something manually if VS didn't do it for them... VS being Windows app, would show dialog...

src/Microsoft.TemplateEngine.Abstractions/IPostAction.cs
```diff
@@ -1,23 +1,44 @@
...
-        string ManualInstructions { get; }
+        /// <summary>
+        /// Gets the instructions that should be manually performed by the user
+        /// as part of this post action.
+        /// Manual instructions are used when the host does not have the associated post action processor implemented.
+        /// </summary>
+        string? ManualInstructions { get; }
 
-        string ConfigFile { get; }
+        string? ConfigFile { get; }
```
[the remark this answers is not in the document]
bozturkMSFT: This doesn't appear to be used, so I removed it all together (internal team discussion also didn't yield concrete results on what this was for). You did ask me if this could be removed, right? For some reason, I had found some ties that prevented me from deleting it. Anyway, its gone now.

Issue #3133, opened 2021-05-05
Localization file format shouldn't use indices


bekir-ozturk:
Post actions and manual instructions are identified by their indices in templatestrings.json files today. Keeping the order of the localized strings are quite difficult when a new post action or manual instruction is added. We have decided to switch to identifiers.

- [x] Update post action localization model to contain an optional string property "id"
- [x] Update manual instruction localization model to contain an optional string property "id"
- [x] Update template localizer to use Ids for post actions and manual instructions instead of indices.
- [x] Update template localizer to automatically set the id of a manual instruction to "default", in the case that there is only one manual instruction defined in the owning post action.
- [x] Update `TemplateLocalizationDeserializer` to read post actions and manual instructions using Ids instead of indices.
- [x] Update template engine logic where a TemplateModel is merged with a TemplateLocalizationModel to make sure post actions and manual instructions are matched in these two types by their Ids and not by their indices.

The following task was moved to task #3132 
- [ ] Add a validation step right after localization deserialization and print errors if validation fails.

Issue #3132, opened 2021-05-05
Validate deserialization of localization files during install and instantiate


bekir-ozturk:
Follow up task to https://github.com/dotnet/templating/issues/3133

- [x] Add a validation step right after localization deserialization and print errors if validation fails.
- [x] Make sure that an error is printed to the screen when user is installing/instantiating a template with broken localization files using CLI.
- [x] Make sure that no error is displayed when user is installing/instantiating a template with valid localization files using CLI.

PR #3134, opened 2021-05-06
PostAction and ManualInstruction localizations shouldn't use indices

bekir-ozturk:
### Problem
https://github.com/dotnet/templating/issues/3133

### Solution
See the issue for planned/delivered changes.

### Checks:
- [x] Added unit tests

### DIFF
--- a/src/Microsoft.TemplateEngine.Abstractions/ILocalizationModel.cs
+++ b/src/Microsoft.TemplateEngine.Abstractions/ILocalizationModel.cs
@@ -34,8 +34,8 @@ namespace Microsoft.TemplateEngine.Abstractions
 
         /// <summary>
         /// Gets the localization models for the post actions defined in this template.
-        /// The keys represend the indices of the post actions as they appear in the template config file.
+        /// The keys represent the id of the post actions.
         /// </summary>
-        IReadOnlyDictionary<int, IPostActionLocalizationModel> PostActions { get; }
+        IReadOnlyDictionary<string, IPostActionLocalizationModel> PostActions { get; }
     }
 }
--- a/src/Microsoft.TemplateEngine.Abstractions/PublicAPI.Shipped.txt
+++ b/src/Microsoft.TemplateEngine.Abstractions/PublicAPI.Shipped.txt
@@ -64,7 +64,6 @@ Microsoft.TemplateEngine.Abstractions.ILocalizationModel.Author.get -> string?
 Microsoft.TemplateEngine.Abstractions.ILocalizationModel.Description.get -> string?
 Microsoft.TemplateEngine.Abstractions.ILocalizationModel.Name.get -> string?
 Microsoft.TemplateEngine.Abstractions.ILocalizationModel.ParameterSymbols.get -> System.Collections.Generic.IReadOnlyDictionary<string!, Microsoft.TemplateEngine.Abstractions.IParameterSymbolLocalizationModel!>!
-Microsoft.TemplateEngine.Abstractions.ILocalizationModel.PostActions.get -> System.Collections.Generic.IReadOnlyDictionary<int, Microsoft.TemplateEngine.Abstractions.IPostActionLocalizationModel!>!
 Microsoft.TemplateEngine.Abstractions.Installer.CheckUpdateResult
 Microsoft.TemplateEngine.Abstractions.Installer.CheckUpdateResult.IsLatestVersion.get -> bool
 Microsoft.TemplateEngine.Abstractions.Installer.IInstaller
@@ -116,9 +115,6 @@ Microsoft.TemplateEngine.Abstractions.IPostAction.Description.get -> string?
 Microsoft.TemplateEngine.Abstractions.IPostAction.ManualInstructions.get -> string?
 Microsoft.TemplateEngine.Abstractions.IPostActionLocalizationModel
 Microsoft.TemplateEngine.Abstractions.IPostActionLocalizationModel.Description.get -> string?
-Microsoft.TemplateEngine.Abstractions.IPostActionLocalizationModel.Instructions.get -> System.Collections.Generic.IReadOnlyDictionary<int, string!>!
-Microsoft.TemplateEngine.Abstractions.IPrioritizedComponent
-Microsoft.TemplateEngine.Abstractions.IPrioritizedComponent.Priority.get -> int
 Microsoft.TemplateEngine.Abstractions.ISearchPackFilter
 Microsoft.TemplateEngine.Abstractions.ISettingsLoader
 Microsoft.TemplateEngine.Abstractions.ISettingsLoader.Save() -> void
--- a/src/Microsoft.TemplateEngine.Abstractions/PublicAPI.Shipped.txt
+++ b/src/Microsoft.TemplateEngine.Abstractions/PublicAPI.Shipped.txt
@@ -108,6 +108,8 @@ Microsoft.TemplateEngine.Abstractions.IPostAction.Args.get -> System.Collections
 Microsoft.TemplateEngine.Abstractions.IPostAction.ContinueOnError.get -> bool
 Microsoft.TemplateEngine.Abstractions.IPostAction.Description.get -> string?
 Microsoft.TemplateEngine.Abstractions.IPostAction.ManualInstructions.get -> string?
+Microsoft.TemplateEngine.Abstractions.IPrioritizedComponent
+Microsoft.TemplateEngine.Abstractions.IPrioritizedComponent.Priority.get -> int
 Microsoft.TemplateEngine.Abstractions.ISearchPackFilter
 Microsoft.TemplateEngine.Abstractions.ISettingsLoader
 Microsoft.TemplateEngine.Abstractions.ISettingsLoader.Save() -> void
--- a/src/Microsoft.TemplateEngine.Abstractions/PublicAPI.Shipped.txt
+++ b/src/Microsoft.TemplateEngine.Abstractions/PublicAPI.Shipped.txt
@@ -59,11 +59,6 @@ Microsoft.TemplateEngine.Abstractions.IGenerator
 Microsoft.TemplateEngine.Abstractions.IIdentifiedComponent
 Microsoft.TemplateEngine.Abstractions.IIdentifiedComponent.Id.get -> System.Guid
 Microsoft.TemplateEngine.Abstractions.ILocalizationLocator
-Microsoft.TemplateEngine.Abstractions.ILocalizationModel
-Microsoft.TemplateEngine.Abstractions.ILocalizationModel.Author.get -> string?
-Microsoft.TemplateEngine.Abstractions.ILocalizationModel.Description.get -> string?
-Microsoft.TemplateEngine.Abstractions.ILocalizationModel.Name.get -> string?
-Microsoft.TemplateEngine.Abstractions.ILocalizationModel.ParameterSymbols.get -> System.Collections.Generic.IReadOnlyDictionary<string!, Microsoft.TemplateEngine.Abstractions.IParameterSymbolLocalizationModel!>!
 Microsoft.TemplateEngine.Abstractions.Installer.CheckUpdateResult
 Microsoft.TemplateEngine.Abstractions.Installer.CheckUpdateResult.IsLatestVersion.get -> bool
 Microsoft.TemplateEngine.Abstractions.Installer.IInstaller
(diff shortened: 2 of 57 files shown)


src/Microsoft.TemplateEngine.Abstractions/ILocalizationModel.cs
```diff
@@ -34,8 +34,8 @@ public interface ILocalizationModel
 
         /// <summary>
         /// Gets the localization models for the post actions defined in this template.
-        /// The keys represend the indices of the post actions as they appear in the template config file.
+        /// The keys represent the id of the post actions.
         /// </summary>
-        IReadOnlyDictionary<int, IPostActionLocalizationModel> PostActions { get; }
+        IReadOnlyDictionary<string, IPostActionLocalizationModel> PostActions { get; }
```
vlada-shubina, line 39: 1) this is a public API change that we are trying to avoid
2) `ILocalizationModel` is interface internal to RunnableProjects. Should we move it (and `IPostActionLocalizationModel`) to `RunnableProjects` and make internal?

src/Microsoft.TemplateEngine.TemplateLocalizer/LocalizableStrings.resx
```diff
@@ -144,7 +144,12 @@
     <comment>{0} is the number of files. Do not localize "--export"</comment>
   </data>
   <data name="command_export_log_templateExportFailedWithError" xml:space="preserve">
-    <value>Generating localization files for the following template.json file has failed: "{0}". Error message: "{1}".</value>
+    <value>Generating localization files for a template.json has failed.
+Reason: {0}.
+File Path: {1}.</value>
+    <comment>{0} is a sentence such as "Json property should contain a field 'ID'"
```
vlada-shubina, line 150: previously we discussed, that if 'id' is missing it would be generated by localizer. Have we dropped this feature?

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/PostActionModel.cs
```diff
@@ -30,93 +32,105 @@ internal class PostActionModel : ConditionedConfigurationElementBase, IPostActio
...
             }
 
-            if (localizations?.Count > localizedPostActions)
+            if (localizations?.Count > localizedPostActionCount)
             {
                 // Localizations provide more translations than the number of post actions we have.
-                string excessPostActionLocalizationIndexes = string.Join(", ", localizations.Keys.Where(k => k < 0 || k > modelList.Count).Select(k => k.ToString()));
-                host.Logger.LogWarning(string.Format(LocalizableStrings.Authoring_InvalidPostActionLocalizationIndex, excessPostActionLocalizationIndexes));
+                string excessPostActionLocalizationIds = string.Join(", ", localizations.Keys.Where(k => !localizedPostActions.Any(p => p.Id == k)).Select(k => k.ToString()));
+                host.Logger.LogWarning(string.Format(LocalizableStrings.Authoring_InvalidPostActionLocalizationIndex, excessPostActionLocalizationIds));
```
vlada-shubina, line 131: do we have integration tests for these warnings? If not, it would be good to add ones.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/PostActionModel.cs
```diff
@@ -30,93 +32,105 @@ internal class PostActionModel : ConditionedConfigurationElementBase, IPostActio
...
                         {
                             text = manualInstructions[i].ToString("text");
                         }
 
-                        instructionOptions.Add(new ManualInstructionModel(text ?? string.Empty, manualInstructions[i].ToString("condition")));
+                        localizedInstructions.Add(new ManualInstructionModel(id, text ?? string.Empty, manualInstructions[i].ToString("condition")));
                     }
 
-                    if (actionLocalizations?.Instructions.Count > localizedManualInstructions)
+                    if (actionLocalizations?.Instructions.Count > localizedInstructionCount)
```
vlada-shubina, line 104: we should also check if `id`s  are unique, otherwise this check might be false positive. 
The following checks should be sufficient:
- `id`s  in post actions and in localization are unique
- `id`s in localization are subset of post action `id`s

vlada-shubina (changes requested):
please:
- review API changes
- nullability of `id` field
- validation of post action localizations (seems not all cases are covered)

src/Microsoft.TemplateEngine.Abstractions/ILocalizationModel.cs
bekir-ozturk: If it is internal to Runnable projects, sounds like a good idea. I will do it.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/ManualInstructionModel.cs
```diff
@@ -11,17 +11,24 @@ namespace Microsoft.TemplateEngine.Orchestrator.RunnableProjects
     /// </summary>
     internal sealed class ManualInstructionModel : ConditionedConfigurationElementBase
     {
-        public ManualInstructionModel(string text)
+        public ManualInstructionModel(string? id, string text)
```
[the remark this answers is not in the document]
bekir-ozturk: It is nullable, because user doesn't have to specify an id when there is only one manual instruction.
Technically, whitespace or empty string is also fine for us. Basically, as long as they are unique we don't care much about the content of the string.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/PostAction.cs
```diff
@@ -11,25 +13,32 @@ namespace Microsoft.TemplateEngine.Orchestrator.RunnableProjects
...
-        Guid IPostAction.ActionId => ActionId;
-
-        bool IPostAction.ContinueOnError => ContinueOnError;
-
-        IReadOnlyDictionary<string, string> IPostAction.Args => Args;
-
-        string IPostAction.ManualInstructions => ManualInstructions;
+        public PostAction(string? id, string? description, string? manualInstructions, Guid actionId, bool continueOnError, IReadOnlyDictionary<string, string> args)
+        {
+            Id = id;
```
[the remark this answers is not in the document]
bekir-ozturk: It looks like not needed. If it becomes necessary in the future, we can add it later. Removing now.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/PostActionModel.cs
```diff
@@ -30,93 +32,105 @@ internal class PostActionModel : ConditionedConfigurationElementBase, IPostActio
...
                 if (manualInstructions != null)
                 {
-                    int localizedManualInstructions = 0;
+                    int localizedInstructionCount = 0;
                     for (int i = 0; i < manualInstructions.Count; i++)
                     {
+                        string? id = manualInstructions[i].ToString("id");
                         string? text = string.Empty;
-                        if (actionLocalizations?.Instructions.TryGetValue(i, out text) ?? false)
+                        string? localizationKey = id ?? (manualInstructions.Count == 0 ? "default" : null);
```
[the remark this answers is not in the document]
bekir-ozturk: If there is only one manual instruction, user doesn't have to specify an id. We use "default" key to lookup the localized string.

src/Microsoft.TemplateEngine.TemplateLocalizer/LocalizableStrings.resx
bekir-ozturk: Are you sure about the discussions? Localizer doesn't make changes to the template.json file. And I don't recall that we came to such a conclusion.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/PostActionModel.cs
bekir-ozturk: We have that planned with https://github.com/dotnet/templating/issues/3132

src/Microsoft.TemplateEngine.Abstractions/PublicAPI.Shipped.txt
```diff
@@ -117,9 +116,6 @@ Microsoft.TemplateEngine.Abstractions.IPostAction.Description.get -> string?
 Microsoft.TemplateEngine.Abstractions.IPostAction.ManualInstructions.get -> string?
 Microsoft.TemplateEngine.Abstractions.IPostActionLocalizationModel
 Microsoft.TemplateEngine.Abstractions.IPostActionLocalizationModel.Description.get -> string?
-Microsoft.TemplateEngine.Abstractions.IPostActionLocalizationModel.Instructions.get -> System.Collections.Generic.IReadOnlyDictionary<int, string!>!
-Microsoft.TemplateEngine.Abstractions.IPrioritizedComponent
```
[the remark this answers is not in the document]
bekir-ozturk: Would there be a CI error if this file was wrong?
I just noticed the error during build and used "Auto Fix for solution".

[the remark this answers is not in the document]
vlada-shubina: I think it might be a bug in analyzer then. 
It removed this line from `shipped` and put it to `unshipped`, though there is no change to the actual file.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/ManualInstructionModel.cs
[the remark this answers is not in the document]
vlada-shubina: okay, it's reasonable. The comment on the property will be helpful, to state the case when it's null.

src/Microsoft.TemplateEngine.TemplateLocalizer/LocalizableStrings.resx
vlada-shubina: Maybe I misunderstood it. 
I thought that if ids are not present the tool will auto-generates them and adds them to `template.json` and `templatestrings`. Imo this will be helpful, as the author won't need to create ids manually. Also it will allow to cover cases of multiple entries, as we can do ids similar to "autogen_1", "autogen_2".

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/PostActionModel.cs
bekir-ozturk: Added checks for uniqueness of post action and manual instruction IDs. Uniqueness of localizations are already ensured while deserializing the JSON.
I used a counter here instead of checking for subsets for performance reasons. If a problem is found, only then the two sets are compared to determine the exact cause of the error.

src/Microsoft.TemplateEngine.Abstractions/PublicAPI.Shipped.txt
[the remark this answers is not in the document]
bekir-ozturk: I think I understand the problem. I accidentally removed it while removing the others. Then "Auto Fix" automatically added into the "Unshipped" file.
Could it be that we should first put API changes to Unshipped file and at the end of a release, we move them all into the "Shipped" file? I'm not sure if that is the reason why we have both shipped and unshipped files, but this could be used to easily what changes are being made to the API in that particular release.
Today, all of our unshipped files are empty and every change is put into shipped files.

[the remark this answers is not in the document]
vlada-shubina: That's correct workflow:
- changing existing API (breaking change) will modify shipped file
- introducing new API (not a breaking change) will add it to unshipped
Once we ship the release, we copy over unshipped -> shipped, and changing it since that moment is a breaking change.

src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/PostActionModel.cs
vlada-shubina: Re: dedup in deserializing: we are not giving error on having entries with same ID, it would be good at least do debug level on it.

vlada-shubina (approved):
I'm not in favor of design decision of doing implicit IDs for a very specific case (1 post action with 1 manual instruction), which is likely not common.

https://github.com/dotnet/templating/pull/3134/files?file-filters%5B%5D=.cs&file-filters%5B%5D=.json&file-filters%5B%5D=.resx&file-filters%5B%5D=.txt#r638646017

I would prefer either to cover all the cases as described above, or demand manual ID entry for all the cases, perhaps we can have .NET 7/icebox issue on improving it.

PR #3149, opened 2021-05-12
Fixes #3085: dotnet new crashes if template contains #ifdef

### LINKED ISSUE #3085: dotnet new crashes if template contains #ifdef
Hi,

I'm experiencing a crash of `dotnet new` if my template contains an `#ifdef`.

Repro steps:
- unzip: [example.zip](https://github.com/dotnet/sdk/files/6325891/example.zip) 
- dotnet new -i .
- dotnet new Foo  (in a different directory)

I get this error:
```
Template "Foo" could not be created.
Error while processing file /foo.cpp
System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')
   at System.Collections.Generic.List`1.get_Item(Int32 index)
   at Microsoft.TemplateEngine.Core.Expressions.Cpp.CppStyleEvaluatorDefinition.Evaluate(IProcessorState processor, Int32& bufferLength, Int32& currentBufferPosition, Boolean& faulted)
   at Microsoft.TemplateEngine.Core.Operations.Conditional.Impl.HandleMatch(IProcessorState processor, Int32 bufferLength, Int32& currentBufferPosition, Int32 token, Stream target)
   at Microsoft.TemplateEngine.Core.Util.ProcessorState.Run()
   at Microsoft.TemplateEngine.Core.Util.Processor.Run(Stream source, Stream target, Int32 bufferSize, Int32 flushThreshold)
   at Microsoft.TemplateEngine.Core.Util.Processor.Run(Stream source, Stream target)
   at Microsoft.TemplateEngine.Core.Util.Orchestrator.ProcessFile(IFile sourceFile, String sourceRel, String targetDir, IGlobalRunSpec spec, IProcessor fallback, IEnumerable`1 fileGlobProcessors, IReadOnlyList`1 locOperations)

```

I have noticed the problem disappear if a 'space' is inserted after the # in the ifdef.
It also disappears if the template contains the file `bar.cpp` that contains `//-:cnd:noEmit`

I see the problem on Windows with .NET SDK (5.0.201). It doesn't appear to happen on Linux.

BR
  Lars

DavidKarlas:
### Problem
Problem was that in `if` statement above took path where `tokens` doesn't add any token, reason for that is that there was no `whitespace` or `brace` token after it, instead just "Literal" which is not fully parsed yet and can't be added to `tokens`, at least that's my understanding...
What is weird about current logic is... that `#ifBAR` is same as `#if BAR`... Easy fix would be to do add space after `if`, but might cause other issues...

### Solution
I think lets not invest too much enery into this and lets just fix crash if user types `#iffoo bar` so it doesn't crash...

### Checks:
- [x] Added unit tests

### DIFF
--- a/src/Microsoft.TemplateEngine.Core/Expressions/Cpp/CppStyleEvaluatorDefinition.cs
+++ b/src/Microsoft.TemplateEngine.Core/Expressions/Cpp/CppStyleEvaluatorDefinition.cs
@@ -114,7 +114,7 @@ namespace Microsoft.TemplateEngine.Core.Expressions.Cpp
             }
 
             int braceDepth = 0;
-            if (tokens[0].Family == TokenFamily.OpenBrace)
+            if (tokens.Count > 0 && tokens[0].Family == TokenFamily.OpenBrace)
             {
                 ++braceDepth;
             }
--- a/test/Microsoft.TemplateEngine.Core.UnitTests/ConditionalTests.CStyleEvaluator.cs
+++ b/test/Microsoft.TemplateEngine.Core.UnitTests/ConditionalTests.CStyleEvaluator.cs
@@ -1596,5 +1596,34 @@ There";
             //  pretend it's false because the inputs and outputs are the same
             Verify(Encoding.UTF8, output, false, value, expected);
         }
+
+        [Theory]
+        [InlineData("foo", true)]
+        [InlineData("foo", false)]
+        [InlineData("def", true)]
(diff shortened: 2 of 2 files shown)


vlada-shubina (approved):
This fixes the problem of #3085, however I suggest we open the long term bug to fix whitespace/delimiter absence for condition inside if-else-endif statement.

bekir-ozturk (approved):

PR #3184, opened 2021-05-18
First step in making build happen automatically after new

Forgind:
This should hopefully make your first actual `dotnet build` call faster. More changes to follow, but I wanted to get something out so you can see where I'm going.

/cc: @bekir-ozturk


src/Microsoft.TemplateEngine.Cli/PostActionProcessors/DotnetBuildPostActionProcessor.cs
```diff
@@ -0,0 +1,141 @@
...
+                        && !Path.GetExtension(pathToRestore).Equals(".sln", StringComparison.OrdinalIgnoreCase)
+                    ))
+                {
+                    continue;
+                }
+
+                Reporter.Output.WriteLine(string.Format(LocalizableStrings.RunningDotnetRestoreOn, pathToRestore));
+
+                // Prefer to restore the project in-proc vs. creating a new process.
+                bool succeeded = Callbacks.RestoreProject == null ?
```
ladipro, line 79: Do I understand it correctly that we don't want to have a similar callback for Build because starting a new dotnet process is by design here? We want to start it, return to the command line, and let it finish "in the background", is that correct?

ladipro (commented):
Is there a document you can point us to to understand the design? There are many questions around this approach. For example, what happens if the user launches build explicitly before this implicit one finishes. Or the interaction with Restore which is currently happening synchronously.

src/Microsoft.TemplateEngine.Cli/PostActionProcessors/DotnetBuildPostActionProcessor.cs
Forgind: https://github.com/ladipro/one-pagers/pull/18/files is what I'm looking at.

My first goal is to get it to build after running restore. Afterwards, I'm planning to consider what happens if something goes wrong (as with the user calling it explicitly) or just replace it with a conglomeration of some of the other ideas here.

I am intentionally starting up a new process, yes.

bekir-ozturk:
Hey @Forgind,
 Is this still on the radar? Can we close it if not?

Forgind:
It is, just not very high up it currently. We can probably close it for now, and I can reopen it when I get back to it.

PR #3207, opened 2021-05-25
Fixes #2742: Cache HostSpecificDataLoader.ReadHostSpecificTemplateData

### LINKED ISSUE #2742: Cache HostSpecificDataLoader.ReadHostSpecificTemplateData
During resolution of templates Microsoft.TemplateEngine.Cli.HostSpecificDataLoader.ReadHostSpecificTemplateData is called which is slows since it needs to open zip files to check host .json file... Find a way to cache this data in templateCache

DavidKarlas:
### Problem
During resolution of templates Microsoft.TemplateEngine.Cli.HostSpecificDataLoader.ReadHostSpecificTemplateData is called which is slows since it needs to open zip files to check host .json file...

### Solution
Turns out majority of problem was not unzipping but usage of .ToObject API, but also the fact that we re-read same .json multiple times, hence introduced caching.

### Checks:
- [ ] Added unit tests


DavidKarlas:
I need to add tests for host config

PR #3637, opened 2021-08-11
Fixes 3224: Add rollForward option to global.json generated by 'dotnet new globaljson'

DavidKarlas:
### Problem
Today, if user wants to set "rollForward" value in global.json file it needs to open it manually and edit global.json file...

### Solution
Give --roll-forward parameter option to do that.

### Checks:
- [x] Added unit tests


bekir-ozturk (approved):

jesuscorral:
I'm trying  _dotnet new globaljson --sdk-version 5.0.204 --roll-forward latestFeature_ and for the first time that I execute this command works ok, but if I execute again with and existing global.json file the command return this error 
image

I've tried using --force parameter with the same result. Are there any way to avoid this error?

PR #3818, opened 2021-08-31
Rename templates: Application to App

### LINKED ISSUE #3791: Change branding from 'Application' to 'App'
There was an intentional branding change in Visual Studio 2019 to rebrand templates from called 'Application' to 'App'. Now that we are picking up templates from the SDK there is an inconsistency:

image

We should rename the template metadate here: https://github.com/dotnet/templating/blob/7b8532826f7fbb96d63ad78bd74e5353656049e5/template_feed/Microsoft.DotNet.Common.ProjectTemplates.6.0/content/ConsoleApplication-CSharp/.template.config/template.json#L8

to `Console App`

cc: @KathleenDollard

bekir-ozturk:
### Problem
Fixes #3791 

### Solution

Renames the templates.

### Checks:
- [ ] Added unit tests


vlada-shubina (approved):
Not sure, if we need to change the name for all templates, not only for net6.0 ones.

Not related to task itself: unpleasantly surprised on number of dependencies in tests, probably we should revise it at some point.

DavidKarlas (approved):

PR #4137, opened 2021-11-23
Fix manual instruction actionId

aokellermann:
### Problem
Documentation inconsistent.

### Solution
`grep -rnw "AC1156F7-BB77-4DB8-B28F-24EEBCCA1E54"` yields only one instance in documentation.

Change documentation to be consistent.


vlada-shubina (approved):
Thank you for contributing!

PR #4318, opened 2022-01-27
Preserve the UTF8 BOM in localized files

ladipro:
### Problem
Template Localizer removes the UTF8 BOM from localized files. This shows as an extra annoying diff hunk.

### Solution
Made the code preserve the BOM in the target file. We still don't add it to files that don't already have it, consistent with the Unicode Standard recommendation.

### Checks:
- [x] Added unit tests

### DIFF
--- a/src/Microsoft.TemplateEngine.TemplateLocalizer.Core/TemplateStringUpdater.cs
+++ b/src/Microsoft.TemplateEngine.TemplateLocalizer.Core/TemplateStringUpdater.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text;
 using System.Text.Encodings.Web;
 using System.Text.Json;
 using System.Text.Unicode;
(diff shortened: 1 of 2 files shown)


vlada-shubina (approved):
Thanks!

src/Microsoft.TemplateEngine.TemplateLocalizer.Core/TemplateStringUpdater.cs
```diff
@@ -157,6 +165,28 @@ await SaveTemplateStringsFileAsync(strings, existingStrings, locFilePath, forceU
...
+        /// <returns>A task that represents the asynchronous operation.</returns>
+        private static async Task TruncateFileWhilePreservingBom(FileStream fileStream, CancellationToken cancellationToken)
+        {
+            byte[] preamble = new byte[Utf8Bom.Length];
+            int offset = 0, read = 0;
+            // Read bytes from the stream until we fill the preamble array or hit EOF.
+            do
+            {
+                read = await fileStream.ReadAsync(preamble, offset, preamble.Length - offset, cancellationToken).ConfigureAwait(false);
+                offset += read;
```
DavidKarlas, line 183:
```suggestion
                offset += read;
                // Optimization to not call .ReadAsync twice
                if (offset == preamble.Length)
                {
                    break;
                }
```

DavidKarlas (approved):
Looks good, suggested small optimization, since 99% of time it will be read in one go

src/Microsoft.TemplateEngine.TemplateLocalizer.Core/TemplateStringUpdater.cs
ladipro: Good point! Committed.

PR #4360, opened 2022-02-04
Fix Search CI

DavidKarlas:
There were 2 problems, with recent change of CI to "more safe":
  * one, that one doesn't have `dotnet` installed hence we failed to evaluate value of MSBuild property, now we execute that step after we install `dotnet` tool...
  * second problem was that this CI also didn't have `az` tool installed, hence added that step too...


vlada-shubina (approved):
looks good to me

PR #4824, opened 2022-06-08
Remove Windows Phone example option from docs

baronfel:
It was pointed out to me that this was confusing and could send the message that we supported windows phone in any way, so let's just remove this to sidestep any confusion.


vlada-shubina (approved):

JanKrivanek (approved):
Good catch and thanks for fixing!

PR #5041, opened 2022-07-29
configure and use Verify.DiffPlex to generate deltas for failing tests

baronfel:
### Problem
It can be hard to see the deltas on our large text verifications in the current output format.

### Solution
This configures Verify.DiffPlex so that it uses a git-style +/- display to show the deltas so that it's easier to see the exact problem.

### Checks:
- [ ] Added unit tests


baronfel:
The Verify.DiffPlex and DiffPlex packages have been added to the dotnet-public feed, it should just be a matter of waiting for the feed caches to update.

vlada-shubina (approved):
Looks good, thanks

JanKrivanek (approved):
Thank you!!

PR #5301, opened 2022-09-23
Fix dependabot does not update package versions

GangWang01:
### Problem
Dependabot doesn't update package version with latest one. From its check log, the latest version is found, but no update needed for the old version. For example, NuGet.Configuration in the check https://github.com/dotnet/templating/network/updates/464268635 

### Solution
Dependabot is not happy with the section overriding package versions for source build. Without that part, Dependabot could work well in my test repo. I guess the version property causes the problem. Hence move the section overriding package versions for source build to `eng/Package.props`

### Checks:
- [ ] Added unit tests

### DIFF
--- a/eng/Package.props
+++ b/eng/Package.props
@@ -1,6 +1,36 @@
 <Project>
   <!-- Import references updated by Dependabot. This file is for package references updated manually or by Darc/Maestro. -->
   <Import Project="dependabot\Packages.props" />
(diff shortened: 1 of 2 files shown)


JanKrivanek (approved):

GangWang01:
msbuild has the same problem with https://github.com/dotnet/msbuild/pull/7798.

PR #5338, opened 2022-09-28
Add named mutex to packing TestTemplates project that is used by multiple tests running in parallel

GangWang01:
### Problem
#5286

### Solution
Add named mutex to prevent concurrent attempts to create same pack for Microsoft.TemplateEngine.TestTemplates.

### Checks:
- [ ] Added unit tests

### DIFF
--- a/test/Microsoft.TemplateEngine.TestHelper/PackageManager.cs
+++ b/test/Microsoft.TemplateEngine.TestHelper/PackageManager.cs
@@ -23,6 +23,9 @@ namespace Microsoft.TemplateEngine.TestHelper
         private static readonly SemaphoreSlim Semaphore = new SemaphoreSlim(1, 1);
         private readonly string _packageLocation = TestUtils.CreateTemporaryFolder("packages");
         private readonly ConcurrentDictionary<string, string> _installedPackages = new ConcurrentDictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+#pragma warning disable IDE0044 // Add readonly modifier
+        private volatile Mutex _packMutex = new Mutex(false, "PackMutex");
+#pragma warning restore IDE0044 // Add readonly modifier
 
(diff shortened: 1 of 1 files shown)


test/Microsoft.TemplateEngine.TestHelper/PackageManager.cs
```diff
@@ -118,11 +119,13 @@ public string PackNuGetPackage(string projectPath)
                         $"{Environment.NewLine}StdErr: {stdErr}.");
```
JanKrivanek, line 119: We should add info about the exit code here as well - I can see cases where pack seemingly succeeded, but returned nonzero exit code - e.g.: https://dev.azure.com/dnceng-public/public/_build/results?buildId=34308&view=ms.vss-test-web.build-test-results-tab&runId=683860&resultId=100021&paneView=debug

More info will be helpful in such a case. It might be related to unresolved issue: https://github.com/dotnet/sdk/issues/12298

test/Microsoft.TemplateEngine.TestHelper/PackageManager.cs
```diff
@@ -83,13 +86,11 @@ public string PackNuGetPackage(string projectPath)
...
                 throw new ArgumentException($"{projectPath} doesn't exist", nameof(projectPath));
             }
-            lock (string.Intern(absolutePath.ToLowerInvariant()))
-            {
-                if (_installedPackages.TryGetValue(absolutePath, out string? packagePath))
-                {
-                    return packagePath;
-                }
 
+            _packMutex.WaitOne();
```
JanKrivanek, line 90: `AbandonedMutexException` should be handled here (just logging and swollowing should be enough). Releasing below might be guarded in `finally` block as well (but handling abandoned named mutex is more important) - otherwise single failed run can cause failures in all other tests calling into this helper until the machine or container is recycled.

JanKrivanek (changes requested):
Thanks for the change!
I have couple of requests for improvements.

JanKrivanek (approved):
Looks good!
Thank you!

Issue #5288, opened 2022-09-21
Infra: Fix OneLocBuild parameters conditional logic


JanKrivanek:
## Background

OneLocBuild was not properly running in release/7.0.1xx. This was fixed by fixing the conditional logic (https://github.com/dotnet/templating/pull/5266) and hardcoding the parameters (https://github.com/dotnet/templating/commit/674c1378666fa57ced403bafd708a01b0647d7ad). The second change is temporary workaround and should be fixed befor merging back to main

## Proposal

Investigate and fix the OneLocBuild  conditional parameters logic (https://github.com/dotnet/templating/blob/release/7.0.1xx/azure-pipelines.yml#L48-L55). Merge the fix back to main

PR #5340, opened 2022-09-28
Output passed parameters/variables for investigating odd CI issue

GangWang01:
### Problem
OneLocBuild for release/7.0.1xx has to force passing the parameter while the conditional parameter logic doesn't get the parameter passed to the job.

### Solution
For investigation output the values of the parameters/variables without force passing the parameters.

### Checks:
- [ ] Added unit tests

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -45,8 +45,6 @@ stages:
       parameters:
         MirrorRepo: templating
         LclSource: lclFilesfromPackage
-        MirrorBranch: $(OneLocBuildBranch)
-        LclPackageId: 'LCL-JUNO-PROD-TEMPLATING'
         ${{ if eq(variables['Build.SourceBranch'], format('{0}{1}', 'refs/heads/', variables['OneLocBuildBranch'])) }}:
           MirrorBranch: $(OneLocBuildBranch)
           LclPackageId: 'LCL-JUNO-PROD-TEMPLATING'
--- a/eng/common/templates/job/onelocbuild.yml
+++ b/eng/common/templates/job/onelocbuild.yml
@@ -55,6 +55,10 @@ jobs:
       
 
   steps:
+    - script: echo $(Build.SourceBranch)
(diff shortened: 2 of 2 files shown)


azure-pipelines.yml
```diff
@@ -45,8 +45,6 @@ stages:
       parameters:
         MirrorRepo: templating
         LclSource: lclFilesfromPackage
-        MirrorBranch: $(OneLocBuildBranch)
-        LclPackageId: 'LCL-JUNO-PROD-TEMPLATING'
         ${{ if eq(variables['Build.SourceBranch'], format('{0}{1}', 'refs/heads/', variables['OneLocBuildBranch'])) }}:
           MirrorBranch: $(OneLocBuildBranch)
           LclPackageId: 'LCL-JUNO-PROD-TEMPLATING'
```
GangWang01, line 50: With some tries, I didn't figure out why https://github.com/dotnet/templating/pull/5266 didn't get `LclPackageId: 'LCL-JUNO-PROD-TEMPLATING'` passed to OneLocBuild successfully.

- First, the condition should be satisfied for release/7.0.1xx.
The change https://github.com/dotnet/templating/pull/5266 started the task OneLocBuild in internal run. It said `condition: or(eq(variables['Build.SourceBranch'], format('{0}{1}', 'refs/heads/', variables['OneLocBuildBranch'] )), eq(variables['Build.SourceBranch'], 'refs/heads/main'))` is true. Actually `eq(variables['Build.SourceBranch'], format('{0}{1}', 'refs/heads/', variables['OneLocBuildBranch'] ))` should be true. So the conditional parameter should be passed.
- Second, OneLocBuild task should get passed values as main branch also uses conditional parameters but doesn't have this problem.

I'd like to output the parameters/variables in the real internal run for investigation since there is no other way to simulate.

JanKrivanek (approved):
Looks as the best way forward in investigating this issue - approving

JanKrivanek:
Added as investigation of https://github.com/dotnet/templating/issues/5288

PR #5419, opened 2022-10-12
Stabilize installing Azure CLI for search cache pipeline

GangWang01:
### Problem
#5344 - Azure CLI installation failed randomly with the error below.
`E: Could not get lock /var/lib/dpkg/lock-frontend. It is held by process 9016 (unattended-upgr)`
The system automatically installs updates in the background. It needs to wait that to finish. 

### Solution
Stabilize installing Azure CLI using retries with some delay.

### Checks:
- [ ] Added unit tests


JanKrivanek (approved):
Thank you!

Issue #5295, opened 2022-09-22
Enable static analysis on templating repo


JanKrivanek:
**Due: Nov/15**

1. Ensure repo is marked "production" or "non-production". CodeQL enrollment is only being tracked on production repos. Instructions [here](https://eng.ms/docs/cloud-ai-platform/devdiv/one-engineering-system-1es/1es-docs/product-catalog/how-to/assign-classify-engineering-artifacts). Please also make sure Service Owners are correctly set in ServiceTree.

1. For C/C++, C#, Typescript, JavaScript, Python, Java (including Android), Ruby, and Go, add the [CodeQL](https://eng.ms/docs/cloud-ai-platform/devdiv/one-engineering-system-1es/1es-docs/codeql/configuring-codeql3000-ado-pipelines) variable to the Variables portion of your ADO Pipeline configuration: ```Codeql.Enabled: true```

1. Be sure to check the "Additional Options" and "How do I know scans are working and complete" sections. (If your language is not listed, there is no scan tool for AzDO Pipelines - you will want to run [Local scanning)](https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki?wikiVersion=GBwikiMaster&_a=edit&pagePath=/Compliance%20%26%20Tenets/Compliance%20Tenets/Security/Static%20Analysis&pageId=1761&anchor=local-scanning-(ide/cmd)-(recommended))

[more details here](https://liquid.microsoft.com/Web/Object/Read/MS.Security/Requirements/Microsoft.Security.SystemsADM.10201)
[and here](https://microsoft-my.sharepoint.com/:p:/r/personal/brbrot_microsoft_com/_layouts/15/Doc.aspx?sourcedoc=%7B2CDF6B89-72E9-41EE-A9F5-48DA6B221235%7D&file=CodeQLEnrollment.pptx&wdOrigin=TEAMS-ELECTRON.teams.chiclet&action=edit&mobileredirect=true)

After done - verify that sdk repo has adopted the anlysis as well (or link an issue here)

PR #5448, opened 2022-10-14
Enables CodeQL in build pipeline

MichalPavlik:
### Problem
#5295

### Solution
This enables CodeQL task in official build pipeline for `main` branch. It uses default settings, so CodeQL will collect data with 72 hours cadence.

### Checks:
We will have to check pipeline logs after merge. It works for MSBuild, I don't expect any problems blocking the build.


JanKrivanek (approved):
Thank you!

Issue #5483, opened 2022-10-20
Triage/Fix CodeQL findings


JanKrivanek:
https://1es.lgtm.microsoft.com/projects/ado/dnceng/internal%2Fdotnet-templating@version=gbmain&lgtmproxymetadata=build_jobname%3dbuild_release%3bbuild_stagename%3dbuild?mode=list

Should be added to our regular schedule as well

JanKrivanek:
@GangWang01 - please reassign back to me if the dashboard doesn't load for you

PR #5537, opened 2022-11-02
Fix CodeQL errors and hide warning/recommendation from query results

GangWang01:
### Problem
#5483

### Solution
For error, fix the violations.
For warning/recommendation, hide the query results by adding lgtm comment. Since [hiding query results via configuration file](https://1es.lgtm.microsoft.com/help/lgtm/showing-hiding-query-results) hides the results of rules applying to all source code rather than specific term/file path, there is some risk missing real problem.

### Checks:
- [ ] Added unit tests


JanKrivanek (approved):
Looks good. Thank you for taking care about this!

Issue #5562, opened 2022-11-04
`Cpp2StyleEvaluatorDefinition` concurency issue


JanKrivanek:
## Problem

`Cpp2StyleEvaluatorDefinition.TokenCache` is static but unguarded against concurrent access.
Following issue is occassinaly seen in CI as a result:

```
System.InvalidOperationException : Operations that change non-concurrent collections must have exclusive access. A concurrent update was performed on this collection and corrupted its state. The collection's state is no longer correct.

at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
   at System.Collections.Generic.Dictionary`2.set_Item(TKey key, TValue value)
   at Microsoft.TemplateEngine.Core.Expressions.Cpp2.Cpp2StyleEvaluatorDefinition.GetSymbols(IProcessorState processor) in /_/src/Microsoft.TemplateEngine.Core/Expressions/Cpp2/Cpp2StyleEvaluatorDefinition.cs:line 130
   at Microsoft.TemplateEngine.Core.Expressions.Shared.SharedEvaluatorDefinition`2.Evaluate(IProcessorState processor, Int32& bufferLength, Int32& currentBufferPosition, String& faultedMessage, HashSet`1 referencedVariablesKeys) in /_/src/Microsoft.TemplateEngine.Core/Expressions/Shared/SharedEvaluatorDefinition.cs:line 39
   at Microsoft.TemplateEngine.Core.Expressions.Shared.SharedEvaluatorDefinition`2.EvaluateFromString(ILogger logger, String text, IVariableCollection variables, String& faultedMessage, HashSet`1 referencedVariablesKeys) in /_/src/Microsoft.TemplateEngine.Core/Expressions/Shared/SharedEvaluatorDefinition.cs:line 98
   at Microsoft.TemplateEngine.Core.Expressions.Shared.SharedEvaluatorDefinition`2.EvaluateFromString(ILogger logger, String text, IVariableCollection variables) in /_/src/Microsoft.TemplateEngine.Core/Expressions/Shared/SharedEvaluatorDefinition.cs:line 78
[...]

PR #5709, opened 2022-11-30
Fix race condition in cpp2 style evaluator definition

YuliiaKovalova:
### Problem
The fix of the [#5562](https://github.com/dotnet/templating/issues/5562)

### Solution
Add lock for non-concurrent object

### Checks:
- [ ] Added unit tests **(N/A)**
- [ ] Added `#nullable enable` to all the modified files [?](https://github.com/dotnet/templating/wiki/Contributing#coding-style) **(N/A)**


vlada-shubina (approved):
This should solve the problem. 

Did you consider making `TokenCache` `ConcurrentDictionary`? Might be cleaner way and better optimized.

Also consider cleaning up the branch to make rebase possible.

vlada-shubina (approved):

PR #5816, opened 2022-12-21
[release/6.0.1xx]  inverted language version conditions in console / classlib templates

YuliiaKovalova:
## Problem
fixes https://github.com/dotnet/templating/issues/5668

## Description
The net6.0 templates doesn't support C# 11 (and later) at the moment. In case C# 11 is requested via --langVersion the template that uses old syntax will be created.
In the fix the condition on language version was reverted, so the default behavior is new syntax and that the C#11 and further language versions will use new syntax.

## Customer Impact
when executing dotnet new console --framework net6.0 --langVersion 11.0 or dotnet new classlib --framework net6.0 --langVersion 11.0 the template that uses old syntax is created.
After the fix, the template that uses latest language features (top-level statements, file-scoped namespaces etc) will be used

## Testing
Automated

## Risk
Very low
Those conditions are already used in .NET 8 templates and passing the tests

## Checks:
 Added unit tests - existing tests cover the change
 Added #nullable enable to all the modified files [?](https://github.com/dotnet/templating/wiki/Contributing#coding-style) - NA

### DIFF
--- a/template_feed/Microsoft.DotNet.Common.ProjectTemplates.6.0/content/ClassLibrary-CSharp/.template.config/template.json
+++ b/template_feed/Microsoft.DotNet.Common.ProjectTemplates.6.0/content/ClassLibrary-CSharp/.template.config/template.json
@@ -7,7 +7,7 @@
   ],
   "name": "Class Library",
   "generatorVersions": "[1.0.0.0-*)",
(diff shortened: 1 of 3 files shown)


JanKrivanek (approved):
Looks good!

PR #5897, opened 2023-01-09
fix pool name for public branch

YuliiaKovalova:
### Problem
 Servicing Builds Running in R&D Pools Alert 
Fixes #https://github.com/dotnet/arcade/issues/12131
### Solution
use -Svc for servicing builds

### Checks:
- [N/A ] Added unit tests
- [N/A ] Added `#nullable enable` to all the modified files [?](https://github.com/dotnet/templating/wiki/Contributing#coding-style)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -23,8 +23,6 @@ variables:
     value: NETDevUX
   - name: _PublishUsingPipelines
     value: true
-  - name: Codeql.Enabled
-    value: true
   - ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
     - group: Templating-SDLValidation-Params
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -23,8 +23,6 @@ variables:
     value: NETDevUX
   - name: _PublishUsingPipelines
     value: true
-  - name: Codeql.Enabled
-    value: true
```
JanKrivanek, line 27: Is this intentional?

JanKrivanek (changes requested):
Keep CodeQl enabled (unless there is a reason not to)
Otherwise looks good

azure-pipelines.yml
YuliiaKovalova: you are right.
done

PR #5954, opened 2023-01-19
Test with all .NET versions for CacheFileTests

GangWang01:
### Problem
CacheFileTests haven't worked for all .NET versions for a long time.

### Solution
Enable pull request triggers search cache pipeline to verify the tests.

### Checks:
- [ ] Added unit tests


GangWang01:
It looks like pr trigger only works if the change is merged but it's not possible.

PR #6015, opened 2023-01-30
Design proposal for interactive invocation

baronfel:
This PR adds a proposed design doc for interactive invocation to open it up for discussion. I've focused on the invocation scenario for now - there are enough open questions to discuss that will likely take a bit to work through.

Need more details on
* [ ] specific opt-in/opt-out mechanism for interactive mode (mirrored from MSBuild Live Logger)


docs/design/proposed/interactive-mode.md
```diff
@@ -0,0 +1,47 @@
...
+
+### Concerns/open questions
+
+* Many templates have symbols that are 'internal' and/or easily confused - e.g. TargetFrameworkOverride and Framework. How to coalesce these to prevent multiple prompts for something a user views as one item?
+* Some parameters are not something I'd expect a user to ever want to change - e.g. skipRestore. Should we have another property marker that hides them?
+  * For initial version, prompt for these too. We can tweak logic based on feedback.
+* Which of the 'general' invocation options like -n and -o should we present here?
+* Should we have a new `prompt` property for this? Descriptions feel odd grammatically (which is why I didn't use them in the below example).
```
vlada-shubina, line 32: this belongs to `dotnetcli.host.json` imo

docs/design/proposed/interactive-mode.md
```diff
@@ -0,0 +1,47 @@
...
+
+### Examples
+
+Here's what the C# Console template might look like using just the data we have available right now:
```
vlada-shubina, line 36: what should be the entry point for interactivity:
- `dotnet new <template name>`
- or `dotnet new` with first ask for template name?

If latter, this feature may be helpful to determine choices: 
https://github.com/dotnet/templating/issues/3402

docs/design/proposed/interactive-mode.md
```diff
@@ -0,0 +1,47 @@
...
+
+### Concerns/open questions
+
+* Many templates have symbols that are 'internal' and/or easily confused - e.g. TargetFrameworkOverride and Framework. How to coalesce these to prevent multiple prompts for something a user views as one item?
+* Some parameters are not something I'd expect a user to ever want to change - e.g. skipRestore. Should we have another property marker that hides them?
+  * For initial version, prompt for these too. We can tweak logic based on feedback.
+* Which of the 'general' invocation options like -n and -o should we present here?
```
vlada-shubina, line 31: imo, that can be a hardcoded list for 1st release.
Note that for some of the templates, e.g. `editorconfig` name is not relevant and the question should be skipped.

This can be potentially determined if `sourceName` is defined in the template.

docs/design/proposed/interactive-mode.md
```diff
@@ -0,0 +1,47 @@
...
+* Should we have a new `prompt` property for this? Descriptions feel odd grammatically (which is why I didn't use them in the below example).
+
+### Examples
+
+Here's what the C# Console template might look like using just the data we have available right now:
+
+![](sample-interactive-mode.png)
```
vlada-shubina, line 38: love colors 👍

docs/design/proposed/interactive-mode.md
baronfel: I was thinking specifically about `dotnet new <template name>` here - to scope the initial unit of work.

docs/design/proposed/interactive-mode.md
baronfel: That's a good point - if the specific data becomes more general then we could promote it to the template config.

docs/design/proposed/interactive-mode.md
vlada-shubina: Item templates can also have a `sourceName`, e.g. class, so likely we need an additional logic based on `sourceName` here.

JanKrivanek (approved):
Love it!

docs/design/proposed/interactive-mode.md
```diff
@@ -0,0 +1,64 @@
...
+
+When interactive mode is entered, the command should prompt the user for each required symbol. If the symbol has a default value, it should be displayed as the default value with some formatting (for example, grayed out text). If the symbol has a description, it should be displayed as part of a prompt. If the symbol has a datatype, the command should validate the input and prompt the user again if the input is invalid. At each stage of the prompt, the user should be able to press a key to skip the prompt and use the default value (if one exists). At each stage of the prompt, the user should be able to press a key to skip all other prompts and use their default values (if they exist). Finally, at each stage of the prompt the user should be able to press a key and quit the entire operation.
+
+### Detecting interactive mode
+
+* If the `--interactive` CLI flag was passed, then interactive mode should be used.
```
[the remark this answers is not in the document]
baronfel: Yep! The .NET CLI already provides both versions of the switch - we should correctly adhere to both, as you say.

docs/design/proposed/interactive-mode.md
```diff
@@ -0,0 +1,64 @@
...
+
+When interactive mode is entered, the command should prompt the user for each required symbol. If the symbol has a default value, it should be displayed as the default value with some formatting (for example, grayed out text). If the symbol has a description, it should be displayed as part of a prompt. If the symbol has a datatype, the command should validate the input and prompt the user again if the input is invalid. At each stage of the prompt, the user should be able to press a key to skip the prompt and use the default value (if one exists). At each stage of the prompt, the user should be able to press a key to skip all other prompts and use their default values (if they exist). Finally, at each stage of the prompt the user should be able to press a key and quit the entire operation.
+
```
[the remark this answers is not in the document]
baronfel: Yes - we should use the implicit default as the 'default value' in this case

docs/design/proposed/interactive-mode.md
```diff
@@ -0,0 +1,64 @@
...
+Here's what the C# Console template might look like using just the data we have available right now:
+
+![](sample-interactive-mode.png)
+
+Key points are the use of defaults where available, and the slight use of color to highlight property/symbol names vs the descriptions.
+One thing that can't been seen here is the use of choices - the 'framework' parameter showed a selection prompt and let me enter that.
+
+## References
+
+* [MSBuild LiveLogger](https://github.com/dotnet/msbuild/tree/main/src/MSBuild/FancyLogger)
```
donJoseLuis, line 62: Please double check the link, I get a 404

donJoseLuis (changes requested):

docs/design/proposed/interactive-mode.md
maridematte: It has been recently renamed to LiveLogger, so the name is right on the doc but the link doesnt quite match rn.

baronfel: Thanks for the fix @maridematte - @donJoseLuis good to clear the change request?

docs/design/proposed/interactive-mode.md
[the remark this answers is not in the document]
JanKrivanek: Ok

Should we then possibly reword `If the symbol has a default value` to something like `The symbol default (or implicit default) value should allways be displayed`.

It's a nitpick - but just so that this doesn't slip.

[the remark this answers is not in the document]
baronfel: done!

docs/design/proposed/interactive-mode.md
baronfel: added to the doc!

YuliiaKovalova:
I like it!
Thank you, Chet!

docs/design/proposed/interactive-mode.md
```diff
@@ -0,0 +1,68 @@
...
+* Where possible, provide an attractive and informative prompting experience through the use of ANSI control codes and terminal formatting. When the environment doesn't support it, use 'boring' formatting.
+
+## Non-Goals
+
+* Provide interactivity to the entire .NET CLI
+* Extend interactivity to commands like `dotnet new search` or `dotnet new list` (i.e. interactive filtering)
+
+## Workflow
+
+When a user invokes a template, the command should evaluate the user-provided parameters and determine if it should enter interactive mode. This is done by checking if the command is being invoked from a terminal (i.e. isatty is true) and if the command has any symbols that are of type `parameter` and are required _while also having required symbols that are not explicitly provided_. If both of these conditions are true, the command should enter interactive mode.
```
JanKrivanek, line 22: This sounds very reasonable. Below in `Detecting interactive mode`, there is only the explicit opt-in option.
Do we want implicit switch to interactive mode under the conditions outlined here as well? Or not for the initial version?

PR #6129, opened 2023-02-22
Add warning on overlapping identity

### LINKED ISSUE #4817: Installing template with same identity silently hides previously installed template
**Background:**
When installing new template with the same `identity`, any previously installed template installed with the same identity (e.g from different location or/and with different name), will be silently hiden - it's not present in `dotnet new list` and cannot be isntantiated via shortname. It's only present in `dotnet new uninstall` list

**Example:**
```console
> dotnet new install templateA
The following template packages will be installed:
   C:\tmp\templateA

Success: C:\tmp\templateA installed the following templates:
Template Name  Short Name  Language  Tags
-------------  ----------  --------  -------
Name A         name-a      [C#]      Console

> dotnet new list
These templates matched your input:

Template Name                        Short Name                               Language    Tags
-----------------------------------  ---------------------------------------  ----------  -------------------------------------
Name A                               name-a                                   [C#]        Console

> dotnet new install templateB
The following template packages will be installed:
   C:\tmp\templateB

Success: C:\tmp\templateB installed the following templates:
Template Name  Short Name  Language  Tags
-------------  ----------  --------  -------
Name B         name-b      [C#]      Console

> dotnet new list
These templates matched your input:

Template Name                        Short Name                               Language    Tags
-----------------------------------  ---------------------------------------  ----------  -------------------------------------
Name B                               name-b                                   [C#]        Console

> dotnet new name-a
No templates found matching: 'name-a'.

To list installed templates, run:
   dotnet new list
To search for the templates on NuGet.org, run:
   dotnet new search name-a

> dotnet new uninstall

   C:\tmp\templateA
      Uninstall Command:
         dotnet new uninstall C:\tmp\templateA

   C:\tmp\templateB
      Templates:
         Name B (name-b) C#
      Uninstall Command:
         dotnet new uninstall C:\tmp\templateB
```

**Expected behavior:**
User is warned and prompted to force the action if needed (or at least to first uninstall the previously installed template).

YuliiaKovalova:
### Problem
Fixes #4817

### Solution
Add a warning when there is an attempt to overwrite the **existing template from the managed package** on a **managed** one with an overlapping identity.

The need to restrict the warning appearance is justified by Project/Item templates installation for sdk project.
E.g.
image

### Checks:
- [*] Added unit tests
- [N/A] Added `#nullable enable` to all the modified files [?](https://github.com/dotnet/templating/wiki/Contributing#coding-style)


JanKrivanek (approved):
Looks good!

src/Microsoft.TemplateEngine.Edge/LocalizableStrings.resx
```diff
@@ -395,4 +395,8 @@ Template cache will be recreated on the next run.</value>
     <value>'IWorkloadsInfoProvider' component provided by host provided some duplicated workloads (duplicates: {0}). Duplicates will be skipped.</value>
     <comment>{0} is the list of duplicates</comment>
   </data>
+  <data name="TemplatePackageManager_Warning_DetectedTemplatesIdentityConflict" xml:space="preserve">
+    <value>The template {0} has the identity conflict with an already existing template {1} installed from package {2}. The template {1} will be overwritten.</value>
```
vlada-shubina, line 399: is it possible to include the package of template {0} here?

src/Microsoft.TemplateEngine.Edge/LocalizableStrings.resx
```diff
@@ -395,4 +395,8 @@ Template cache will be recreated on the next run.</value>
...
+    <comment>The template {0} has the identity conflict with an already existing template {1} installed from package {2}. The template {1} will be overwritten.</comment>
```
vlada-shubina, line 400: Comment should contain the information on what placeholders contain, with examples if applicable, or blank.

src/Microsoft.TemplateEngine.Edge/Settings/TemplateCache.cs
```diff
@@ -31,6 +31,14 @@ public TemplateCache(ScanResult?[] scanResults, Dictionary<string, DateTime> mou
                 }
                 foreach (ITemplate template in scanResult.Templates)
                 {
+                    if (templateDeduplicationDictionary.ContainsKey(template.Identity))
+                    {
+                        _logger.LogWarning(string.Format(
+                            LocalizableStrings.TemplatePackageManager_Warning_DetectedTemplatesIdentityConflict,
+                            template.Name,
```
vlada-shubina, line 38: please include identity in the message too.

src/Microsoft.TemplateEngine.Edge/Settings/TemplateCache.cs
```diff
@@ -31,6 +31,14 @@ public TemplateCache(ScanResult?[] scanResults, Dictionary<string, DateTime> mou
...
+                            templateDeduplicationDictionary[template.Identity].Template.Name,
+                            templateDeduplicationDictionary[template.Identity].Template.MountPointUri));
+                    }
```
vlada-shubina, line 41: is it possible to get package display name instead? Mount Point URI is a path.

vlada-shubina: What would be the final message example in case 2 template packages (NuGet based) have same template?

src/Microsoft.TemplateEngine.Edge/LocalizableStrings.resx
```diff
@@ -395,4 +395,8 @@ Template cache will be recreated on the next run.</value>
     <value>'IWorkloadsInfoProvider' component provided by host provided some duplicated workloads (duplicates: {0}). Duplicates will be skipped.</value>
     <comment>{0} is the list of duplicates</comment>
   </data>
-</root>
\ No newline at end of file
+  <data name="TemplatePackageManager_Warning_DetectedTemplatesIdentityConflict" xml:space="preserve">
+    <value>The template {0} installed from package {1} has the identity {2} that conflicts with an already existing template {3}. The template {3} will be overwritten.</value>
```
vlada-shubina, line 399:
```suggestion
    <value>The template {0} installed from package {1} has the identity {2} that conflicts with an already existing template {3} from package {4}. The latter template will be overwritten.</value>
```
(just in case {0} and {3} are same)

src/Microsoft.TemplateEngine.Edge/LocalizableStrings.resx
```diff
@@ -395,4 +395,8 @@ Template cache will be recreated on the next run.</value>
...
+    <comment>The template 'TemplateB' installed from package 'PackageB' has the identity 'OverlappingIdentity' that conflicts with an already existing template 'TempalteA'. The template 'TemplateA' will be overwritten.</comment>
```
vlada-shubina, line 400:
```suggestion
    <comment>Example of usage is: "`changed string`".</comment>
```

YuliiaKovalova:
@baronfel , we are discussing one more probable case when we have several conflicting packages.
In the current implementation, there will be a separate warning per each one:
_"The template '{0}' installed from package '{1}' has the identity '{2}' that is the same as the identity of the template '{3}' from package '{4}'. The latter template will not be used."_

 but it is possible to group them and print something like:
_"The following templates have the same identity 'id1': 'template A' (package:  'Package A'), 'template B' (package:  'Package B'), 'template C' (package:  'Package C').  The template 'template A' (package:  'Package A') will be used, the other templates are ignored.  To resolve the conflict, uninstall conflicting template packages."_

What is your consideration here?
And please adjust texting if something doesn't sound like a user friendy message.

baronfel:
I think grouping the messages into a compound message makes the most sense - great idea!

### Proposed wording

> The following templates use the same identity 'IDENTITY':
>
> * 'TEMPLATE' from 'PACKAGE_ID'
> * _repeat for each duplicate template package_
>
> The template from 'PACKAGE_ID' will be used. To resolve this conflict, uninstall the conflicting template packages.

### Reasoning

Formatting each entry on a new line helps users scan for the important information - all the templates on a single line makes it very hard to identify the key parts of the message.

One question though - in the error message displayed in the PR description, the package is identified by full path - for a NuGet-installed package what will the PACKAGE_ID look like? Hopefully just the package name?

src/Microsoft.TemplateEngine.Edge/Settings/TemplateCache.cs
```diff
@@ -5,20 +5,28 @@
...
+
         private readonly ILogger _logger;
 
-        public TemplateCache(ScanResult?[] scanResults, Dictionary<string, DateTime> mountPoints, ILogger logger)
+        public TemplateCache(IReadOnlyList<ITemplatePackage> allTemplatePackages, ScanResult?[] scanResults, Dictionary<string, DateTime> mountPoints, ILogger logger)
         {
             _logger = logger;
+
+            // this dictionary contains information about managed templates with overlapping identities
+            var overlappingIdentitiesMap = new Dictionary<DuplicatedIdentity, IList<(string TemplateName, string PackageId)>>(new DuplicatedIdentityComparer());
```
vlada-shubina, line 28: would it be easier as 
```suggestion
            var overlappingIdentitiesMap = new Dictionary<string, IList<(ITemplateInfo template, string PackageId)>>();
```
where the key is identity.

then likely replacement logic used in CheckForOverlappingIdentity won't be needed, it would be enough to add the data.

vlada-shubina (changes requested):
Please check the comment in CheckForOverlappingIdentity - likely there is still case which is unhandled.

src/Microsoft.TemplateEngine.Edge/Settings/TemplateCache.cs
YuliiaKovalova: I need to have a path for both entries - the one from the key is used in the warning message too
image

src/Microsoft.TemplateEngine.Edge/Settings/TemplateCache.cs
```diff
@@ -139,5 +156,77 @@ public TemplateCache(JObject? contentJobject, ILogger logger)
...
+        private void CheckForOverlappingIdentity(
+            IDictionary<DuplicatedIdentity, IList<(string TemplateName, string PackageId)>> overlappingIdentitiesMap,
+            ITemplate template,
+            ITemplate checkedTemplate,
+            IReadOnlyList<ITemplatePackage> allTemplatePackages)
+        {
+            // add warning for the case when there is an attempt to overwrite existing managed by new managed template
+            var templatePackage = allTemplatePackages.FirstOrDefault(tp => tp.MountPointUri == template.MountPointUri);
+            var checkedTemplatePackage = allTemplatePackages.FirstOrDefault(tp => tp.MountPointUri == checkedTemplate.MountPointUri);
+            if (templatePackage is IManagedTemplatePackage managedTP && checkedTemplatePackage is IManagedTemplatePackage checkedManagedTP)
```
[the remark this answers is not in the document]
YuliiaKovalova: fixed

src/Microsoft.TemplateEngine.Edge/Settings/TemplateCache.cs
```diff
@@ -5,46 +5,69 @@
...
+                        };
+                    }
                 }
             }
 
             var templates = new List<TemplateInfo>();
-            foreach (var newTemplate in templateDeduplicationDictionary.Values)
+            foreach (var duplicatedIdentities in templateDeduplicationDictionary)
             {
+                var newTemplate = duplicatedIdentities.Value.LastOrDefault();
```
vlada-shubina, line 61: it would be good to include the comment why the last is taken, also `OrDefault` is not needed.

src/Microsoft.TemplateEngine.Edge/Settings/TemplateCache.cs
```diff
@@ -139,5 +162,34 @@ public TemplateCache(JObject? contentJobject, ILogger logger)
...
         }
+
+        // add warning for the case when there is an attempt to overwrite existing managed by new managed template
+        private void PrintOverlappingIdentityWarning(IDictionary<string, IList<(ITemplate Template, string PackageDisplayName, ILocalizationLocator? Localization)>> templateDeduplicationDictionary)
+        {
+            foreach (var identityToTemplates in templateDeduplicationDictionary)
+            {
+                // we print the message only if managed template wins and we have > 1 managed templates with overlapping identities
+                var lastTemplate = identityToTemplates.Value.LastOrDefault();
+                var managedTemplates = identityToTemplates.Value.Where(templateInto => templateInto.Template is IManagedTemplatePackage).ToArray();
```
vlada-shubina, line 173:
```suggestion
                var managedTemplates = identityToTemplates.Value.Where(templateInto => templateInto.Template is IManagedTemplatePackage);
```

src/Microsoft.TemplateEngine.Edge/Settings/TemplateCache.cs
```diff
@@ -139,5 +162,34 @@ public TemplateCache(JObject? contentJobject, ILogger logger)
...
             return null;
         }
+
+        // add warning for the case when there is an attempt to overwrite existing managed by new managed template
+        private void PrintOverlappingIdentityWarning(IDictionary<string, IList<(ITemplate Template, string PackageDisplayName, ILocalizationLocator? Localization)>> templateDeduplicationDictionary)
+        {
+            foreach (var identityToTemplates in templateDeduplicationDictionary)
+            {
+                // we print the message only if managed template wins and we have > 1 managed templates with overlapping identities
+                var lastTemplate = identityToTemplates.Value.LastOrDefault();
```
vlada-shubina, line 172:
```suggestion
                var lastTemplate = identityToTemplates.Value.Last();
```

vlada-shubina: if last template is not IManagedTemplatePackage, we can continue and avoid creating the array.

src/Microsoft.TemplateEngine.Edge/Settings/TemplateCache.cs
```diff
@@ -139,5 +162,34 @@ public TemplateCache(JObject? contentJobject, ILogger logger)
...
+                var managedTemplates = identityToTemplates.Value.Where(templateInto => templateInto.Template is IManagedTemplatePackage).ToArray();
+                if (lastTemplate.Template is IManagedTemplatePackage && managedTemplates.Length > 1)
```
vlada-shubina, line 174:
```suggestion
                if (lastTemplate.Template is IManagedTemplatePackage && managedTemplates.Take(2).Count() > 1)
```

vlada-shubina: as an alternative not to create the array.

vlada-shubina (commented):
Consider creating tests for new functionality.

The logging messages can be tested as:
https://github.com/dotnet/templating/blob/fcb35d2728f9951088cd95d2c22c0a8bf19ce8ce/test/Microsoft.TemplateEngine.Orchestrator.RunnableProjects.UnitTests/TemplateConfigTests/LocalizationTests.cs#L405-L407

The list of template packages and scan results can be mocked as in https://github.com/dotnet/templating/blob/main/test/Microsoft.TemplateEngine.Edge.UnitTests/TemplateCacheTests.cs.

vlada-shubina (approved):
looks good, thanks for adding tests

Issue #5972, opened 2023-01-23
Unexpected Internal Software Issue (EX_SOFTWARE)


pantonis:
I'm trying to install a new template and I get the following error: 

````
Error: Failed to retrieve template packages from provider 'Global Settings'.
Details: Error reading JObject from JsonReader. Path '', line 0, position 0.
dotnet : Error reading JObject from JsonReader. Path '', line 0, position 0.
At line:1 char:1
+ dotnet new install .\
+ ~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (Error reading J... 0, position 0.:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
 
   at Newtonsoft.Json.Linq.JObject.Load(JsonReader reader, JsonLoadSettings settings)
   at Microsoft.TemplateEngine.JExtensions.ReadObject(IPhysicalFileSystem fileSystem, String path)
   at Microsoft.TemplateEngine.Edge.BuiltInManagedProvider.GlobalSettings.GetInstalledTemplatePackagesAsync(CancellationToken cancellationToken)
   at Microsoft.TemplateEngine.Edge.BuiltInManagedProvider.GlobalSettingsTemplatePackageProvider.InstallAsync(IEnumerable`1 installRequests, CancellationToken cancellationToken)
   at Microsoft.TemplateEngine.Cli.TemplatePackageCoordinator.EnterInstallFlowAsync(InstallCommandArgs args, CancellationToken cancellationToken)
   at Microsoft.TemplateEngine.Cli.Commands.InstallCommand.ExecuteAsync(InstallCommandArgs args, IEngineEnvironmentSettings environmentSettings, TemplatePackageManager templatePackageManager, 
InvocationContext context)
   at Microsoft.TemplateEngine.Cli.Commands.BaseCommand`1.InvokeAsync(InvocationContext context)

For details on the exit code, refer to https://aka.ms/templating-exit-codes#70
````

.NET SDK 7:

```
7.0.101 [C:\Program Files\dotnet\sdk]
7.0.102 [C:\Program Files\dotnet\sdk]
```

YuliiaKovalova:
Hi @pantonis ,

It looks like a cache issue.
Could you check if the file C:\Users\%USERPROFILE%\.templateengine\packages.json isn't empty and wasn't corrupted?

pantonis:
It was empty. I found out after 7 hours. Can't you just add a more friendly message?

[...]

PR #6145, opened 2023-02-23
Check the availability of global settings file

GangWang01:
### Problem
#5972

### Solution
1. Check the availability of global settings file before and after the operations to detect any corruption and return the error in operation result.
2. Report the error according to the error in operation result. (Pending to do in another PR in SDK repo)

### Checks:
- [ ] Added unit tests


GangWang01:
Overall it needs getting all template packages first that reads global settings file prior to the operations(install/uninstall/update package. Checking global settings file inside the operations could not detect the corruption when getting all template packages. This PR doesn't work.
In addition, getting template packages is done by Microsoft.TemplateEngine.Edge.Settings.TemplatePackageManager in GetTemplatePackagesAsync() when loading providers. It doesn't throw out the exception caused by the corruption. This needs a good way to get the providers only for checking global settings.

Issue #5859, opened 2023-01-05
Disallow HTTP feeds


donJoseLuis:
Disallowing HTTP feeds, disallowing untrusted/unverified nuget.org categories

baronfel:
Package Source management should be handled by nuget.config transparently, so to me this one mostly looks like additional filters on `search` for trusted only/verified only packages.

YuliiaKovalova:
This handling is relevant for install/update actions.
We suggest **do not allow** installation/update if customer adds option `--add source` and specifies url with **http** feed.
But still it makes sense to allow use **http** source that if` --force` flag is specified

@baronfel, we would love to hear your opinion here.

baronfel:
That seems like a good balance. 

Error case:
```
> dotnet new install flibbertygibbet --nuget-source http://insecure-feed.org
The following template packages will be installed:
   flibbertygibbet

The nuget source 'http://insecure-feed.org' is insecure and will not be searched for the template 'myTemplate'.

flibbertygibbet could not be installed, the package does not exist.

For details on the exit code, refer to https://aka.ms/templating-exit-codes#103
```

Override case:
```
> dotnet new install flibbertygibbet --nuget-source http://insecure-feed.org --force
The following template packages will be installed:
   flibbertygibbet

Success: flibbertygibbet::0.32.0 installed the following templates:
Template Name    Short Name     Language  Tags
-------------    -------------  --------  ------------------------
flibbertygibbet  fg             C#        Scaffold/Console/Library
```

Are there other flags that the nuget API uses to mark sources as secure/insecure? If there is an API for this we should use that instead.

YuliiaKovalova:
the question about validation on Nuget side https://github.com/NuGet/Home/discussions/12484

PR #6304, opened 2023-03-23
Remove http feeds when searching for templates

maridematte:
### Problem
Solution for this issue https://github.com/dotnet/templating/issues/5859

### Solution
When installing or updating packages, if `--force` is not specified, do not search feeds that are not secure (`http` feeds) and give a warning about that.

### Checks:
- [ ] Added unit tests


maridematte:
Opened a new PR in https://github.com/dotnet/templating/pull/6313. There was an error with the localization file that I was not able to solve, so I redid the changes in a new branch.

PR #6362, opened 2023-04-11
update InstallAzureCLI.sh for fixing function not found issue

YuliiaKovalova:
### Problem
Update cache pipeline fails due to  InstallAzureCLI.sh script

### Solution
Update function according to described solution https://stackoverflow.com/questions/12468889/bash-script-error-function-not-found-why-would-this-appear#:~:text=I-,faced,-the%20same%20problem

Run with check https://dev.azure.com/dnceng/internal/_build/results?buildId=2154986&view=results

### Checks:
- [ n/a ] Added unit tests
- [ n/a ] Added `#nullable enable` to all the modified files [?](https://github.com/dotnet/templating/wiki/Contributing#coding-style)

### DIFF
--- a/eng/SearchCache/InstallAzureCLI.sh
+++ b/eng/SearchCache/InstallAzureCLI.sh
@@ -1,4 +1,4 @@
-function retry {
+retry() {
   local n=0
   local max=15
   local delay=60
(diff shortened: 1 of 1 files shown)


JanKrivanek (approved):
LGTM!

PR #6392, opened 2023-04-14
Fix variable expression used in condition

GangWang01:
### Problem
#6377 - the variable expression used in condition is wrong.

### Solution
Correct the expression.

### Checks:
- [ ] Added unit tests


JanKrivanek (approved):
LGTM

PR #6786, opened 2023-06-21
preview6 branding

marcpopMSFT:


marcpopMSFT:
@YuliiaKovalova it looks like you updated main to preview 7 but I didn't see a preview 6 branch here or internally so I created one and then this is to fix up the branding for that branch.

YuliiaKovalova (approved):

YuliiaKovalova:
It's strange a bit, because I was targeting main in my PR, but the changes were merged in both main/preview6 branches.
Thank you for fixing!

marcpopMSFT:
I think it's because the preview 6 branch hadn't been created yet (or if it had, I overwrote it somehow without any feedback from the tooling)

PR #6845, opened 2023-07-11
Remove dependabot on branch 7.0.1xx，7.0.2xx and 7.0.3xx

v-wuzhai:
I noticed that PRs for branch 7.0.1xx，7.0.2xx and 7.0.3xx were closed, so I removed dependabot on those branches.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,36 +9,6 @@ updates:
     labels:
       - "dependencies"
       - "dependabot: main"
-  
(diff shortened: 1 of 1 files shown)


marcpopMSFT:
Seems reasonable to me. I think main and the latest release branch make sense to me for this. If there ends up being a security fix, we can backport but most of these are just bug fix level and it increases the churn (and risk) in servicing.  CC @joeloff @MiYanni

PR #6961, opened 2023-08-16
Fix parameter in null/empty check

joeloff:
Fixing a typo for `IsNullOrWhitespace` checking the wrong parameter


baronfel (approved):

baronfel:
is this something that should go to .NET 8?

joeloff:
Probably, but then this has likely been around for a long time. I can port the change to release/8.0.1xx and then it gets picked for RC2

PR #6988, opened 2023-08-25
Fix the search test behavior

marcpopMSFT:
Port [fix ](https://github.com/dotnet/sdk/pull/34843/files) for localized template from SDK
Reenable test

### Problem
1. A customer pushed a new localzied tempalte which broke the search sort ordering in the test
2. This blocked internal codeflow as the CI build was running tests

### Solution
I ported the fix from the SDK repo: https://github.com/dotnet/sdk/pull/34843/files
I disabled running tests when running in internal CI
I reenabled the test


MiYanni (approved):
:shipit:

PR #7098, opened 2023-09-28
8.0.2xx branding

Forgind:
### Checks:
- [ ] Added unit tests

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -4,7 +4,7 @@
   <PropertyGroup>
     <UsingToolXliff>true</UsingToolXliff>
     <UsingToolNetFrameworkReferenceAssemblies>true</UsingToolNetFrameworkReferenceAssemblies>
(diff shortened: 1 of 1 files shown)


eng/Versions.props
```diff
@@ -4,7 +4,7 @@
   <PropertyGroup>
     <UsingToolXliff>true</UsingToolXliff>
     <UsingToolNetFrameworkReferenceAssemblies>true</UsingToolNetFrameworkReferenceAssemblies>
-    <VersionPrefix>8.0.100</VersionPrefix>
+    <VersionPrefix>8.0.200</VersionPrefix>
```
marcpopMSFT, line 7: I believe line 13 should be preview

Forgind: Good catch; fixed

marcpopMSFT (approved):

PR #7103, opened 2023-09-29
Fix localization warning on project template installation

MiYanni:
Fixes: https://github.com/dotnet/sdk/issues/35768
Fixes: https://github.com/dotnet/aspnetcore/issues/51003

## Summary
When determining the choices of a localization symbol, it was accidentally using all strings instead of the grouping for that particular symbol. Installing any project template would cause the warnings to appear. I used `dotnet new install Microsoft.DotNet.Web.Spa.ProjectTemplates` to observe the issue.

## Details
I went through the changes in the `templatestrings.xx.json` between RC1 and RC2. I found a string that was causing an issue and then figured out why it was problematic after looking at all the new strings that were added. These 2 were causing the warning to occur:
- `"symbols/InteractivityPlatform/choices/None/description": "No interactivity (static server rendering only)",`
- `"symbols/auth/choices/None/description": "No authentication",`

The warning was being thrown by [line 115](https://github.com/dotnet/templating/blob/main/src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/LocalizationModelDeserializer.cs#L115) because it was being sent **all** the `choices` entries. This caused it to get 2 entries for `None/description`, which fails the `Single` call. The code change in this PR should now only be sending the `choices` entries that apply to the particular symbol. For clarity, the name of the symbol is the segment after `symbols/`. The two in question are `InteractivityPlatform` and `auth`.

## Testing
The repo has some validation tests around localization parsing. I've added entries to the existing test `templatestrings.xx.json` files that would cause this `choices` bug to occur. With the fix to the code, these tests pass appropriately.

Tests without the codefix:
image

Tests after the codefix:
image


halter73 (approved):

baronfel:
Sorry @halter73 - there's basically no way. RC2 is locked down super tight now.

nagilson (approved):
Thank you for writing up a thorough explanation, that is quite the nasty bug. LGTM :)

marcpopMSFT:
@MiYanni this should have gone through tactics approval since we're in QB mode now. I marked it and we can do post-merge approval.

I also created a known issue for RC2 so customers know what the impact is (I assume that help for the blazor templates will be unlocalized). Please review: https://github.com/dotnet/core/pull/8811

MiYanni:
@marcpopMSFT
> this should have gone through tactics approval since we're in QB mode now.

When did the `8.0.1xx` branch go into QB mode? The earliest date I see for the build after RC2 is Oct 17th on the schedule. Maybe I'm not understanding the schedule...

Edit: Looked at the schedule again. Looks like every QB mode is in Red (`#ff0000`). The last one was in RC2 but says it ended Sept 29th. So, I definitely have no idea when the QB mode happened.

marcpopMSFT:
Looking at the dotnet schedule, it has M2 approval until RC2 snaps but doesn't explicitly call out that the runtime enters QB mode after that point (which was mid-Sept iirc). For tooling, we've always compromised between the VS and .net schedules for this. The VS schedule entered QB mode with the final snap for preview 3 which happened on 9/29.

We can chat offline as the exact date for the SDK is a bit nebulous because teams checking into VS are still making changes outside of QB mode even after the RC2 snap (which is when we'd normally go into QB mode) so we allow some flexibility for tooling.

PR #7177, opened 2023-10-19
Revert "Bump Verify.XUnit from 20.4.0 to 22.1.3 in /eng/dependabot"

MiYanni:
Reverts dotnet/templating#7145

For now, we're not updating the XUnit versions for release branches.


MiYanni:
@v-wuzhai Just FYI to not update the XUnit versions. The flow for the version relies on Arcade and it is not something easily fixable, currently.

joeloff (approved):

PR #7223, opened 2023-10-31
Switch back to the shipped RC2 version. 

marcpopMSFT:
We should explore removing this dependency entirely.

@MiYanni , the change to remove the runtime seems pretty simple so maybe we should just do that instead:
https://github.com/dotnet/templating/commit/7e0fb4b044b6a9b89c6af104eee261705e74a0fa

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -4,7 +4,6 @@
     <clear />
     <!--Begin: Package sources managed by Dependency Flow automation. Do not edit the sources below.-->
     <!--  Begin: Package sources from dotnet-runtime -->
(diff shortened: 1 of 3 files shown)


joeloff (approved):

MiYanni (approved):
I saw you mentioned about the removal of the runtime dependency here: https://github.com/dotnet/templating/pull/7184#discussion_r1369216988

I wasn't sure what that meant exactly. Looking at the commit you linked, it still isn't clear to me why the runtime dependency was added in the first place. Something VS related potentially?

marcpopMSFT:
I pinged in the CLI channel and Jan sent me an old thread talking about the runtime dependency and it affecting codeflow but I'm still not sure why it's needed.

PR #7235, opened 2023-11-03
Remove dotnet/runtime dependency

MiYanni:
Related: https://github.com/dotnet/templating/pull/7184#discussion_r1369216988

## Summary
There was an email discussion about this dependency being added in the past for testing purposes. We no longer think this is required as the SDK repo runs tests on templating when it flows into that. Having the extra dependency on runtime for this repo just complicates the code flow.

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -4,14 +4,6 @@
 
   <Import Project="$(RepositoryEngineeringDir)Package.props" />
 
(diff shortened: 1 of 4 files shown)


marcpopMSFT (approved):

marcpopMSFT:
@MiYanni we can probably do this in 1xx now as well I think. Not sure what else is required other porting this change and then fixing up codeflow.

MiYanni:
@marcpopMSFT Created the backport here: https://github.com/dotnet/templating/pull/7282

PR #7578, opened 2024-02-21
Set ProducesDotNetReleaseShippingAssets property in Publishing.props

MilenaHristova:
Add a boolean property named `ProducesDotNetReleaseShippingAssets` in Publishing.props for repos that produce .NET shipping packages (packages that we ship with the release infra)
Based on this property we will select which packages to ship as part of .NET on release day.

This is a infrastructure only change. It will add extra metadata to the MergedManifest.xml produced during CI build.

Issue: https://github.com/dotnet/release/issues/822

### DIFF
--- a/eng/Publishing.props
+++ b/eng/Publishing.props
@@ -2,5 +2,6 @@
 <Project>
    <PropertyGroup>
       <PublishingVersion>3</PublishingVersion>
+      <ProducesDotNetReleaseShippingAssets>true</ProducesDotNetReleaseShippingAssets>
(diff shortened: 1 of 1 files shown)


joeloff (approved):

MilenaHristova:
> @MilenaHristova I approved the back ports, I think we may have to wait for servicing changes (branding and codeflow to complete) before merging this into the release/xyz branches

sorry, I merged one of the PRs before I saw this. Why is it a problem to merge it now?

PR #7724, opened 2024-03-15
Fix the broken template scanning pipeline job

baronfel:
### Problem
The [pipelines](https://dev.azure.com/dnceng/internal/_build?definitionId=1025) for the template cache scanner have been broken for some time now due to an invariant changing - the tools never expected that a template would just...be completely removed.

### Solution
We allow this case and add it to the set of removed templates. It can always be re-added back once it's back on NuGet.org.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,24 @@
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "Run template scanner",
+      "type": "coreclr",
+      "request": "launch",
+      "program": "${workspaceFolder}\\artifacts\\bin\\Microsoft.TemplateSearch.TemplateDiscovery\\Debug\\net9.0\\Microsoft.TemplateSearch.TemplateDiscovery.dll",
+      "args": [
+        "--basePath",
+        "./templates",
+        "--verbose",
+        "--savePacks",
+        "--diff",
+        "true",
+        "--test",
+        "--allowPreviewPacks"
+      ]
+    }
+  ]
+}
--- a/src/Microsoft.TemplateSearch.Common/Abstractions/TemplatePackageSearchData.cs
+++ b/src/Microsoft.TemplateSearch.Common/Abstractions/TemplatePackageSearchData.cs
@@ -11,6 +11,7 @@ namespace Microsoft.TemplateSearch.Common
     /// <summary>
     /// Template package searchable data.
     /// </summary>
+    [System.Diagnostics.DebuggerDisplay("{Name}@{Version}")]
     public partial class TemplatePackageSearchData : ITemplatePackageInfo
     {
         public TemplatePackageSearchData(ITemplatePackageInfo packInfo, IEnumerable<TemplateSearchData> templates, IDictionary<string, object>? data = null)
--- a/src/Microsoft.TemplateSearch.Common/TemplateSearchCache/TemplateSearchCache.cs
+++ b/src/Microsoft.TemplateSearch.Common/TemplateSearchCache/TemplateSearchCache.cs
@@ -1,7 +1,9 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System;
 using System.Collections.Generic;
+using System.Linq;
 using Newtonsoft.Json;
 
 namespace Microsoft.TemplateSearch.Common
@@ -12,12 +14,14 @@ namespace Microsoft.TemplateSearch.Common
 
         internal TemplateSearchCache(IReadOnlyList<TemplatePackageSearchData> data)
         {
-            TemplatePackages = data;
+            // when creating from freshly-generated data, order the results for clarity
+            TemplatePackages = data.OrderBy(x => x.Name, StringComparer.OrdinalIgnoreCase).ToArray();
             Version = CurrentVersion;
         }
 
         private TemplateSearchCache(IReadOnlyList<TemplatePackageSearchData> data, string version)
         {
+            // don't order results when creating from a read file
             TemplatePackages = data;
             Version = version;
         }
--- a/tools/Microsoft.TemplateSearch.TemplateDiscovery/NuGet/NugetPackProvider.cs
+++ b/tools/Microsoft.TemplateSearch.TemplateDiscovery/NuGet/NugetPackProvider.cs
@@ -141,7 +141,6 @@ namespace Microsoft.TemplateSearch.TemplateDiscovery.NuGet
             {
                 if (File.Exists(outputPackageFileNameFullPath))
                 {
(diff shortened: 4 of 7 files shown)


.vscode/launch.json
```diff
@@ -0,0 +1,24 @@
+{
```
baronfel, line 1: I added this file to make f5 debugging of the tool as it is run in the pipeline quite simple to do.

src/Microsoft.TemplateSearch.Common/Abstractions/TemplatePackageSearchData.cs
```diff
@@ -11,6 +11,7 @@ namespace Microsoft.TemplateSearch.Common
     /// <summary>
     /// Template package searchable data.
     /// </summary>
+    [System.Diagnostics.DebuggerDisplay("{Name}@{Version}")]
```
baronfel, line 14: This workflow deals with large lists of these items, having a debugger visualization is very handy.

tools/Microsoft.TemplateSearch.TemplateDiscovery/PackChecking/FilteredPackageInfo.cs
```diff
@@ -6,6 +6,7 @@
 
 namespace Microsoft.TemplateSearch.TemplateDiscovery.PackChecking
 {
+    [System.Diagnostics.DebuggerDisplay("{Name}@{Version} - {Reason}")]
```
baronfel, line 9: This workflow deals with large lists of these items, having a debugger visualization is very handy.

tools/Microsoft.TemplateSearch.TemplateDiscovery/NuGet/NugetPackProvider.cs
```diff
@@ -45,11 +45,7 @@ internal NuGetPackProvider(string name, string query, DirectoryInfo packageTempB
 
             _searchUriFormat = $"{searchResources[0].Uri}?{query}&skip={{0}}&take={{1}}&prerelease={includePreviewPacks}&semVerLevel=2.0.0";
 
-            if (Directory.Exists(_packageTempPath))
-            {
-                throw new Exception($"temp storage path for NuGet packages already exists: {_packageTempPath}");
-            }
-            else
+            if (!Directory.Exists(_packageTempPath))
```
baronfel, line 48: These two changes allow us to re-use the already-downloaded nupkgs when re-running the application. This _really_ cuts down on cycle time.

tools/Microsoft.TemplateSearch.TemplateDiscovery/PackChecking/PackSourceChecker.cs
```diff
@@ -156,10 +156,10 @@ await foreach (ITemplatePackageInfo sourceInfo in packProvider.GetCandidatePacks
                 var packageInfo = await provider.GetPackageInfoAsync(package.Name, cancellationToken).ConfigureAwait(false);
                 if (packageInfo == default)
                 {
-                    Console.WriteLine($"Package {package.Name} cannot be verified.");
-                    throw new Exception($"Package {package.Name} is missing and cannot be verified.");
+                    Console.WriteLine($"Package {package.Name} no longer exists in the provider.");
+                    scanningStats.RemovedTemplatePacks.Add(package);
                 }
-                if (packageInfo.Removed)
+                else if (packageInfo.Removed)
```
baronfel, line 162: Finally, the core change. Templates that are unlisted are added to the 'filter me out from future use' cache, and it stands to reason that removed one could be too. This is a per-package-version filter so newer versions that are added would be re-added to the cache.

MiYanni (approved):
:shipit:

baronfel:
I'm doing a quick compare between my local version of the search cache and the version available at the search cache url - will report back if the diffs seem sensible

tools/Microsoft.TemplateSearch.TemplateDiscovery/PackChecking/PackSourceChecker.cs
marcpopMSFT: What are the different cases where packageInfo will be default versus removed?

baronfel:
Converted this to a draft because there were a lot of removals in the new cache generated locally, and I'm not confident in the results quite yet.

tools/Microsoft.TemplateSearch.TemplateDiscovery/PackChecking/PackSourceChecker.cs
baronfel: `Removed` in this tool's vocabulary maps to 'unlisted' in NuGet terminology. `default` is because the `PackageInfo` type is a struct and if not found the `GetPackageInfoAsync` method will return the `default` struct - essentially mocking out a nullable type here.

tools/Microsoft.TemplateSearch.TemplateDiscovery/NuGet/NugetPackProvider.cs
```diff
@@ -143,20 +139,28 @@ public async Task<IDownloadedPackInfo> DownloadPackageAsync(ITemplatePackageInfo
...
-                if (await _downloadResource.CopyNupkgToStreamAsync(
-                    packinfo.Name,
-                    new NuGetVersion(packinfo.Version!),
-                    packageStream,
-                    _cacheContext,
-                    NullLogger.Instance,
-                    token).ConfigureAwait(false))
+                if (File.Exists(outputPackageFileNameFullPath))
                 {
+                    using Stream existingPackageStream = File.OpenRead(outputPackageFileNameFullPath);
```
joeloff, line 144: What's the purpose of just opening the file. Doesn't seem like you're doing anything with the stream.

baronfel: good eye, I was on autopilot through this part. we should be able to skip this entirely.

baronfel:
Ok, I took the time to do a detailed diff and now I'm happy with the results. There's some shuffling of what templates are/are not in the cache file, but these is just a side-effect of the way we are querying the overall NuGet API - because we're stuck at the first 4kish responses due to the API we're using and there's no consistent package ordering, we're going to vary somewhat on the packages in the overall listing.

src/Microsoft.TemplateSearch.Common/TemplateSearchCache/TemplateSearchCache.cs
```diff
@@ -12,12 +14,14 @@ internal partial class TemplateSearchCache
 
         internal TemplateSearchCache(IReadOnlyList<TemplatePackageSearchData> data)
         {
-            TemplatePackages = data;
+            // when creating from freshly-generated data, order the results for clarity
```
baronfel, line 17: This change should make it easier for us to do side-by-side compares in the future. When I did this, it made it very easy to see that most diffs were like this:

image

meaning descriptions, versions, download counts that had changed over time. The other big source of diffs was additions/removals of entire packages from the cache file, which is inevitable until we move to searching the entire NuGet feed.

baronfel:
Two runs are failing, but I can't see the results - are they not reported in AzDO?

baronfel:
Ok, a rerun fixed this. Since it's been approved and I'm confident it in, going to merge it now. We should check the pipeline runs in an hour or so and see if they green up.

PR #7734, opened 2024-03-18
Migrate to 1ES templates

joeloff:


nagilson (commented):
I think we should delete the code ql pipeline if we arent going to use it anymore

joeloff:
> I think we should delete the code ql pipeline if we arent going to use it anymore

Deleted

azure-pipelines.yml
```diff
@@ -38,177 +37,187 @@ variables:
...
+extends:
+  ${{ if notin(variables['Build.Reason'], 'PullRequest') }}:
+    template: v1/1ES.Official.PipelineTemplate.yml@1esPipelines
+  ${{ else }}:
+    template: v1/1ES.Unofficial.PipelineTemplate.yml@1esPipelines
+  parameters:
+    sdl:
+      sourceAnalysisPool:
+        name: $(DncEngInternalBuildPool)
+        image: 1es-windows-2022-pt
```
marcpopMSFT, line 67: fyi, you don't have to use the -pt images anymore and in fact, they will go away eventually.

marcpopMSFT (approved):

azure-pipelines.yml
joeloff: I'll fix that up

PR #7736, opened 2024-03-18
Migrate to 1ES Pipeline for Cache Searching

nagilson:
Per company policy, we must migrate to this new pipeline format soon.
For further information from reviewers, please contact me for an internal link to this migration effort.

Also the log is showing changes from https://github.com/dotnet/templating/pull/7724 which was needed to get things working. 

Removes tests for resources which no longer exist and are out of support.

### DIFF
--- a/search-cache-pipeline.yml
+++ b/search-cache-pipeline.yml
@@ -27,8 +27,8 @@ extends:
   parameters:
     pool:
       name: NetCore1ESPool-Internal
-      image: 1es-windows-2022
-      os: windows
+      image: 1es-ubuntu-2204
+      os: linux
     customBuildTags:
     - ES365AIMigrationTooling
     stages:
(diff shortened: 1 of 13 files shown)


nagilson:
Updated the SAS tokens and it got further but there appear to be more azure issues. Anyway it looks like the 1ES part of this is done, this pipeline has not worked since september so I dont want to block the migration effort by trying to fix the azure resources that were already broken before this change. This improves the pipeline by making the tests pass which did not pass before this PR.

nagilson:
image
Dale from DDFUN was able to resolve the remaining issues on the azure PME and SAW. Pipeline is working again 🔥 
https://dev.azure.com/dnceng/internal/_build/results?buildId=2410140&view=results

search-cache-pipeline.yml
```diff
@@ -15,160 +18,166 @@ parameters:
...
+      sourceAnalysisPool:
+        name: NetCore1ESPool-Internal
+        image: 1es-windows-2022
+        os: windows
+    pool:
+      name: NetCore1ESPool-Internal
+      image: 1es-ubuntu-2204
+      os: linux
+    customBuildTags:
+    - ES365AIMigrationTooling
```
MiYanni, line 41: We don't need this. It is tracking information for using the conversion tool, which we didn't use here.

MiYanni (approved):

search-cache-pipeline.yml
nagilson, line 41:
```suggestion
```

PR #7748, opened 2024-03-21
Update to 1ESPT

joeloff:


nagilson (approved):
LGTM, but I dont think we want to continue to use `ubuntu-latest`. We have a 1ES Image for that now.

joeloff:
> LGTM, but I dont think we want to continue to use `ubuntu-latest`. We have a 1ES Image for that now.

Good catch, let me update that. We only do non-Windows builds for PRs and public builds, but we should use the new images for internal PRs

nagilson:
Thanks, if we did not catch that in main, we probably need to update it there too.

PR #7883, opened 2024-04-18
stabilize branding

nagilson:


eng/Versions.props
```diff
@@ -6,7 +6,7 @@
     <UsingToolNetFrameworkReferenceAssemblies>true</UsingToolNetFrameworkReferenceAssemblies>
     <VersionPrefix>8.0.300</VersionPrefix>
     <!-- When StabilizePackageVersion is set to 'true', this branch will produce stable outputs for 'Shipping' packages -->
-    <StabilizePackageVersion Condition="'$(StabilizePackageVersion)' == ''">false</StabilizePackageVersion>
+    <StabilizePackageVersion Condition="'$(StabilizePackageVersion)' == ''">true</StabilizePackageVersion>
```
joeloff, line 9: What's the reason for changing the default? This would produce stable versions for local builds when you don't set this.

nagilson: This was supposed to be 8 0 3, not main. Whoops. Thank you for catching it!

marcpopMSFT (approved):

nagilson:
rerunning osx failed jobs again..

PR #7916, opened 2024-04-30
Update referenced NuGet version

mthalman:
### Problem
Component governance is reporting this repo as referencing a vulnerable version of NuGet.Packaging: 6.8.0. This is a transitive dependency of the existing dependencies declared.

### Solution
Upgraded to NuGet 6.8.1.


mthalman:
I don't know why the tests are failing. It doesn't provide any info in the log and it's not uploading any test artifacts.

joeloff (approved):

PR #8028, opened 2024-05-29
Update cache pipeline to use AzureFileCopy task

MiYanni:
## Summary

Instead of using the Azure CLI, this uses the [AzureFileCopy@6 task](https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/azure-file-copy-v6) to push the cache json files to Azure Blob storage. It was required to make a separate job to run the new task since the task is Windows-only and the creation of the json files uses Bash (Linux).

This task will upload the 3 json files to Azure Storage:
- `NuGetTemplateSearchInfoVer2.json`
- `NuGetTemplateSearchInfo.json`
- `nonTemplatePacks.json`

I also deleted the script to install the Azure CLI since it isn't used anymore.

## Successful Runs
- https://dev.azure.com/dnceng/internal/_build/results?buildId=2462730


baronfel (approved):
Lovely change - big fan of the split between the _creation_ of the artifacts and the _publishing_ of the artifacts.

PR #8078, opened 2024-06-11
Environment.GetEnvironmentVariables always returns strings

SimonCropp:


MiYanni (approved):
I read through [this thread](https://github.com/dotnet/runtime/issues/58438) and looked at the source, and it seems it will always produce a key and value of type string:
- Windows: https://github.com/dotnet/runtime/blob/41d4c4614ebee37b861511949556aed67f3d85eb/src/libraries/System.Private.CoreLib/src/System/Environment.Variables.Windows.cs#L109-L110
- Unix: https://github.com/dotnet/runtime/blob/41d4c4614ebee37b861511949556aed67f3d85eb/src/libraries/System.Private.CoreLib/src/System/Environment.Variables.Unix.cs#L160-L162

PR #8080, opened 2024-06-11
change return type to be not null where known

SimonCropp:


MiYanni (approved):
I checked if any public members were part of public classes. They were all internal/private. Any public classes with changes only had private members modified. So, this looks good.

PR #8087, opened 2024-06-11
fix Conditionaly typo

SimonCropp:


src/Microsoft.TemplateEngine.Abstractions/PrecedenceDefinition.cs
```diff
@@ -34,7 +34,7 @@ public enum PrecedenceDefinition
     /// <summary>
     /// Parameter might become disabled - value would not be needed nor used in such case.
     /// </summary>
-    ConditionalyDisabled,
+    ConditionallyDisabled,
```
akoeplinger, line 37: we could avoid the breaking change by adding the corrected enum value and hiding the existing one

```csharp
[EditorBrowsable(EditorBrowsableState.Never)]
ConditionalyDisabled
```
and
```csharp
ConditionallyDisabled = ConditionalyDisabled
```

SimonCropp: or mark ConditionalyDisabled as obsolete

akoeplinger: or both :)

PR #8101, opened 2024-06-12
remove some un-used variables

SimonCropp:


Forgind (approved):
I'm curious if some of these were put in for testing, that is, assign a variable then print it out later and see if it has the value you expect. That's a bit redundant with a debugger, of course, but I've occasionally felt it makes things a bit easier. Regardless, I think this is still a positive change overall.

MiYanni (approved):

PR #8102, opened 2024-06-12
use IsInstanceOfType instead of IsAssignableFrom

SimonCropp:


Forgind (commented):
Is there a significant difference between these two perf-wise? They look pretty much the same to me.

SimonCropp:
no there is no perf difference. i just find IsInstanceOfType more readable since there is less noise without the `.GetType()`

MiYanni (approved):
I looked into it via our docs:
- https://learn.microsoft.com/en-us/dotnet/api/system.type.isassignablefrom?view=net-8.0
- https://learn.microsoft.com/en-us/dotnet/api/system.type.isinstanceoftype?view=net-8.0

After some research, for this situation, they should function the same way. I think this answer on SO is a good explanation of when to use these methods: https://stackoverflow.com/a/15853224/294804

PR #8105, opened 2024-06-12
remove some list allocations

SimonCropp:


src/Microsoft.TemplateEngine.Edge/Settings/TemplateMatchInfo.cs
```diff
@@ -9,9 +9,9 @@ namespace Microsoft.TemplateEngine.Edge.Settings
     [Obsolete("This implementation is deprecated, use " + nameof(TemplateMatchInfo) + " instead")]
     internal class TemplateMatchInfoEx : ITemplateMatchInfo
     {
-        private readonly IList<MatchInfo> _matchDisposition;
+        private readonly List<MatchInfo> _matchDisposition;
 
-        private readonly IList<MatchInfo> _dispositionOfDefaults;
+        private readonly List<MatchInfo> _dispositionOfDefaults;
```
Forgind, line 14: Why does it help to turn an IList into a List here? If something is a List under the hood, isn't ToList free?

Forgind (approved):
Looks good to me!

MiYanni (approved):

src/Microsoft.TemplateEngine.Edge/Settings/TemplateMatchInfo.cs
SimonCropp: because it allows us to avoid a tolist below. ie IList is not a IReadOnlyList, but List is

> If something is a List under the hood, isn't ToList free?

no. ToList always creates a new copy of the list

PR #8112, opened 2024-06-13
remove un used template locations

SimonCropp:


SimonCropp:
i missed the fact that this file is included with a csproj include

PR #8119, opened 2024-06-13
Clean-up NuGet feeds

ViktorHofer:


Forgind (approved):
I thought this was managed by arcade? But as long as CI is happy, I'm happy

ViktorHofer:
The NuGet.config file is only updated by "Arcade" during servicing and templating doesn't participate in servicing. Also, in servicing feeds are only added but existing feeds are never removed (afaik).

PR #8125, opened 2024-06-14
allow comparer to be omitted in EqualityExtensions

SimonCropp:
it would seem the intent of AllAreTheSame is to allow people to omit passing a comparer ?

given the defaulting of the comparer to a value
```
comparer ??= EqualityComparer<TValue?>.Default;
```

but the api does not allow null to be passed?

if this is not the intent, i will do a PR that removed the defaulting


MiYanni (changes requested):
To do this change without impacting the public surface in a breaking way, you can simply add an overload without the `comparer` parameter.

MiYanni:
Also, if you add the new overload, you'd need to make sure to build as it will add the API surface to the appropriate PublicAPI.Unshipped.txt document.

PR #8135, opened 2024-06-18
Add Polyfill and leverage ModuleInitializer for Verifier settings

SimonCropp:
Using a ModuleInitializer makes it much simpler to configure test settings. I illustrated its usage for the Verifier settings, but it could likely be used for many of once off config.

This PR adds Polyfill as a compile time dependency, ie `PrivateAssets="all"`. Polyfill is already in the list of allowed packages for use by MS projects. However i needed to release an update to work around a stylecop bug (it was incorrectly scanning source only packages). So you will need to add Polyfill Version 5.5.2 to the allow list. perhaps  @MattKotsenas can help with that.

this also unlocks using other newer features in netstandard, like [nullabilty attributes](https://github.com/SimonCropp/Polyfill?tab=readme-ov-file#nullable-attributes). 

if you like the polyfill approach i will submit more PRs that leverage the features it unlocks.

if not, close the PR :)


MiYanni:
Howdy Simon. Looks like Polyfill is not onboarded into our source-build process. Additionally, having a package that adds features into other runtimes is a bit tricky since this repo (along with many others including runtime) get built into the SDK itself. So, it might add confusion to be adding features that aren't actually in those runtimes to certain DLLs (like templating here). I appreciate the effort, though.

SimonCropp:
> Additionally, having a package that adds features into other runtimes

This is not what Polyfill does. It only unlocks feature that have been added to new versions of c#. (add adds some missing apis)

Take for example ModuleInitializers. The feature has been in the runtime since .net 1, but was only exposed developers via an addition to c# in later versions. so all Polyfill does in this case is add a marker attribute to the project

```
sealed class ModuleInitializerAttribute :
    Attribute;
```

C# handles the rest.

Note that all additions are internal and cant leak to consuming libraries

This is the same approach you are already using for enabling init only properties: https://github.com/dotnet/templating/blob/main/src/Shared/IsExternalInit.cs

If you still dont like the Polyfill reference, i can re-open this PR but use add the ModuleInitializerAttribute in the same way as the existing IsExternalInit?

PR #8231, opened 2024-07-16
Avoid allocations from calls to CultureInfo.GetCultures

ToddGrun:
A local profile taken during running the vsixinstaller on roslyn indicated this call was allocating quite heavily in the "devenv /updateconfiguration" process the vsixinstaller kicks off. DirectoryBasedTemplate.ParseLocFileName was allocating 8.1% of all allocations in that devenv process, nearly all of which comes from the CultureInfo.GetCultures call.

Instead of allocating all cultures and an array to hold them by calling CultureInfo.GetCultures, this code can use CultureInfo.GetCultureInfo with the requested culture name and use that result instead. Note that GetCultureInfo throws if the requested culture name isn't supported, so we need to catch the corresponding CultureNotFoundException.

image

### DIFF
--- a/src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/DirectoryBasedTemplate.cs
+++ b/src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/DirectoryBasedTemplate.cs
@@ -136,9 +136,14 @@ namespace Microsoft.TemplateEngine.Orchestrator.RunnableProjects
         {
             string filename = locFile.Name;
             string localeStr = filename.Substring(LocalizationFilePrefix.Length, filename.Length - LocalizationFilePrefix.Length - LocalizationFileExtension.Length);
+            CultureInfo? locale = null;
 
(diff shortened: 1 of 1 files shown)


joeloff (approved):

joeloff:
@ToddGrun is this something we need sooner in VS, e.g. in the current servicing builds? main is currently .NET 9 which won't ship until later this year and only in newer VS releases after 17.10.

MiYanni (approved):

ToddGrun:
> @ToddGrun is this something we need sooner in VS, e.g. in the current servicing builds? main is currently .NET 9 which won't ship until later this year and only in newer VS releases after 17.10.

As we chatted offline, this isn't high priority, but it might be worth porting to 8.0 SDKs.

marcpopMSFT:
Triage: Approved for porting to 8.0 versions. We can trigger backports. Did you want to fill the tactics template or should we?

ToddGrun:
> Triage: Approved for porting to 8.0 versions. We can trigger backports. Did you want to fill the tactics template or should we?

Would appreciate some help with that. I can try to provide any information needed.

PR #8248, opened 2024-07-29
Add the Microsoft.IO.Redist package version for CPM to version 6.0.1

marcpopMSFT:


Forgind (approved):

ericstj:
Looks to me like this will impact https://github.com/dotnet/templating/blob/5d43d89ab9da3a63359215c5abea86f63516e4f9/tools/Microsoft.TemplateEngine.Authoring.Tasks/Microsoft.TemplateEngine.Authoring.Tasks.csproj#L18.  Can we check the output of that?  What does it ship?

ericstj:
I built this branch locally and examined Microsoft.TemplateEngine.Authoring.Tasks to ensure no new package references were added it its nuspec, IO.Redist.dll does not appear in the content (the "excludeassets" is preserved from the MSBuild reference).

I do see that the new version of IO.Redist is added to 
Microsoft.TemplateEngine.Authoring.Tasks.9.0.100-dev.nupkg.730\tools\netframework\Microsoft.TemplateEngine.Authoring.Tasks.dll.config

```xml
      <dependentAssembly>
        <assemblyIdentity name="Microsoft.IO.Redist" publicKeyToken="cc7b13ffcd2ddd51" culture="neutral" />
        <bindingRedirect oldVersion="0.0.0.0-6.0.0.1" newVersion="6.0.0.1" />
      </dependentAssembly>
```

I don't think MSBuild even honors these - maybe we should stop shipping them.  @rainersigwald

marcpopMSFT:
@ericstj @rainersigwald what are the next steps I should take here? Should I exclude the reference in source build like I did in installer?

ericstj:
I think it would be nice to know if you could actually remove that config file, since it mentions things which aren't included in the package at all and may be a higher version than what's provided by MSBuild.

This is yet another Task that's responsible for updating MSBuild's dependencies.  We need an SDK to avoid these cc @baronfel.

ericstj (approved):
It doesn't make things worse than what was here before, but you should follow up on eliminating the config file.

marcpopMSFT:
@rainersigwald is there a flag to not generate that config file I can include in this PR?

ViktorHofer:
@marcpopMSFT yes, you can set the following two properties to disable the generation of an app.config file for .NET Framework:

```xml
    <AutoGenerateBindingRedirects>false</AutoGenerateBindingRedirects>
    <GenerateSupportedRuntime>false</GenerateSupportedRuntime>
```

Issue #6554, opened 2023-05-17
Correcting a link in the wiki


andychiare:
Please, can you replace the current reference to Auth0 Templates in the _[Available templates for dotnet new](https://github.com/dotnet/templating/wiki/Available-templates-for-dotnet-new)_ wiki page? The old link is no longer valid, and a new version of the templates is now available.

Currently, it's on line 27:

Replace 
```
| [Auth0 Templates](https://github.com/auth0-community/auth0-dotnet-templates) | `dotnet new install "Auth0.Templates"` |
```

with

```
| [Auth0 Templates](https://github.com/auth0/auth0-dotnet-templates) | `dotnet new install Auth0.Templates::2.0.0"` |
```

YuliiaKovalova:
The link was updated, thank you!

PR #8336, opened 2024-08-31
Update Auth0 Templates reference in Available-templates-for-dotnet-new.md

andychiare:
### Problem
The current reference to the Auth0 Templates points to a specific NuGet package version, which is obsolete.

### Solution
Point to the latest NuGet package version


MiYanni (commented):
@joeloff Does this seem fine? I don't know why it would point to a specific version of the template anyway.

andychiare:
Any update on this? 🤔

MiYanni:
@andychiare Looking at the history, [you were the one](https://github.com/dotnet/templating/issues/6554) that suggested adding the `::2.0.0` specifier. Was there was reason for this?

andychiare:
The reason was my inexperience with NuGet packages :-)

MiYanni (approved):

PR #8495, opened 2024-10-24
Remove feeds that were only needed prior to release

marcpopMSFT:
When adding the asn1 dependency, we had to add the item into the version.details.xml to make source build happy. Doing so added additional feeds as the package wasn't available yet. Now that it's available, the old feeds got deleted so deleting them manually (this is really meant to be automated by darc but we don't have codeflow from runtime anymore).


marcpopMSFT:
BTW, internal CI builds are currently broken in source build because of this.

joeloff (approved):

PR #8594, opened 2024-12-16
Change binskim to filter and only run on build legs

marcpopMSFT:
Binskim by default doesn't filter the output folder. This changes that but we have to be careful about how we do it to ensure that the folder exists every time we choose to scan it.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -45,8 +45,6 @@ parameters:
   default:
   # Always build the Release configuration, but never sign for PRs.
   - buildConfig: Release
-  ### ARCADE ###
-  preSteps: []
 
 extends:
   ${{ if notin(variables['Build.Reason'], 'PullRequest') }}:
@@ -93,7 +91,10 @@ extends:
             enableInternalSources: true
           enableTelemetry: true
           helixRepo: dotnet/templating
-          preSteps: ${{ parameters.preSteps }}
+          # WORKAROUND: BinSkim requires the folder exist prior to scanning.
+          preSteps:
+          - powershell: New-Item -ItemType Directory -Path $(Build.SourcesDirectory)/artifacts/bin -Force
+            displayName: Create artifacts/bin directory
           jobs:
           - ${{ each config in parameters.buildConfigurations }}:
             - job: Windows_NT_${{ config.buildConfig }}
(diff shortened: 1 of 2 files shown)


joeloff (approved):

azure-pipelines.yml
```diff
@@ -89,6 +89,14 @@ extends:
             enableInternalSources: true
           enableTelemetry: true
           helixRepo: dotnet/templating
+          # WORKAROUND: BinSkim requires the folder exist prior to scanning.
```
MiYanni, line 92: This should be above the `preSteps` section.

azure-pipelines.yml
```diff
@@ -89,6 +89,14 @@ extends:
...
+          templateContext:
+            sdl:
+              binskim:
+                analyzeTargetGlob: +:f|artifacts\bin\**\*.dll;+:f|artifacts\bin\**\*.exe;
+          preSteps:
+          - powershell: New-Item -ItemType Directory -Path $(Build.SourcesDirectory)/artifacts/bin -Force
+            displayName: Create artifacts/bin directory
```
MiYanni, line 99: This won't do anything without this Arcade PR: https://github.com/dotnet/arcade/pull/15325

azure-pipelines.yml
```diff
@@ -27,6 +27,8 @@ variables:
     value: true
   - name: EnableReleaseOneLocBuild
     value: false
+  - name: GDN_ALLOW_ANALYZE_NO_TARGETS
+    value: true
```
MiYanni, line 31: Where is this used?

azure-pipelines.yml
marcpopMSFT: I didn't think we were taking 15325 anymore. We needed that to do the prestep for the publish leg but by setting the templatecontext for the build itself, we only add the filter for the build. 

I thought in the SDK repo, the prestep worked for the build legs.

azure-pipelines.yml
marcpopMSFT: This was one of their recommendations earlier. I can remove it and see if it still works.

azure-pipelines.yml
MiYanni: Oh, derp. I thought this was a publish step. My bad.

azure-pipelines.yml
MiYanni: Looking at the 1ES templates, this value of `GDN_ALLOW_ANALYZE_NO_TARGETS` is set to true already as an environment variable to the BinSkim task. I don't see anything that would read this as a variable, though. It is just hardcoded as true.

PR #8595, opened 2024-12-16
Update CDN endpoint

joeloff:


marcpopMSFT:
Wait to January to merge as it'll get picked up in the pipeline right away and we don't want to risk it

richlander:
Which .NET SDK branches will this change get applied to?

baronfel:
This only applies to the main branch. 

The only hardcoded urls are part of a backend content generation process, the results of which are downloaded by user SDK installations through another redirect that isn't subject to the work we have to do for this one.

richlander:
I see. I hear you saying that the scope here is very narrow. That's good.

richlander:
We will want to consider moving to a vanity domain instead of AFD directly. We can discuss that offline. If it's worth using a CDN, it is likely worth using a vanity domain.

baronfel (approved):

baronfel:
Went ahead and merged this because we want to use templating as a low-blast-radius test of the AFD infrastructure to inform some decision making happening on Monday.

richlander:
https://github.com/dotnet/core/issues/9671

PR #8635, opened 2025-01-10
Remove parallel call Fixes #7946

### LINKED ISSUE #7946: Possible parallel workload resolver use
### Product

dotnet CLI (dotnet new)

### Describe The Bug

It seems like there may be some parallel calls to the same workload resolver after it's initialized but before any (serialized) access calls (version or manifests), and it means that we follow the lazy process for computing our manifests multiple times, leading to errors. This should not happen in the first place because the workload resolver should only ever be used from one thread.

One example of this was https://github.com/dotnet/installer/pull/19222

A partial solution is https://github.com/dotnet/sdk/pull/40485, but that doesn't get at the root cause of this instance. We currently believe that root cause is somewhere here.

### To Reproduce

See https://github.com/dotnet/installer/pull/19222

### dotnet Info

<details>
  <summary>output</summary>
  
  
  
  
</details>

### Visual Studio Version

_No response_

### Additional context

_No response_

Forgind:
Fixes #7946

dsplaisted found a stack [here](https://github.com/dotnet/templating/issues/7946#issuecomment-2108859222) that was very helpful in helping me find the source of the parallel call to the workload resolver. As seen in the top stack, this call fairly quickly reaches the WorkloadResolver, where it attempts to initialize everything concurrently, causing the issue. It's called from TemplatePackageCoordinator in the SDK repo, which in turn is called by various things.

I do not think the Parallel.For call in TemplatePackageManager (see https://github.com/dotnet/templating/issues/7946#issuecomment-2108853692) is an issue in this case. As I understand it, ScanAsync just looks at a precomputed set of ITemplatePackages, which means it doesn't have to hit the resolver to find them.

@dsplaisted @nagilson

### DIFF
--- a/src/Microsoft.TemplateEngine.Edge/Constraints/WorkloadConstraintFactory.cs
+++ b/src/Microsoft.TemplateEngine.Edge/Constraints/WorkloadConstraintFactory.cs
@@ -11,6 +11,8 @@ namespace Microsoft.TemplateEngine.Edge.Constraints
 {
     public sealed class WorkloadConstraintFactory : ITemplateConstraintFactory
     {
+        private static readonly SemaphoreSlim Mutex = new(1);
+
         Guid IIdentifiedComponent.Id { get; } = Guid.Parse("{F8BA5B13-7BD6-47C8-838C-66626526817B}");
 
         string ITemplateConstraintFactory.Type => "workload";
@@ -97,8 +99,17 @@ namespace Microsoft.TemplateEngine.Edge.Constraints
                 }
 
                 token.ThrowIfCancellationRequested();
-                IEnumerable<WorkloadInfo> currentProviderWorkloads = await providers[0].GetInstalledWorkloadsAsync(token).ConfigureAwait(false);
-                workloads = currentProviderWorkloads.ToList();
+
+                await Mutex.WaitAsync(token).ConfigureAwait(false);
+                try
+                {
+                    IEnumerable<WorkloadInfo> currentProviderWorkloads = await providers[0].GetInstalledWorkloadsAsync(token).ConfigureAwait(false);
+                    workloads = currentProviderWorkloads.ToList();
+                }
+                finally
+                {
+                    Mutex.Release();
+                }
 
                 if (workloads.Select(w => w.Id).HasDuplicates(StringComparer.InvariantCultureIgnoreCase))
                 {


src/Microsoft.TemplateEngine.Edge/TemplateConstraintManager.cs
```diff
@@ -24,7 +24,7 @@ public TemplateConstraintManager(IEngineEnvironmentSettings engineEnvironmentSet
             _logger.LogDebug($"Found {constraintFactories.Count()} constraints factories, initializing.");
             foreach (var constraintFactory in constraintFactories)
             {
-                _templateConstrains[constraintFactory.Type] = Task.Run(() => constraintFactory.CreateTemplateConstraintAsync(engineEnvironmentSettings, _cancellationTokenSource.Token));
+                _templateConstrains[constraintFactory.Type] = constraintFactory.CreateTemplateConstraintAsync(engineEnvironmentSettings, _cancellationTokenSource.Token);
```
MiYanni, line 27: I've been staring at this for a few minutes. My thoughts are... it is still running this asynchronously, since there is no `await`. The difference is that without `Task.Run`, it isn't part of a thread pool; it runs each asynchronously on the same thread.

I don't know the exact issue here, but I did notice one thing. The original `Task.Run` call forgot to pass the cancellation toaken. Meaning, it could have been:
> Task.Run(() => constraintFactory.CreateTemplateConstraintAsync(engineEnvironmentSettings, _cancellationTokenSource.Token), _cancellationTokenSource.Token);

nagilson (commented):
Thank you for investigating this change. Sorry for my delay, this was fairly complicated. I think I should have communicated the issue to you better -- you're right and it's a good observation that, that call stack doesn't call the workload constraints. ( I had to try to learn what a 'mount point' was...)

I think the simplest way to make this change (can be done with a few lines) would be to use `SemaphoreSlim.WaitAsync`. [See how to use it here](https://learn.microsoft.com/en-us/archive/msdn-magazine/2013/march/async-await-best-practices-in-asynchronous-programming#:~:text=Permits%20Asynchronous%20Synchronization-,XML,-Copy), and put it around this call to the workload resolver: https://github.com/dotnet/templating/blob/130f01eed959cbfe329cab2abcde501a1699df50/src/Microsoft.TemplateEngine.Edge/Constraints/WorkloadConstraintFactory.cs#L100

The semaphore slim is a way to have a lock object that you can run await code inside of the lock. The standard static lock in C# is not friendly with await code, which is why this semaphore slim was designed. You grab the exclusive mutex for this exact purpose. (You may already be familiar with this, but I thought it may be helpful to provide context regardless)

https://stackoverflow.com/questions/7612602/why-cant-i-use-the-await-operator-within-the-body-of-a-lock-statement This article may also be helpful.

I think this approach you're taking may be valid, but the above approach should take about 5 lines or so (Create Mutex, Acquire, Release, {, }), so it would be an easier to read solution with less churn risk.

dsplaisted (approved):
Looks good.  Were you able to test this manually at all?  It would be good to repro the issue and then verify that it doesn't repro anymore with the fix.

That may be kind of difficult, but in the past I think there have been times when we could repro this regularly.

Forgind:
> Looks good. Were you able to test this manually at all? It would be good to repro the issue and then verify that it doesn't repro anymore with the fix.
> 
> That may be kind of difficult, but in the past I think there have been times when we could repro this regularly.

I have never been able to reproduce this problem, personally, which is part of why I spent so long trying to read through code and stack traces for this...I agree that would make me feel better about it, but I don't think this will make the situation worse, at minimum. If we see a very similar-looking issue crop up again in a few months, we can revisit this to see if we can get a better fix.

PR #8816, opened 2025-03-12
Switch to AwesomeAssertions

marcpopMSFT:
Modified the CommandResultAssertion to account for the changes in version 8.

### DIFF
--- a/eng/Signing.props
+++ b/eng/Signing.props
@@ -5,7 +5,13 @@
   </PropertyGroup>
   
   <!--
-    Arcade no longer signs so signing the test biinaries used in arcade is no longer needed.
+    Arcade no longer signs so signing the test binaries used in arcade is no longer needed. Don't sign the test binaries.
   -->
+  <ItemGroup>
+    <ItemsToSign Remove="$(ArtifactsNonShippingPackagesDir)**\Microsoft.TemplateEngine.Mocks.*.nupkg" />
+    <ItemsToSign Remove="$(ArtifactsNonShippingPackagesDir)**\Microsoft.TemplateEngine.TestHelper.*.nupkg" />
+    <ItemsToSign Remove="$(ArtifactsNonShippingPackagesDir)**\Microsoft.TemplateEngine.TestTemplates.*.nupkg" />
+    <ItemsToSign Remove="$(ArtifactsNonShippingPackagesDir)**\Microsoft.TeamplateSearch.TemplateDiscovery.*.nupkg" />
+  </ItemGroup>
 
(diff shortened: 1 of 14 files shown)


marcpopMSFT:
@danmoseley a bit more changes in this repo but still relatively small. I think CommandResultAssertions is copied in a bunch of different repos based on a quick search through our codebase but the same fixes here can be used there.

eng/Signing.props
```diff
@@ -16,7 +16,7 @@
     <FileSignInfo Include="Verify.dll" CertificateName="3PartySHA2" />
     <FileSignInfo Include="Verify.Xunit.dll" CertificateName="3PartySHA2" />
     <FileSignInfo Include="EmptyFiles.dll" CertificateName="3PartySHA2" />
-    <FileSignInfo Include="FluentAssertions.dll" CertificateName="3PartySHA2" />
```
joeloff, line 19: Are we actually signing these files and redisting them? If so, we'll need to update the TPN files to remove FluentAssertions and replace it with AwesomeAssertions.

joeloff: Reviewed TPN files, didn't see anything for FA, so we don't need to add AA either

joeloff (approved):

eng/Signing.props
```diff
@@ -5,19 +5,7 @@
   </PropertyGroup>
   
   <!--
-    These are third party libraries that we use in Arcade. We need to sign them even if they
-    are already signed. However, they must be signed with a 3rd party certificate.
+    Arcade no longer signs so signing the test biinaries used in arcade is no longer needed.
```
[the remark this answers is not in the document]
marcpopMSFT: I knew I was going to need this but I wanted to remove all of the test packages first, try to sign, and get a binlog out so I could figure out which items to exclude from signing.

edvilme:
I think arcade does not like the certificate `MicrosoftDotNet500` for any of the packages signing

marcpopMSFT:
> I think arcade does not like the certificate `MicrosoftDotNet500` for any of the packages signing

We were previously signing these with a 3rd party cert. AA is not signed at all and so can't be resigned that way. These are test assemblies we don't ship so we shouldn't sign them at all. I'm trying to figure out the right way to exclude them as my last try still included them in the signing flow.

marcpopMSFT:
Re-requesting reviews as I made a lot of changes since the approvals to clear out the signing issue.

joeloff (approved):

MiYanni (approved):

Issue #8784, opened 2025-03-02
Issue: XML Conditional Processing Not Working in .slnx Files


StevenTCramer:
### Is your feature request related to a problem? Please describe.

# Issue: XML Conditional Processing Not Working in .slnx Files

## Problem Description
I'm trying to use XML conditional processing format in a .slnx file within a .NET template, but it doesn't work as expected. Since .slnx files are XML-based, I expected the standard XML conditional processing syntax to work, but it seems the templating engine doesn't recognize .slnx files as XML for conditional processing purposes.

## Steps to Reproduce
1. Create a .NET template with a .slnx file
2. Add conditional processing directives to the .slnx file using XML comment syntax:
   ```xml
   <!--#if (api)-->
   <Folder Name="/05-ContainerApps/Api/">
     <!-- Projects inside this folder -->
   </Folder>
   <!--#endif-->
   ```
3. Generate a project from the template with the condition set to false (e.g., `--api false`)
4. Observe that the conditional sections are not processed correctly - the comment directives are included in the output file, but the content between them is not excluded

## Expected Behavior
When generating a project with `--api false`, the entire section (including the conditional directives and the content between them) should be excluded from the generated .slnx file.

## Actual Behavior
The conditional directives (`<!--#if (api)-->` and `<!--#endif-->`) are included in the output file as literal text, and the content between them is also included regardless of the condition value.

## Environment
- .NET SDK version: 9.0.200
- OS: Windows 11 Pro

## Additional Context
The .slnx file format is relatively new and appears to be an XML-based solution file format. Since it's XML-based, I expected it to work with the standard XML conditional processing syntax as described in the documentation: https://github.com/dotnet/templating/wiki/Conditional-processing-and-comment-syntax#xml-based-formats

I believe this issue occurs because the .slnx extension is not mapped to XML for conditional processing purposes in the templating engine.

## Possible Solution
The templating engine could be updated to recognize .slnx files as XML-based files for conditional processing purposes, similar to how it handles .csproj, .vbproj, and other XML-based files.

Alternatively, if there's a different recommended approach for handling conditional content in .slnx files, documentation on this would be helpful.

### Describe the solution you'd like.

The templating engine could be updated to recognize .slnx files as XML-based files for conditional processing purposes, similar to how it handles .csproj, .vbproj, and other XML-based files.

baronfel:
This seems entirely reasonable to me - would you be willing to make a contribution?

StevenTCramer:
[...]

PR #8863, opened 2025-03-26
Added *.slnx with Xml conditions to OperationConfigDefault.

karl-sjogren:
### Problem
*.slnx files doesn't support conditional processing as described in #8784

### Solution
Added *.slnx with `ConditionalType.Xml` in `OperationConfigDefault.cs`. 

### Checks:
- [X] Added unit tests

I've added tests, but I'm running into some problems.

1. Can't build the `main` branch using `build.sh`, I'm getting a ton of IDE0055 errors all over the project.
2. If I disable IDE0055 in editorconfig the project builds, but I get 8 failed tests in `Microsoft.TemplateEngine.Edge.UnitTests`, still in `main`.
3. Due to the failed tests I'm not even sure if it gets to running the test I added in `End2EndTests.cs`.

### DIFF
--- a/test/Microsoft.TemplateEngine.TestTemplates/test_templates/TemplateWithConditions/test.slnx
+++ b/test/Microsoft.TemplateEngine.TestTemplates/test_templates/TemplateWithConditions/test.slnx
@@ -1,0 +1,9 @@
+﻿#if (A)
+# comment foo
+foo
+#endif
+##if (B)
+## comment bar
+#bar
+#endif
+baz
--- a/test/Microsoft.TemplateEngine.TestTemplates/test_templates/TemplateWithConditions/test.slnx
+++ b/test/Microsoft.TemplateEngine.TestTemplates/test_templates/TemplateWithConditions/test.slnx
@@ -1,9 +1,9 @@
-﻿#if (A)
-# comment foo
+﻿<!--#if (A) -->
+<!-- comment foo -->
 foo
-#endif
-##if (B)
-## comment bar
-#bar
-#endif
+<!--#endif -->
+<!--#if (B)
+<!-- comment bar -- >
+bar
+#endif -->
 baz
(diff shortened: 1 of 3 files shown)


karl-sjogren:
@dotnet-policy-service agree

karl-sjogren:
CI seems to pass though, even for OSX, so not sure why it is such as problem locally for me.

Forgind (approved):
/cc: @edvilme (slnxpert)

Forgind:
> CI seems to pass though, even for OSX, so not sure why it is such as problem locally for me.

I'm not really sure why it would be failing for you locally...best guess is some random configuration thing or out-of-date files not getting updated properly. I tend to trust CI to have a nice, clean environment prior to running tests.

test/Microsoft.TemplateEngine.TestTemplates/test_templates/TemplateWithConditions/test.slnx
```diff
@@ -0,0 +1,9 @@
+﻿<!--#if (A) -->
+<!-- comment foo -->
+foo
+<!--#endif -->
+<!--#if (B)
```
edvilme, line 5: I am not familiar with the syntax here, but is there a reason the condition here is not closed like in the first line of this file? Other than that, looks good to me.
```suggestion
<!--#if (B) -->
```

test/Microsoft.TemplateEngine.TestTemplates/test_templates/TemplateWithConditions/test.slnx
```diff
@@ -0,0 +1,9 @@
...
+<!-- comment bar -- >
+bar
+#endif -->
```
edvilme, line 8: Same thing here
```suggestion
<!--#endif -->
```

edvilme (changes requested):
> Due to the failed tests I'm not even sure if it gets to running the test I added in End2EndTests.cs.

Update: From this and the comments above, this leads me to believe the test was possibly not run correctly

test/Microsoft.TemplateEngine.TestTemplates/test_templates/TemplateWithConditions/test.slnx
edvilme: See https://github.com/dotnet/templating/blob/main/test/Microsoft.TemplateEngine.TestTemplates/test_templates/TemplateWithConditions/test.axaml

karl-sjogren: I copied the sample from https://github.com/dotnet/templating/blob/main/test/Microsoft.TemplateEngine.TestTemplates/test_templates/TemplateWithConditions/nuget.config which had this syntax when testing an XML file. So I assume that it should be an allowed syntax.

edvilme (approved):
LGTM, thanks!

PR #8865, opened 2025-03-26
Update nuget and microsoft extension packages to latest for 9.0.2xx

marcpopMSFT:
### Problem
Tooling has identified old packages being downloaded during the build. They aren't used in the SDK itself but we should get clean on CG

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -16,17 +16,17 @@
     <PackageVersion Include="Microsoft.Build.Utilities.Core" Version="17.8.3" />
    
     <!-- NuGet dependencies -->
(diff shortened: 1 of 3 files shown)


MiYanni (approved):

marcpopMSFT:
@dotnet/product-construction can I just disable the source build leg in 9.0.2xx?

akoeplinger:
@dotnet/source-build ^

marcpopMSFT:
Discussed with @MichaelSimons offline. We could go update the SBRP but source build is only used for the 1xx branches. Instead, let's disable that leg here and in SDK (but make sure the changes don't flow to main) as that will allow us to more easily update nuget versions for future updates of these servicing branches.

PR #8864, opened 2025-03-26
Update the nuget and microsoft extensions packages

marcpopMSFT:
### Problem
Tooling has identified old packages being downloaded during the build. They aren't used in the SDK itself but we should get clean on CG

PR #8952, opened 2025-04-16
Fix tasks not loading on .NET Framework

ViktorHofer:
... due to inconsistent TFMs. See https://github.com/dotnet/templating/pull/8864#issuecomment-2811060078 for more details

### Checks:
- [ ] Added unit tests


MiYanni (approved):
I like the change overall but looks like some of the code is now missing features from certain .NET versions (according to the build failures).

ViktorHofer:
Probably needs the ValueTuple package. Will fix tomorrow.

PR #9019, opened 2025-05-14
Use System.Text.Json in tests instead of unlisted log4net

v-wuzhai:
The original test package `log4net 2.0.3` has been unlisted on NuGet, which caused metadata retrieval to fail and led to test failures.

Instead of modifying product code to include unlisted packages, this PR updates the tests to use `System.Text.Json 8.0.4`, which is still listed and has known vulnerabilities. This keeps the product logic unchanged while ensuring the tests remain valid and meaningful.

### DIFF
--- a/src/Microsoft.TemplateEngine.Edge/Installers/NuGet/NugetApiPackageManager.cs
+++ b/src/Microsoft.TemplateEngine.Edge/Installers/NuGet/NugetApiPackageManager.cs
@@ -376,7 +376,7 @@ namespace Microsoft.TemplateEngine.Edge.Installers.NuGet
                 IEnumerable<IPackageSearchMetadata> packageMetadata = await resource.GetMetadataAsync(
                     packageIdentifier,
                     includePrerelease: includePrerelease,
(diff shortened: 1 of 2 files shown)


v-wuzhai:
@dotnet/templating-engine-maintainers If this approach is accepted, we can consider cherry-picking it into other branches to fix the test failures there as well.

marcpopMSFT:
This is probably not the fix we want as this would include unlisted for all customers for a specific test issue. Probably the best bet is to just pick a different package and update the tests to cover that. May I suggest using System.Text.Json version 8.0.4 for this. Can you try changing this PR to modify those two tests instead? 

An alternative would be modifying the Download api to take in an include unlisted parameter but doing that only for a test seems like more than we'd want. Sometimes we modify product code to make it more testable but when there are alternatives we can do by modifying the tests themselves, I generally advocate for that.

marcpopMSFT (approved):

marcpopMSFT:
@v-wuzhai LGTM, feel free to back port to fix the servicing branches.

PR #9939, opened 2026-03-07
Fix CI flakiness: async race conditions, disposal crash, and stdout truncation

mmitche:
## Fix CI Flakiness: 4 Root Causes → 25/25 Consecutive Passes

**Baseline**: ~63% failure rate (20/32 recent main branch builds failed)
**Result**: 25 consecutive passing CI runs with 0 failures

#### 1. `Parallel.For` async void anti-pattern (TemplatePackageManager.cs)
`Parallel.For(0, N, async (i) => ...)` silently converts the lambda to `async void`. `Parallel.For` only waits for the synchronous portion, leaving `scanResults[]` with nulls and causing `NullReferenceException`.

**Fix**: `await Task.WhenAll(Enumerable.Range(...).Select(async ...))`

#### 2. GlobalSettings disposal race condition (GlobalSettings.cs + 2 others)
`FileSystemWatcher` callbacks fire on threadpool threads that race with `Dispose()`. The original code set `_disposed = true` after `_watcher?.Dispose()`, creating a window for callbacks to run on disposed state.

**Fix**: 
- Reversed disposal ordering (`_disposed = true` first)
- Added pre-lock and post-lock disposal guards in `FileChanged`
- Added disposal guards in `GlobalSettingsTemplatePackageProvider.OnGlobalSettingsChanged` and `TemplatePackageManager` event handler

**Disposal Guard Validation**: Stress testing (100 iterations of dispose-during-callback) confirms:
- Pre-lock guard fires ~1% of races (callback after `_disposed` set)
- Post-lock guard fires ~99% of races (disposed while waiting for lock)  
- `ObjectDisposedException` catch: 0 hits (defense-in-depth for future subscribers)
- All guards are necessary and correctly prevent use-after-dispose

#### 3. Console output truncation (ExecutableCommand.cs)
`SimpleConsoleLogger`'s background queue not fully drained before exit.

**Fix**: `Console.Out.Flush()`/`Console.Error.Flush()` after logger disposal

#### 4. Artifact upload collision — all platforms (azure-pipelines-pr.yml)
Both Debug and Release matrix jobs upload to the **same** `PackageArtifacts` and `BlobArtifacts` containers simultaneously via Arcade SDK publish targets. Concurrent blob chunk uploads corrupt each other, causing ""Blob is incomplete"" errors.

**Fix**: `/p:Publish=false` for Debug builds on Windows, macOS, and Linux

### Files Changed
- `azure-pipelines-pr.yml` — artifact collision fix
- `src/.../Settings/TemplatePackageManager.cs` — Parallel.For + disposal guard
- `src/.../BuiltInManagedProvider/GlobalSettings.cs` — disposal ordering + guards
- `src/.../BuiltInManagedProvider/GlobalSettingsTemplatePackageProvider.cs` — disposal guard
- `test/.../GlobalSettingsTests.cs` — test handler comment
- `tools/.../Commands/ExecutableCommand.cs` — console flush

### CI Validation
25 consecutive passing builds on `feature/ci-flakiness` branch (build IDs: 1327510–1329184)

### DIFF
--- a/src/Microsoft.TemplateEngine.Edge/Settings/TemplatePackageManager.cs
+++ b/src/Microsoft.TemplateEngine.Edge/Settings/TemplatePackageManager.cs
@@ -22,6 +22,7 @@ namespace Microsoft.TemplateEngine.Edge.Settings
         private readonly Scanner _installScanner;
         private volatile TemplateCache? _userTemplateCache;
         private Dictionary<ITemplatePackageProvider, Task<IReadOnlyList<ITemplatePackage>>>? _cachedSources;
+        private volatile bool _disposed;
 
         /// <summary>
         /// Creates the instance.
@@ -116,6 +117,7 @@ namespace Microsoft.TemplateEngine.Edge.Settings
 
         public void Dispose()
         {
+            _disposed = true;
             if (_cachedSources == null)
             {
                 return;
@@ -255,6 +257,12 @@ namespace Microsoft.TemplateEngine.Edge.Settings
             {
                 provider.TemplatePackagesChanged += () =>
                 {
+                    // Events from providers may be in-flight when Dispose is called. Guard against
+                    // updating _cachedSources or raising TemplatePackagesChanged on a disposed instance.
+                    if (_disposed)
+                    {
+                        return;
+                    }
                     _cachedSources[provider] = provider.GetAllTemplatePackagesAsync(default);
                     TemplatePackagesChanged?.Invoke();
                 };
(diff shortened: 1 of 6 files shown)


src/Microsoft.TemplateEngine.Edge/Settings/TemplatePackageManager.cs
```diff
@@ -342,7 +342,7 @@ private async Task<TemplateCache> UpdateTemplateCacheAsync(bool needsRebuild, Ca
             }
 
             var scanResults = new ScanResult?[allTemplatePackages.Count];
-            Parallel.For(0, allTemplatePackages.Count, async (index) =>
+            await Task.WhenAll(Enumerable.Range(0, allTemplatePackages.Count).Select(async index =>
```
mmitche: @marcpopMSFT This bit looks reasonable. I **think** it is correct that previously, the Pararallel.For would have returned prior to the tasks completing.

MiYanni (approved):

PR #10206, opened 2026-05-05
Disable azure devops dependabot

marcpopMSFT:
https://eng.ms/docs/products/dependabot/configuration/disable-features

It's creating internal PRs but we can't accept those as this repo is managed in GH.


marcpopMSFT:
They're going to disable us in the backend.

PR #10384, opened 2026-07-23
Fix flaky output truncation in CLI integration tests

joeloff:
StreamForwarder read the child process's redirected output on a thread-pool thread (Task.Run) one character at a time. Under CI load the thread pool can be starved, delaying the reader so the OS pipe fills and the child blocks on Console.Write. The child's asynchronous console logger (AddSimpleConsole) then drops queued messages when its bounded flush-on-dispose (~1.5s) times out, truncating captured output and producing cut-off Verify received.txt files.

Read on a dedicated long-running thread and in 4096-char chunks so the pipe is drained promptly and the child is never blocked long enough to lose output.


baronfel:
wow, what a find

baronfel (approved):

